PR #20, opened 2014-04-28
Added support to sub property.

snirs90:
- Should merge before merging "entity-reference".


plugins/restful/RestfulEntityBase.php
```diff
@@ -290,18 +290,31 @@ public function viewEntity($entity_id, $request, $account) {
...
       $property = $info['property'];
+
       if ($info['wrapper_method'] == 'value') {
 
         if (empty($wrapper->{$property})) {
           throw new Exception(format_string('Property @property does not exist.', array('@property' => $property)));
         }
 
-        if (!$value = $wrapper->{$property}->value()) {
+        if (!empty($info['sub_property'])) {
```
amitaibu: you should set a default value  `sub_proeprty = FALSE` above, like this you can just check `if(!$info['sub_property'])`

PR #21, opened 2014-04-30
Added callback function to perform on returned value.

snirs90:


plugins/restful/RestfulEntityBase.php
```diff
@@ -340,6 +346,18 @@ public function viewEntity($entity_id, $request, $account) {
         if (!$value = $sub_wrapper->value()) {
           continue;
         }
+
+        if (!empty($info['callback'])) {
+          if (method_exists($this, $info['callback'])) {
```
amitaibu: no need to check for the method. It's up to the implementing module to make sure their code is correct + it will cause an error anyway.

exceptions/RestfulCallbackException.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulCallbackException
+ */
+
+class RestfulCallbackException extends RestfulException {
```
amitaibu: This exception isn't needed, as the classes extending `RestfulException` should throw an exception that was caused due to a _client_ error, not internal error.

plugins/restful/RestfulEntityBase.php
```diff
@@ -37,6 +39,9 @@
    *      TRUE - on the entity.
    *      FALSE - on the property.
    *
+   *    @property (optional) callback
```
amitaibu: Lets call this `process_callback`, as we'll probably also add a `callback` one.

plugins/restful/RestfulEntityBase.php
amitaibu: maybe we can use `call_user_func()` that cane get a string with a callback, or array for object + method.

plugins/restful/RestfulEntityBase.php
```diff
@@ -30,6 +30,8 @@
    *
    *    @property wrapper_method
    *      The wrapper's method name to perform on the field.
+   *      (default: value)
```
amitaibu: Docs of the optional properties should be done in [this](http://drupalcode.org/project/message_subscribe.git/blob/HEAD:/message_subscribe.module#l20) form

plugins/restful/RestfulEntityBase.php
```diff
@@ -340,6 +334,14 @@ public function viewEntity($entity_id, $request, $account) {
         if (!$value = $sub_wrapper->value()) {
           continue;
         }
+
+        if (!empty($info['process_callback'])) {
+          if (!$value = call_user_func($info['process_callback'], $value)) {
+            throw new RestfulException(
```
amitaibu: remove

plugins/restful/RestfulEntityBase.php
```diff
@@ -340,6 +336,18 @@ public function viewEntity($entity_id, $request, $account) {
         if (!$value = $sub_wrapper->value()) {
           continue;
```
snirs90: Maybe to remove this?
Because if the value is empty, the field is not returned to the API and this can cause problems and double checks to the client on every field returned from the JSON.

PR #22, opened 2014-05-01
Sub property fix

snirs90:
When property is empty, the value didn't reset and the previous value was preserved.


amitaibu:
Merged.

PR #23, opened 2014-05-01
Added a callback option.

snirs90:
Merge before #22


plugins/restful/RestfulEntityBase.php
```diff
@@ -303,53 +312,68 @@ public function viewEntity($entity_id, $request, $account) {
         continue;
       }
 
-      // Set default values.
-      $info += array(
-        'wrapper_method' => 'value',
```
amitaibu, line 308: keep the set defaults here, and also add `'callback' => FALSE`.
Like this we can return early from the callback.

plugins/restful/RestfulEntityBase.php
```diff
@@ -309,47 +318,63 @@ public function viewEntity($entity_id, $request, $account) {
...
         'sub_property' => FALSE,
         'process_callback' => FALSE,
+        'callback' => FALSE,
       );
 
-      $property = $info['property'];
+      if (!empty($info['callback'])) {
+        // Calling a callback to receive the value.
+
+        if (!$value = call_user_func($info['callback'], $value)) {
```
amitaibu: a callback may return NULL, so we should actually use `is_callable()` before and throw exception.
This should be changed for `callback_process` as-well.

plugins/restful/RestfulEntityBase.php
```diff
@@ -309,47 +318,63 @@ public function viewEntity($entity_id, $request, $account) {
         'wrapper_method_on_entity' => FALSE,
         'sub_property' => FALSE,
         'process_callback' => FALSE,
+        'callback' => FALSE,
```
amitaibu, line 333: maybe we need to call it `view_callback`, as create/ update might need a different callback.

amitaibu:
Merged.

PR #26, opened 2014-05-07
Add Entity with Multiple bundles base class for listing

snirs90:
This class allows a query on a specific entity, with explicitly defined bundles and the resource that should be used in order to render them.
This can be used for example with the "Message" module, where you would like to get a list of predefined messages form the server.

restful-multiple-bundles


plugins/restful/RestfulEntityBaseMultipleBundles.php
```diff
@@ -0,0 +1,66 @@
...
+ * @file
+ * Contains RestfulEntityBaseMultipleBundles.
+ */
+
+class RestfulEntityBaseMultipleBundles extends RestfulEntityBase {
+
+  /**
+   * Define the bundles to expose to the API.
+   *
+   * Bundle => Resource
```
amitaibu: @var array

amitaibu: `Array keyed by the bundle machine and, the the RESTful resource as the value`

amitaibu:
Add tests and merged. Thanks @snirs90 for starting this.

PR #28, opened 2014-05-08
Allow Update resource (via PUT)

amitaibu:
This PR is doing exactly what http://williamdurand.fr/2014/02/14/please-do-not-patch-like-an-idiot/ asks not to do. Even though it's not pure REST, it's much easier for the client to update the resource like that.


amitaibu:
Merged.

Issue #34, opened 2014-05-18
Add base taxonomy class


snirs90:
Resources from entity type "taxonomy" need a separate taxonomy base class because the base class of "Node" have a property condition of `status` to expose only published nodes which taxonomy doesn't have.

PR #35, opened 2014-05-18
Add base taxonomy class

snirs90:
#34


plugins/restful/RestfulEntityBaseTaxonomy.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+
+/**
+ * @file
+ * Contains RestfulEntityBaseTaxonomy.
```
amitaibu: Name (and comment) should be `RestfulEntityBaseTaxonomyTerm`

amitaibu:
Instead of this PR, we should have hook_entity_info_alter() and add the `label` property under `entity keys`.

amitaibu:
We actually already have the `label` property, so we can even remove the node base class, and move the logic into `RestfulBase` (`entity_get_info()`...)

snirs90:
The logic of `getPublicFields()` override in `RestfulEntityBaseNode` has been moved to `RestfulEntityBase` and implemented with `entity_get_info()`...

amitaibu:
I've omitted the term class as it's empty, and committed the rest. Thanks.

PR #36, opened 2014-05-19
Increased verbosity in phpdocs.

e0ipso:
@amitaibu I was reading the code this morning and started fixing some PHPStorm warnings. That rapidly became a PR being more verbose in phpdocs.

Not a very useful PR, but I just started reading the stuff :stuck_out_tongue:

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -312,7 +312,7 @@ abstract class RestfulEntityBase implements RestfulEntityInterface {
    *
    * @throws RestfulBadRequestException
    */
-  public function getList($request = NULL, $account = NULL) {
+  public function getList($request = NULL, stdClass $account = NULL) {
     $entity_type = $this->entityType;
     $result = $this
       ->getQueryForList($request, $account)
@@ -349,7 +349,7 @@ abstract class RestfulEntityBase implements RestfulEntityInterface {
    *
    * @throws RestfulBadRequestException
    */
-  public function getQueryForList($request, $account = NULL) {
+  public function getQueryForList($request, stdClass $account = NULL) {
     $query = new EntityFieldQuery();
     $query->entityCondition('entity_type', $this->entityType);
 
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -12,53 +12,62 @@
 abstract class RestfulEntityBase implements RestfulEntityInterface {
 
   /**
-   * The entity type.
+   * @var string $entityType
+   *   The entity type.
    */
   protected $entityType;
 
   /**
-   * The bundle.
+   * @var string $bundle
+   *   The bundle.
    */
   protected $bundle;
 
   /**
-   * The plugin definition.
+   * @var array $plugin
+   *   The plugin definition.
    */
   protected $plugin;
 
   /**
-   * The public fields that are exposed to the API.
+   * @var array $publicFields
+   *   The public fields that are exposed to the API.
    *
-   *  Array with the optional values:
-   *  - "property": The entity property (e.g. "title", "nid").
-   *  - "sub_property": A sub property name of a property to take from it the
-   *    content. This can be used for example on a text field with filtered text
-   *    input format where we would need to do $wrapper->body->value->value().
-   *    Defaults to FALSE.
-   *  - "wrapper_method": The wrapper's method name to perform on the field.
-   *    This can be used for example to get the entity label, by setting the
-   *    value to "label". Defaults to "value".
-   *  - "wrapper_method_on_entity": A Boolean to indicate on what to perform
-   *    the wrapper method. If TRUE the method will perform on the entity (e.g.
-   *    $wrapper->label()) and FALSE on the property or sub property
-   *    (e.g. $wrapper->field_reference->label()). Defaults to FALSE.
-   *  - "callback": A callable callback to get a computed value. Defaults To
-   *    FALSE.
-   *  - "process_callback": A callable callback to perform on the returned
-   *    value, or an array with the object and method. Defaults To FALSE.
-   *  - "resource": This property can be assigned only to an entity reference
-   *    field. Array of restful resources keyed by the target bundle. For
-   *    example, if the field is referencing a node entity, with "Article" and
-   *    "Page" bundles, we are able to map those bundles to their related
-   *    resource. Items with bundles that were not explicetly set would be
-   *    ignored.
-   *    array(
-   *      'article' => 'articles',
-   *      'page' => 'pages',
-   *    );
+   *    Array with the optional values:
+   *    - "property": The entity property (e.g. "title", "nid").
+   *    - "sub_property": A sub property name of a property to take from it the
+   *      content. This can be used for example on a text field with filtered text
+   *      input format where we would need to do $wrapper->body->value->value().
+   *      Defaults to FALSE.
+   *    - "wrapper_method": The wrapper's method name to perform on the field.
+   *      This can be used for example to get the entity label, by setting the
+   *      value to "label". Defaults to "value".
+   *    - "wrapper_method_on_entity": A Boolean to indicate on what to perform
+   *      the wrapper method. If TRUE the method will perform on the entity (e.g.
+   *      $wrapper->label()) and FALSE on the property or sub property
+   *      (e.g. $wrapper->field_reference->label()). Defaults to FALSE.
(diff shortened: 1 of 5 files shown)


e0ipso:
I think that there are 2 tests that won't pass since they're broken in `7.x-1.x`.

amitaibu:
Hi! 
Thanks for the PR -- I'll have a look. 

> I think that there are 2 tests that won't pass since they're broken in 7.x-1.x.

Tests are passing AFAIK in 7.x-1.x -- or that's at least what Travis is indicating on the homepage.

plugins/restful/RestfulEntityBase.php
```diff
@@ -12,53 +12,62 @@
 abstract class RestfulEntityBase implements RestfulEntityInterface {
 
   /**
-   * The entity type.
```
amitaibu, line 15: We should use D8 docs, e.g.

``` php
  /**
   * The entity type.
   *
   * @var string
   */
```

plugins/restful/RestfulEntityBase.php
```diff
@@ -165,6 +179,9 @@ public function getBundle() {
    *   (optional) The request.
    * @param null $account
    *   (optional) The user object.
+   *
+   * @return bool
```
amitaibu: Return value is wrong -- we return an array with the `entityView()`. Here and other methods.

plugins/restful/RestfulEntityBase.php
```diff
@@ -234,9 +263,12 @@ public function process($path = '', $request = NULL, $account = NULL, $method =
   /**
    * Return the controller from a given path.
    *
-   * @param $path
-   * @param $http_method
-   * @return null|string
+   * @param string $path
+   *   The requested path.
+   * @param string $http_method
+   *   The requested HTTP verb.
```
amitaibu: verb => method

plugins/restful/RestfulEntityBase.php
```diff
@@ -503,32 +537,38 @@ protected function getTargetTypeFromEntityReference($property) {
    *
    * @param EntityMetadataWrapper $wrapper
    *   The wrapped object.
-   * @param $property
+   * @param string $property
    *   The property name (i.e. the field name).
+   * @param array $resource
+   * @param array $request
+   * @param stdClass $account
+   *   Drupal user.
```
amitaibu: => The user object.

plugins/restful/RestfulEntityBase.php
```diff
@@ -503,32 +537,38 @@ protected function getTargetTypeFromEntityReference($property) {
...
    *   The value if found, or NULL if bundle not defined.
    */
-  protected function getValueFromResource(EntityMetadataWrapper $wrapper, $property, $resource, $request, $account) {
+  protected function getValueFromResource(EntityMetadataWrapper $wrapper, $property, $resource, $request, stdClass $account) {
     $handlers = &drupal_static(__FUNCTION__, array());
 
     $target_type = $this->getTargetTypeFromEntityReference($property);
     if (!$entity = $wrapper->value()) {
-      return;
+      return NULL;
```
amitaibu: NULL isn't needed here -- as `return;` is actually retuning NULL.

plugins/restful/RestfulEntityBase.php
```diff
@@ -503,32 +537,38 @@ protected function getTargetTypeFromEntityReference($property) {
...
     }
 
+
     if (empty($handlers[$bundle])) {
       $version = $this->getVersion();
       $handlers[$bundle] = restful_get_restful_handler($resource[$bundle], $version['major'], $version['minor']);
     }
-
-    return $handlers[$bundle]->viewEntity($id, $request, $account);
+    /** @var RestfulEntityBase $bundle_handler Handler for the bundle. */
```
amitaibu, line 575: This isn't in D8 standart AFAIK, so we can remove.

plugins/restful/RestfulEntityBase.php
```diff
@@ -597,14 +641,15 @@ public function createEntity($request, $account) {
   /**
    * Set properties of the entity based on the request, and save the entity.
    *
-   * @param EntityMetadataWrapper $wrapper
```
amitaibu, line 600: You should keep EntityMetadataWrapper, as the wrapper also a property, not just an entity.

plugins/restful/RestfulEntityBase.php
```diff
@@ -618,7 +663,8 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, $
 
 
       $property_name = $info['property'];
-      if (!$this->checkPropertyAccess($wrapper->{$property_name})) {
+      $op = $wrapper->is_new->value() ? 'create' : 'edit';
```
amitaibu: op should be only `edit` not `create` -- that's how the `$wrapper->access()` is working.

amitaibu:
Oh, regarding the failing tests-- you must have missed the section that requires 2 patches to Entity API - https://github.com/Gizra/restful#module-dependencies

Feel free to review those patches and RTBC ;)

plugins/restful/RestfulEntityBase.php
```diff
@@ -730,10 +780,20 @@ public function getPublicFields() {
     return $public_fields;
   }
 
+  /**
+   * Get the request array if any.
+   *
+   * @return array
+   *
+   * @todo There is no $this->request populated anywhere.
```
amitaibu, line 793: You are right. We should also have and use (in a follow PR) `$this->getAccount()` instead of having it in different function signatures.

plugins/restful/RestfulEntityBase.php
e0ipso: I'm having some exceptions in my local due to the fact that with the `edit` op an nid and a vid are expected but not there.

plugins/restful/RestfulEntityBase.php
e0ipso, line 575: I was trying to be verbose and letting IDEs work their magic for DX.

plugins/restful/RestfulEntityBase.php
e0ipso, line 600: Gotcha.

plugins/restful/RestfulEntityBase.php
e0ipso: I don't have strong opinions on this. The same considerations as below apply, IDEs get confused when doing `return` with and without arguments within the same function/method.

amitaibu: Yeah, I prefer to be lazy and use the short `return;` for now.

plugins/restful/RestfulEntityBase.php
e0ipso: Alright, then this means that the docblock for the `process` method is also wrong.

amitaibu: sorry, I added this comment in the wrong place.
`getBundle()` returns string or bool

`patch()`/ `put()`/ `post()` return an array with the `entityView()`

plugins/restful/RestfulEntityBase.php
e0ipso: fair enough

e0ipso:
didnt read

I had not applied the patches :grin:

plugins/restful/RestfulEntityBase.php
```diff
@@ -428,6 +467,7 @@ public function viewEntity($entity_id, $request, $account) {
         if ($sub_wrapper instanceof EntityListWrapper) {
           // Multiple value.
           foreach ($sub_wrapper as $item_wrapper) {
+            /** @var EntityMetadataWrapper $item_wrapper EMW for each item. */
```
amitaibu, line 470: Please remove this as-well.

plugins/restful/RestfulEntityBase.php
amitaibu, line 575: yeah, lets remove them for now -- unless you can see an example of this in D8 core.

plugins/restful/RestfulEntityBase.php
amitaibu, line 793: Actually this function should be completely removed. We can do it in a follow up.

amitaibu:
@mateu-aguilo-bosch after the final fixes, could you git rebase so this is a single commit?

amitaibu:
@mateu-aguilo-bosch would you like me to finalize this PR?

e0ipso:
@amitaibu please do.

amitaibu:
Thanks, merged manually (with `git merge --squash`). Committed with correct credit.

PR #37, opened 2014-05-21
Authorization handlers structure.

e0ipso:
This PR adds the necessary structure to start supporting Authentication/Authorization.

This has to be seen as the base to build on top of it, since there are some gray areas that need more work. Namely:
- Make the authentication required and its flow more like D8, where we default to cookie auth and if that fails the anonymous user is returned. Right now if you don't use `authentication required` in your plugin definition, no authentication happens.
- Make the `RestfulAuthenticationBase` class to have a `canAuthenticate` method (similar to D8 Authentication Manager).
- Add authentication for Basic Auth. This shouldn't be that hard.

What this has:
- The business logic to add the authorization plugins into the restful handler.
- The business logic to return a 403 if the authorization fails.
- A new `authorization` plugin type to define authorization types. We may need an `authentication` type in the future, I'm not quite sure yet.
- A **dummy** auth-z plugin that will check the presence of a key/value as a GET param. The key and value are stored in a variable. This base class may be leveraged to check the presence of any negotiated token in the future (thinking about OAuth, Facebook, …).
- Some extra helper methods in `RestfulEntityBase`.
- Tests for the GET param authentication type.

Maybe @juampy72 is curious about this PR.

:boom:

### DIFF
--- a/plugins/authorization/RestfulAuthenticationGetString.class.php
+++ b/plugins/authorization/RestfulAuthenticationGetString.class.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * @file
+ * Authorization plugin class.
+ */
+
+class RestfulAuthenticationGetString extends RestfulAuthorizationBase {
+  /**
+   * Overrides RestfulAuthenticationBase::authorize().
+   */
+  public function authorize() {
+    // Get the name of the get parameter.
+    if ($param_name = $this->{$this->settings['parameter name method']}()) {
+      if (!empty($_GET[$param_name])) {
+        return $this->{$this->settings['parameter value method']}() == $_GET[$param_name];
+      }
+    }
+    return FALSE;
+  }
+
+  /**
+   * Get the parameter name to check.
+   */
+  protected function getParamName() {
+    return variable_get('restful_authentication_get_string_name', 'token');
+  }
+
+  /**
+   * Get the parameter value to check.
+   */
+  protected function getParamValue() {
+    $default = drupal_hmac_base64('restful_authentication_get_string_value', drupal_get_hash_salt());
+    return variable_get('restful_authentication_get_string_value', $default);
+  }
+
+}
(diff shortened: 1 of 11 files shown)


amitaibu:
Wow, an authorization PR, how cool!

amitaibu:
@mateu-aguilo-bosch @juampy72 
My question is do we _really_ need Authorizers - wouldn't an authentication be enough.

As far as I see it:
1. The entire crud is controlled on the property/ entity level thanks to Entity API, and I believe we should keep it there. This is actually our Authrization level.
2. So what we need is to provide a way for a client to authenticate, login the user, and possibly get a token back so all requests would be done with that token. Authentication can be a plugin, like you did for the Authrizers, as we might provide, cookies, token, oauth, etc.

So the next question is, assuming you agree with the above, is do we need to have the authentication plugin part of RESTful module, as it will be a generic solution?

e0ipso:
@amitaibu I had the same first instinct about just doing authentication and calling it a day. But I think that we can benefit from some sort of authorization layer/gate. Some other times you won't want to authorize a Drupal user, you will only care about the API being public or not. I am thinking about development environments, where you only want to open your API according to several rules.

Also it can be useful to offload some Drupal bootstraps to deny access if you know you're going to do it before hand. _Deny all except the ones that come from an IP range_. Of course you can do this in an upper layer, but not everyone is using those from start…

That being said, I think that we need more conversation about this (since I'm not sold on any of these two ideas) and I don't want this PR to get stalled, so I'm going to reconvert this into authentication handlers and keep the conversation going.

amitaibu:
> I am thinking about development environments, where you only want to open your API according to several rules.

Indeed for that we have `\RestfulInterface::access()`

> That being said, I think that we need more conversation about this 

I agree -- @juampy72 donde estas hombre? :)

>  so I'm going to reconvert this into authentication handlers and keep the conversation going.

Cool!

e0ipso:
I closed this issue, but please keep the conversation going.

Please, do not delete the branch.

Issue #39, opened 2014-05-23
Add basic auth provider.


e0ipso:

e0ipso:
Moving conversation to PR #41.

PR #40, opened 2014-05-23
Add the authentication provider initial structure

e0ipso:
This is a rework of https://github.com/Gizra/restful/pull/37 into authentication providers.

I like verbosity when coding, this means that I'm OK on forcing devs to make a choice on the authentication method. This PR provides a `cookie_auth` provider, but defaults to anonymous user if the `restful` _CTools_ plugin does not specify a `authentication types` array. This is open to discussion.

All in all the summary is:
- There is an authentication manager in the `RestfulEntityBase`.
- The manager knows about all the active authentication providers.
- You can ask the manager to identify the request and give you a user.
- Every provider implements `RestfulAuthenticationInterface`:
  - `applies()`: to check wether authentication can be done.
  - `authenticate()`: that gives you a user.
- The authentication providers are CTools plugins that provide a `class`. That class should implement `RestfulAuthenticationInterface` and can extend `RestfulAuthenticationBase`.

Follow ups:
- [ ] https://github.com/Gizra/restful/issues/38 (Add more (and better) authentication tests)
- [x] https://github.com/Gizra/restful/issues/39 (Add basic auth provider.)


modules/restful_example/plugins/restful/node/articles/1.0/articles__1_0.inc
```diff
@@ -4,6 +4,9 @@ $plugin = array(
   'label' => t('Articles'),
   'resource' => 'articles',
   'name' => 'articles__1_0',
+  'authentication types' => array(
+    'cookie_auth',
```
amitaibu: I think plugin name can be `cookies` - the `auth` suffix isn't needed.

plugins/authentication/RestfulAuthenticationBase.php
```diff
@@ -0,0 +1,47 @@
...
+   */
+  public function __construct($plugin) {
+    $this->settings = $plugin['settings'];
+    $this->plugin = $plugin;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applies() {
```
amitaibu, line 35: What's the use case for this -- when will a provider return FALSE?

plugins/authentication/RestfulAuthenticationCookie.class.php
```diff
@@ -0,0 +1,21 @@
...
+ * Contains RestfulAuthenticationCookie
+ */
+
+class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements RestfulAuthenticationInterface {
+
+  /**
+   * Implements RestfulAuthenticationInterface::authenticate().
+   */
+  public function authenticate() {
+    global $user;
```
amitaibu: I always prefer returning early, so I would write this:

``` php
if (!drupal_session_started()) {
  return;
}
```

Also, since `$user` is a partial object, I think we should

``` php
global $user;
return user_load($user->uid);
```

plugins/authentication/RestfulAuthenticationCookie.class.php
```diff
@@ -0,0 +1,21 @@
+<?php
+/**
+ * @file
+ * Contains RestfulAuthenticationCookie
```
amitaibu: Missing dot.

plugins/authentication/RestfulAuthenticationBase.php
e0ipso, line 35: Right now I'm working in BasicAuth. If the request does not have both headers we do nothing.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -0,0 +1,54 @@
...
+
+class RestfulAuthenticationManager extends \ArrayObject {
+
+  /**
+   * Adds the auth provider to the list.
+   *
+   * @param RestfulAuthenticationInterface $provider
+   *   The authentication plugin object.
+   */
+  public function setAuthNProvider(RestfulAuthenticationInterface $provider) {
```
amitaibu: I prefer to have a verbose name. `addAuthenticationProvider()` maybe?

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Get the user account for the request.
+   *
+   * @return stdClass
+   *   The user object.
+   */
+  public function getAccount() {
+    static $account;
```
amitaibu: `drupal_static()`

amitaibu: Just an idea. 

Maybe we can remove the $account from the `process()` function, and instead we can have `$handler->setAccount()`.
Then `$handler->getAccount()` will either return the set account, or call the authentication provider to get one.

amitaibu: Also in case we already have an account we can return early. This makes the function have one less indentation, which IMO makes it more readable.

plugins/authentication/RestfulAuthenticationBase.php
amitaibu, line 35: Makes sense -- worth adding this as example to the PHPdocs of the interface.

plugins/authentication/RestfulAuthenticationManager.php
e0ipso: > `drupal_static()`

I don't think that we ever need to reset the static cache outside the current scope, so I'm not sure that `drupal_static()` is a good fit here.

> Maybe we can remove the $account from the process() function, and instead we can have $handler->setAccount().
> Then $handler->getAccount() will either return the set account, or call the authentication provider to get one.

I almost implemented this pattern, but discarded for overkill. Now that it makes two of us seems a good idea now.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -0,0 +1,54 @@
...
+    return $account;
+  }
+
+  /**
+   * Get the array of authorization provider names. The plugin names.
+   *
+   * @return array
+   *   Array of provider names.
+   */
+  public function providerNames() {
```
amitaibu: This can probably be removed, I see it's only used in the Tests

plugins/restful/RestfulEntityBase.php
```diff
@@ -115,6 +121,7 @@ public function __construct($plugin) {
     $this->plugin = $plugin;
     $this->entityType = $plugin['entity_type'];
     $this->bundle = $plugin['bundle'];
+    $this->authenticationManager = new \RestfulAuthenticationManager();
```
amitaibu: Maybe we can do a poor's man dependency injection here, and have 

``` php
function __construct($plugin, RestfulAuthenticationInterface $auth_manager) {
  ...
  $this->authenticationManager = $auth_manager ? $auth_manager : new \RestfulAuthenticationManager();
}
```

plugins/authentication/RestfulAuthenticationManager.php
e0ipso: No strong opinion about this. I thought it may come useful in the future for some kind of listing feature. It can go away.

plugins/restful/RestfulEntityBase.php
```diff
@@ -213,13 +220,12 @@ public function delete($path = '', $request = NULL, $account = NULL) {
   }
 
   public function process($path = '', $request = NULL, $account = NULL, $method = 'get') {
-    global $user;
-    if (!$method_name = $this->getControllerFromPath($path, $method)) {
-      throw new RestfulBadRequestException('Path does not exist');
+    if (empty($account)) {
```
amitaibu: this would be replaced by `$this->getAccount()` based on https://github.com/Gizra/restful/pull/40/files#r13012617

restful.info
```diff
@@ -10,6 +10,9 @@ files[] = plugins/restful/RestfulEntityBaseNode.php
 files[] = plugins/restful/RestfulEntityBaseMultipleBundles.php
 files[] = plugins/restful/RestfulEntityInterface.php
 files[] = plugins/restful/RestfulInterface.php
+files[] = plugins/authentication/RestfulAuthenticationInterface.php
+files[] = plugins/authentication/RestfulAuthenticationBase.php
```
amitaibu: Sorry for being anal, but can this be in alphabet order? ;)

restful.module
```diff
@@ -33,24 +33,35 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
+    $plugin['options'] += array(
+    );
+  }
+  else if ($info['type'] == 'authentication') {
+    $plugin += array(
+      'description' => '',
+      'settings' => array(),
+    );
+
+    $plugin['settings'] += array(
```
amitaibu: Can be removed.

restful.module
```diff
@@ -67,6 +78,20 @@ function restful_get_restful_plugins() {
...
 /**
+ * Helper function to include CTools plugins and get all authentication plugins.
+ *
+ * @param string $plugin_name
+ *   (optional) If provided this function only returns the selected plugin.
+ *
+ * @return array
+ *   The selected plugin or all plugins for restful authentication.
+ */
+function restful_get_authentication_plugins($plugin_name = NULL) {
```
amitaibu: Lets have two separate function `restful_get_authentication_plugin()` (single) and `restful_get_authentication_plugins()` (multiple)

plugins/restful/RestfulEntityBase.php
e0ipso: LOL. I like the concept of _poor man's dependency injection_.

like it

restful.module
```diff
@@ -146,7 +171,21 @@ function restful_get_restful_handler($resource_name, $major_version = 1, $minor_
   $plugin = end($valid_plugins);
 
   $class = ctools_plugin_load_class('restful', 'restful', $plugin['name'], 'class');
-  $cache[$identifier] = new $class($plugin);
+  $handler = new $class($plugin);
+  // If the restful plugin needs authentication load the corresponding
+  // authentication plugin.
+  if (!empty($plugin['authentication types'])) {
+    // Since we can have multiple authentication plugins
```
amitaibu: => `// We can have multiple authentication plugins.`

restful.module
```diff
@@ -33,24 +33,35 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
-  );
+  if ($info['type'] == 'restful') {
+    $plugin += array(
+      'description' => '',
+      'options' => array(),
+      'major_version' => 1,
+      'minor_version' => 0,
+      'entity_type' => FALSE,
+      'bundle' => FALSE,
+    );
```
amitaibu, line 45: add `authentication_types` => array()`` to set default value.

restful.module
```diff
@@ -146,7 +171,21 @@ function restful_get_restful_handler($resource_name, $major_version = 1, $minor_
   $plugin = end($valid_plugins);
 
   $class = ctools_plugin_load_class('restful', 'restful', $plugin['name'], 'class');
-  $cache[$identifier] = new $class($plugin);
+  $handler = new $class($plugin);
+  // If the restful plugin needs authentication load the corresponding
+  // authentication plugin.
+  if (!empty($plugin['authentication types'])) {
```
amitaibu: Since we initialize the values, you can foreach without checking if !empty() as it will be an empty array.

restful.info
e0ipso: Un-anal peer review are worthless.

restful.module
```diff
@@ -146,7 +171,21 @@ function restful_get_restful_handler($resource_name, $major_version = 1, $minor_
...
+    // Since we can have multiple authentication plugins
+    foreach ($plugin['authentication types'] as $auth_plugin_name) {
+      $auth_plugin = restful_get_authentication_plugins($auth_plugin_name);
+      $auth_class = ctools_plugin_get_class($auth_plugin, 'class');
+      if (!empty($auth_class)) {
```
amitaibu: remove this check. I prefer getting a 500 error, then skipping an expected authentication.

restful.module
```diff
@@ -146,7 +171,21 @@ function restful_get_restful_handler($resource_name, $major_version = 1, $minor_
...
+        $handler->authenticationManager->setAuthNProvider(new $auth_class($auth_plugin));
```
amitaibu: for readability, please separate it into two lines

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulAuthorizationEntityTestCase
```
amitaibu: dot

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+class RestfulAuthorizationEntityTestCase extends DrupalWebTestCase {
+
+  /**
+   * Holds the generated account.
+   * @var stdClass
```
amitaibu: missing line break.

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  function setUp() {
+    parent::setUp('restful_test');
+    $this->account = $this->drupalCreateUser();
+    $this->drupalLogin($this->account);
+  }
+
+  /**
+   * Test authorizing an entity.
```
amitaibu: Maybe better `Test authenticating a user.` ?

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+   *
+   * v1.1 - Authorize GET request.
+   */
+  function testAuthenticationEntity() {
```
amitaibu: => `testAuthentication()`?

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+    global $user;
+    // v1.1 - GET string authorization.
```
amitaibu: Seems like a wrong copy paste

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+        '%name' => $provider_name,
+      )));
+    }
+
+    // Case 2. Test that the account from the authentication manager is the
+    // logged in user.
+    $user = $this->account;
+    $this->assertEqual($this->account->uid, $handler->authenticationManager->getAccount()->uid, 'The authentication manager resolved the currently logged in user.');
+
+    $auth_plugin = restful_get_authentication_plugins('cookie_auth');
```
amitaibu: maybe better have a `restful_get_authentication_handler()`

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+    $auth_class = ctools_plugin_get_class($auth_plugin, 'class');
+    $cookie_provider = new $auth_class($auth_plugin);
+
+    // Case 3. Test the 'cookie_auth' authentication provider.
+    $this->assertEqual($this->loggedInUser->uid, $cookie_provider->authenticate()->uid, 'The authentication manager resolved the currently logged in user.');
+
+    $this->drupalLogout();
+    $user = drupal_anonymous_user();
```
amitaibu: => `$user1`

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,68 @@
...
+    // Case 4. Test that the 'cookie_auth' resolves the anonymous user.
+    $this->assertEqual(0, $cookie_provider->authenticate()->uid, 'The authentication manager resolved the anonymous user.');
+
+    // Put back the user object.
+    $user = $this->originalUser;
```
amitaibu: Do we need this?

plugins/restful/RestfulEntityBase.php
amitaibu: yeah, when you put it like this, I guess it doesn't make much sense :P

plugins/authentication/RestfulAuthenticationManager.php
amitaibu: > I don't think that we ever need to reset the static cache outside the current scope

I believe this is a best practice that is important also for testing, where simpleTest needs a way to clear static cache.

plugins/authentication/RestfulAuthenticationManager.php
e0ipso: I'm going to take your suggestion despite _AuthN_ and _AuthZ_ are [commonly used terms](http://en.wikipedia.org/wiki/Authentication#Authorization).

plugins/restful/RestfulEntityBase.php
e0ipso: Since I strongly believe that the business logic to get the account belongs to the Authentication manager I will add a proxy method to get the account from `RestfulEntityInterface`.

``` php
  /**
   * Proxy method to get the account from the authenticationManager.
   *
   * @return \stdClass
   *   The user object.
   */
  public function getAccount() {
    return $this->authenticationManager->getAccount();
  }
```

restful.module
e0ipso: This matched the pattern of `ctools_get_plugins`.

restful.module
e0ipso: I think that `php new NULL();` will cause a PHP error, but I can throw an Exception.

tests/RestfulAuthenticationEntityTestCase.test
e0ipso: Actually we need to abuse the $GLOBALS['user'] for the tests, otherwise I'm not sure how to test authentication.

Core's tests actually look for success strings after posting the login form, not sure I want to go that route.

tests/RestfulAuthenticationEntityTestCase.test
e0ipso: Yeah, we need to restore the user to be the original user again.

plugins/authentication/RestfulAuthenticationCookie.class.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements RestfulAuthenticationInterface {
+
+  /**
+   * Implements RestfulAuthenticationInterface::authenticate().
+   */
+  public function authenticate() {
+    if (!drupal_session_started() && !$this->isCli()) {
+      return NULL;
```
amitaibu: Minor, but we can simply `return;`

plugins/authentication/RestfulAuthenticationInterface.php
```diff
@@ -0,0 +1,34 @@
...
+   * Authenticate the request by trying to match a user.
+   *
+   * @return \stdClass|null
+   *   The user object.
+   */
+  public function authenticate();
+
+  /**
+   * Determines if the request can be checked for authentication.
+   *
```
amitaibu, line 22: How about adding:

`For example, when authenticating with HTTP header, return FALSE if the header values do not exist.`

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulAuthenticationManager
```
amitaibu: Missing dot ;)

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -0,0 +1,78 @@
...
+   *
+   * @return \stdClass
+   *   The user object.
+   */
+  public function getAccount() {
+    // Return the previously resolved user, if any.
+    if (!empty($this->account)) {
+      return $this->account;
+    }
+    else {
```
amitaibu: This `else` isn't needed, as we return in the above IF, so it saves an indentation.

plugins/authentication/RestfulAuthenticationManager.php
amitaibu: Yes, please remove.

plugins/restful/RestfulEntityBase.php
amitaibu: The above line is now a duplicate.

restful.info
```diff
@@ -12,14 +15,15 @@ files[] = plugins/restful/RestfulEntityInterface.php
...
+files[] = exceptions/RestfulBadRequestException.php
 files[] = exceptions/RestfulException.php
 files[] = exceptions/RestfulForbiddenException.php
-files[] = exceptions/RestfulBadRequestException.php
 files[] = exceptions/RestfulGoneException.php
 files[] = exceptions/RestfulUnsupportedMediaTypeException.php
 files[] = exceptions/RestfulUnprocessableEntityException.php
 
 ; Tests
+files[] = tests/RestfulAuthenticationEntityTestCase.test
```
amitaibu: RestfulAuthenticationEntityTestCase.test => RestfulAuthenticationTestCase.test

restful.module
```diff
@@ -33,24 +33,35 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
+      'major_version' => 1,
+      'minor_version' => 0,
+      'entity_type' => FALSE,
+      'bundle' => FALSE,
+      'authentication types' => array(),
+    );
 
-  $plugin['options'] += array(
-  );
+    $plugin['options'] += array(
```
amitaibu: this and the one below can be removed.

restful.module
```diff
@@ -33,24 +33,35 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
+    );
+    $plugin['authentication types'] += array(
+    );
+  }
+  else if ($info['type'] == 'authentication') {
```
amitaibu: => `elseif`

restful.module
```diff
@@ -67,6 +78,32 @@ function restful_get_restful_plugins() {
...
+ * @return array
+ *   All plugins for restful authentication.
+ */
+function restful_get_authentication_plugins() {
+  ctools_include('plugins');
+  return ctools_get_plugins('restful', 'authentication');
+}
+
+/**
+ * Helper function to include CTools plugins and get the specified
```
amitaibu: This should be one liner.

restful.module
```diff
@@ -146,10 +183,40 @@ function restful_get_restful_handler($resource_name, $major_version = 1, $minor_
...
   $class = ctools_plugin_load_class('restful', 'restful', $plugin['name'], 'class');
-  $cache[$identifier] = new $class($plugin);
+  $handler = new $class($plugin);
+  // If the restful plugin needs authentication load the corresponding
+  // authentication plugin.
+
+  // We can have multiple authentication plugins.
+  foreach ($plugin['authentication types'] as $auth_plugin_name) {
+    $auth_handler = restful_get_authentication_handler($auth_plugin_name);
+    $handler->authenticationManager
```
amitaibu: can this be in a single line please

restful.module
```diff
@@ -146,10 +183,40 @@ function restful_get_restful_handler($resource_name, $major_version = 1, $minor_
...
+ * @return \RestfulAuthenticationInterface
+ *   The authentication provider object.
+ *
+ * @throws \RestfulException if the authentication provider does not exist.
+ */
+function restful_get_authentication_handler($auth_plugin_name) {
+  $auth_plugin = restful_get_authentication_plugin($auth_plugin_name);
+  $auth_class = ctools_plugin_get_class($auth_plugin, 'class');
+  if (empty($auth_class)) {
+    throw new \RestfulException('Unrecognized authentication provider.');
```
amitaibu: Do we need to throw our own exception -- we can just let PHP fail here, and it will reach the client as 500 error, which is ok.

Also if did need to throw an exception, we probably shouldn't throw a `RestfulException` since the error message of this class, and it's children reaches the client.

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,63 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulAuthorizationEntityTestCase.
+ */
+
+class RestfulAuthorizationEntityTestCase extends DrupalWebTestCase {
```
amitaibu: `RestfulAuthorizationEntityTestCase` => `RestfulAuthenticationTestCase`

tests/RestfulAuthenticationEntityTestCase.test
```diff
@@ -0,0 +1,63 @@
...
+    // logged in user.
+    $user = $this->account;
+    $this->assertEqual($this->account->uid, $handler->getAccount()->uid, 'The authentication manager resolved the currently logged in user.');
+
+    $cookie_provider = restful_get_authentication_handler('cookie');
+
+    // Case 3. Test the 'cookie_auth' authentication provider.
+    $this->assertEqual($this->account->uid, $cookie_provider->authenticate()->uid, 'The cookie provider resolved the currently logged in user.');
+
+    $user = drupal_anonymous_user();
```
amitaibu: maybe worth commenting about the fact that we need to hack the global user, so others won't think it's by mistake.

amitaibu:
Added some nitpick comments.
@mateu-aguilo-bosch It's really fun reviewing your code -- really high quality and attention to details!

restful.module
e0ipso: It wraps beyond 80 lines…

restful_module_-_restful_drupal_-____sites_restful_drupal_

restful.module
e0ipso: https://github.com/Gizra/restful/pull/40#discussion_r13013278 :laughing:

e0ipso:
Awwww :blush: 

I've got to say that it's a really fun project to work on.

restful.module
amitaibu: Yeah, in that case the comment should be changed to fit the 80 chars. Maybe remove the "Helper function" part.
Docs always start with one liner, and any more info that is needed can be added below that.

amitaibu: (but it's really minor, so don't worry about it)

amitaibu:
@mateu-aguilo-bosch can you please re-roll -- I see merge conflicts.

tests/modules/restful_test/plugins/restful/entity_test/entity_tests/1.1/entity_tests__1_1.inc
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$plugin = array(
```
amitaibu: Sorry, didn't notice this before -- Since this plugin is for a single bundle (`main`) it should be under `main/` folder and called `RestfulTestMainResource__1_5`

e0ipso: No Problem

e0ipso:
I f*cked my branches so hard…

amitaibu:
Merged. Thanks!!

PR #41, opened 2014-05-24
Basic auth provider

e0ipso:
This PR depends on #40 being merged. But feel free to review.

Meanwhile #40 is not merged, use [this link](https://github.com/mateu-aguilo-bosch/restful/compare/authentication-handlers...39-basic-auth-provider?expand=1) to review.


e0ipso:
This PR adresses #39.

plugins/authentication/RestfulAuthenticationBasic.class.php
```diff
@@ -0,0 +1,73 @@
...
+
+    // Do not allow any login from the current user's IP if the limit has been
+    // reached. Default is 50 failed attempts allowed in one hour. This is
+    // independent of the per-user limit to catch attempts from one IP to log
+    // in to many different user accounts.  We have a reasonably high limit
+    // since there may be only one apparent IP for all users at an institution.
+    if (!flood_is_allowed('failed_login_attempt_ip', variable_get('user_failed_login_ip_limit', 50), variable_get('user_failed_login_ip_window', 3600))) {
+      throw new RestfulFloodException(format_string('Rejected by ip flood control.'));
+    }
+    $uid = db_query_range("SELECT uid FROM {users} WHERE name = :name AND status = 1", 0, 1, array(':name' => $username))->fetchField();
```
amitaibu: Return early when possible.

``` php
if (!$uid = db_query_range("SELECT uid FROM {users} WHERE name = :name AND status = 1", 0, 1, array(':name' => $username))->fetchField()) {
  // Always register an IP-based failed login.
  flood_register_event('failed_login_attempt_ip', variable_get('user_failed_login_ip_window', 3600), ip_address());
  return;
}
```

plugins/authentication/RestfulAuthenticationBasic.class.php
```diff
@@ -0,0 +1,73 @@
...
+      // Don't allow login if the limit for this user has been reached.
+      // Default is to allow 5 failed attempts every 6 hours.
+      if (flood_is_allowed('failed_login_attempt_user', variable_get('user_failed_login_user_limit', 5), variable_get('user_failed_login_user_window', 21600), $identifier)) {
+        // We are not limited by flood control, so try to authenticate.
+        if ($uid = user_authenticate($username, $password)) {
+          // Clear the user based flood control.
+          flood_clear_event('failed_login_attempt_user', $identifier);
+          return user_load($uid);
+        }
+        else {
```
amitaibu: no need for `else` when the above IF has a return.

plugins/authentication/RestfulAuthenticationBasic.class.php
```diff
@@ -0,0 +1,73 @@
...
+          flood_register_event('failed_login_attempt_user', variable_get('user_failed_login_user_window', 3600), $identifier);
```
amitaibu: It would be nice to return a json with some info to the user, about wrong username or password. Maybe worth throwing a 401 Unauthorized exception?

e0ipso: I think that lies under the authorization domain.

Maybe we could start with a couple of dummy _authorization providers_ like:
- **User role**. So we can trigger a 401 when the user could not be authenticated and therefore the provided user does not have a `authenticated user` role.
- **User permission**. So we can limit the requests to certain user roles. I'm thinking about maybe defining permissions per endpoint & version, or just leaving it up to existing permissions.

amitaibu: > I think that lies under the authorization domain.

I don't think so. A client sent wrong username or password, so the authentication part isn't done yet. By throwing an exception after registering the flood, we can make sure the correct message will reach the client.
But maybe I'm missing your point? :)

e0ipso: I think that we should stick with the expected pattern in Drupal:
- There user could be identified: a user object is loaded.
- The user could not be identified: no error is generated and the anonymous user is loaded.

What we could do is add some sort of HAL like `_messages` section with a warning saying that the credentials are wrong. But altering the execution flow based on user roles (`authenticated user` vs `anonymous user`) seems authorization to me.

amitaibu: Lets check how other services (e.g. github) react when calling an endpoint with a wrong credentials. I would assume they would error, as the authentication errored.

amitaibu: Indeed we are getting an error:

``` bash
curl --user "caspyin:PASSWD" https://api.github.com/users/caspyin
{
  "message": "Bad credentials",
  "documentation_url": "https://developer.github.com/v3"
}
```

(example taken from [here](https://gist.github.com/caspyin/2288960))

e0ipso: Let's better move this to the Authentication manager as you did [here](https://github.com/Gizra/restful/pull/44/files#diff-84cb9136178cff57648ad43f6e671adeR54). This way we'll throw a exception if none of the authentication providers can provide a user.

e0ipso:
Let me know if there is anything else here.

amitaibu:
Can you re-roll please.

e0ipso:
Re-rolled

amitaibu:
Awesome, and thanks for the new tests. Merged!

Issue #42, opened 2014-05-24
Token auth provider


amitaibu:
- I think we should add a submodule called `restful_token_auth` that will provide an endpoint to login/ register/ retrieve password via CURL.
- Response will return a token string.
- We can have a fieldable `token_auth` entity that will have the token info, along with the user it's related to, and the created/ expire info.

Once we have this we can create an example repo of a completely decoupled AngularJs webapp talking with a completely RESTful Drupal. Which is pretty awesome ;)

e0ipso:
I believe this is closed now. w00t!

PR #44, opened 2014-05-25
Authenticate a REST call using a token

amitaibu:
#42

This is based on #41


e0ipso:
@amitaibu feel free to drop me a mention if you want me to peer review this. Just let me know when's ready.

amitaibu:
I sure will. In fact with this PR the authoriztion layer makes sense, as we'll be able to create a token with scopes.
I'll work on it a bit more, so we'll be able to discuss with some code to relate to.

amitaibu:
It's still messy, but maybe worth having a look, to see I'm getting this on the correct path.

To test you can enable `drush en restful_auth_token restful_example`

``` php
  // Get the non-expirable token.
  $handler = restful_get_restful_handler('auth_tokens');
  $result = $handler->get();
  $access_token = $result['token'];

  // Call the resource that is "protected" by the token authentication.
  $handler = restful_get_restful_handler('articles', 1, 3);
  // If $access_token isn't passed, you will get a "Bad credentials" exception.
  $result = $handler->get(1, array('access_token' => $access_token));
  dpm($result);
```

Notice the [centrelaized](https://github.com/Gizra/restful/pull/44/files#diff-84cb9136178cff57648ad43f6e671adeR54) exception of bad credentials if all providers fail.

modules/restful_auth_token/plugins/authentication/token.inc
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$plugin = array(
+  'label' => t('Token based authentication'),
+  'description' => t('Authenticate requests based on the token sent in the request.'),
+  'name' => 'token',
+  'class' => 'RestfulAuthenticationToken',
+);
```
e0ipso: Let's not hard code the arbitrary _access_token_ param name and centralize the value in the plugin definition.

``` php
$plugin = array(
  'label' => t('Token based authentication'),
  'description' => t('Authenticate requests based on the token sent in the request.'),
  'name' => 'token',
  'options' => array('param_name' => 'access_token'),
  'class' => 'RestfulAuthenticationToken',
);

```

modules/restful_auth_token/plugins/restful/restful_auth_token/auth_tokens/1.0/RestfulAuthTokens.class.php
```diff
@@ -0,0 +1,78 @@
...
+
+  /**
+   * Create a token for a user, and return its value.
+   */
+  public function getOrCreateToken($request = NULL, stdClass $account = NULL) {
+    // Check if there is a token that did not expire yet.
+    $query = new EntityFieldQuery();
+    $result = $query
+      ->entityCondition('entity_type', $this->entityType)
+      ->propertyCondition('uid', $account->uid)
```
e0ipso: Maybe check that `$account` is not NULL.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -27,24 +27,36 @@ public function addAuthenticationProvider(RestfulAuthenticationInterface $provid
...
    *   The user object.
    */
-  public function getAccount() {
+  public function getAccount($request = NULL) {
     // Return the previously resolved user, if any.
     if (!empty($this->account)) {
       return $this->account;
     }
     // Resolve the user based on the providers in the manager.
+    // @todo: Move flood control here, so we don't mark an authentication
```
e0ipso: We only need flood control when checking username & pass credentials, this means that cookie and maybe other should not register a flood control.

Besides, I think that is we test & fail 5 different credentials –using 5 different providers– we ought to increment the flood counter by 5.

modules/restful_auth_token/plugins/restful/restful_auth_token/auth_tokens/1.0/RestfulAuthTokens.class.php
```diff
@@ -0,0 +1,78 @@
...
+
+    $token_exists = FALSE;
+
+    if (!empty($result['restful_auth_token'])) {
+      $id = key($result['restful_auth_token']);
+      $auth_token = entity_load_single('restful_auth_token', $id);
+
+      if (!empty($auth_token->expire) && $auth_token->expire < REQUEST_TIME) {
+        // Token has expired, so we can delete this token.
+        $auth_token->delete();
```
e0ipso: We are operating under the assumption that there is only one token per user, but we are not enforcing it anywhere (?). I think we have 3 options:
- Enforce that in the DB layer (meh…).
- Sort EFQ by expiration date and get the one that expires the latest, if that one is expired then load all tokens for that user and delete them all. If you don't feel like bringing down the performance of the user thread this can be queued and dealt with later.
- Have some sort of garbage collection process that removes expired tokens. (Careful, sometimes having an expired token in the server means something different than not having a token at all. Specially when dealing with temporal access windows).

If we don't remove old tokens –i.e. assuming there's only this way to create them– we are coding ourselves into a potentially ever growing table with dead rows.

amitaibu: > We are operating under the assumption that there is only one token per user

Eventually no - there can be multiple tokens -- and this is were scope/ authorization can play a part. But for now I want to keep it simple.
My though it to have a variable that will decide wether to delete tokens (here and in `hook_cron()`) or keep as is.

modules/restful_auth_token/plugins/restful/restful_auth_token/auth_tokens/1.0/RestfulAuthTokens.class.php
amitaibu: The token plugin has an authentication type, so if authentication failed, we know it will throw an exception.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -27,24 +27,36 @@ public function addAuthenticationProvider(RestfulAuthenticationInterface $provid
...
+    // attempt for each provider.
     $account = NULL;
     foreach ($this as $provider) {
-      if ($provider->applies() && $account = $provider->authenticate()) {
+      if ($provider->applies($request) && $account = $provider->authenticate($request)) {
```
amitaibu, line 45: So maybe we could move the flood control here?

e0ipso, line 45: I think I'm missing the point completely. Why would you want an abstracted flood control when flood control –as far as Drupal core's one goes– only applies to a specific subset of authentication types? Isn't it better to let the authentication providers deal with it?

For instance I don't see the need of flood control for: _cookie_, _token_, _oauth_, … I do see the need for it in _basic auth_, _post raw credentials and get a cookie_, …

amitaibu, line 45: You are right :)

e0ipso, line 45: We'll need some sort of `RestfulAuthenticationFloodableInterface` (having the flood method) that _basic auth_ will have to implement.

Overkill?

amitaibu, line 45: For now lets let the providers take care of flooding.
Since we are going so fast (thanks to your amazing PRs), I'd like to make sure not to over-abstract stuff, and make sure we have a real use case for each scenario.

amitaibu:
@mateu-aguilo-bosch btw, what's your take on providing registration -- I'm not sure how we want to tackle this, do you have a clearer idea maybe?

What I imagine is an endpoint that allows you to base_auth + provide email, and if user is confirmed it responds with the token.

e0ipso:
> what's your take on providing registration

@amitaibu I think that once we have this PR in good shape we can see how to leverage –if possible without stepping into OAuth module– the token auth to conform a common OAuth flow.

Registration would be left for 2-legged OAuth, where a user/consumer (they're the same in 2-legged OAuth) can only register but not link to an existing account. 3-legged authentication will allow the user (through the consumer) to authenticate their requests with an existing account.

Finally we could provide a traditional way to register, but we will face 2 inconveniences:
- Either we send plain text passwords. Man in the middle.
- We create an account and make them go to the site to set the password. If they go to the site they may as well create the account.

amitaibu:
> We create an account and make them go to the site to set the password. 

I'm trying to think about a case where there is no site. There's the webapp, and the Drupal as backend that isn't open to users. 

> Either we send plain text passwords. Man in the middle.

Is it  different than providing a regular login/ register form where the password is sent to the server? Solved by having HTTPS?

amitaibu:

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -27,27 +27,40 @@ public function addAuthenticationProvider(RestfulAuthenticationInterface $provid
...
+      if ($this->count()) {
+        // User didn't authenticate against any provider, so we throw an error.
+        // @todo: Let each provider register a better exception message?
+        throw new \RestfulUnauthorizedException('Bad credentials');
+      }
+
+      // If the account could not be authenticated default to the global user.
       // Most of the cases the cookie provider will do this for us.
-      $account = drupal_anonymous_user();
+      $account = $user->uid ? user_load($user->uid) : drupal_anonymous_user();
```
e0ipso: I think this is confusing from a functional point of view. This is basically coupling the cookie provider into the authorization manager. Since we can have multiple authorization providers, let the implementer have the cookie authorization provider at the end of their `authorization types` instead of forcing it here.

IMO it is the developer's responsibility to provide the authentication types they want.

amitaibu: the reason we need it is to allow working with the API from inside Drupal, and unbreaking the tests. The outcome is actually as you would expect it, if you use it via API as logged in user, you don't need to re-authenticate

e0ipso: Then what's the point of having the [cookie authentication provider](https://github.com/Gizra/restful/blob/7.x-1.x/plugins/authentication/RestfulAuthenticationCookie.class.php)? Since it allows you to plug in –almost– the exact same code that there is here?

I still think that devs should be able to specify a list of authentication providers that was respected by Drupal. As a Drupal user you would expect to be logged by a cookie in your browser, that does not mean that is a best practise as an API user (specially if we are decoupling from Drupal).

e0ipso: Also, you can use in tests:

``` php
global $user;
$handler->setAccount(user_load($user->uid));
```

amitaibu: > Then what's the point of having the cookie authentication provider? 

The way I see it the authentication providers are there for you when you use curl. 
However, when you call the handler from inside Drupal and you didn't explicitly set `$handler->setAccount()`, then I would expect it to use the current user.

amitaibu: @mateu-aguilo-bosch ^^ does it sound right to you?

e0ipso: So we should be checking the `$request` variable that would only be populated if the request was processed by `restful_menu_process_callback` (meaning _via CURL_ call). We could just add some kind of flag to it to allow testability.

Let me do a quick PR.

amitaibu: > restful_menu_process_callback (meaning API call).

I think we are using opposite terminology, so lets call it `via Drupal` and `via CURL` for clarity :)

> Let me do a quick PR.

Sure.

e0ipso:
Good to go!

good stuff

amitaibu:
Merged.

PR #45, opened 2014-05-25
Customized listing example.

e0ipso:
This PR adds some minor changes to `RestfulEntityBase` to be able to alter the EFQ to allow customized listing.

I also added a customized listing example and some tests to it. I think that there's value in having an example that may be used as _documentation_ on how to alter a listing of entities.

### DIFF
--- a/modules/restful_example/plugins/restful/node/administrator/1.0/RestfulExampleRoleResource.class.php
+++ b/modules/restful_example/plugins/restful/node/administrator/1.0/RestfulExampleRoleResource.class.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * Contains RestfulExampleRoleResource.
+ */
+
+class RestfulExampleRoleResource extends \RestfulEntityBase implements \RestfulEntityInterface {
+
+  /**
+   * Overrides \RestfulEntityBase::__construct().
+   */
+  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL) {
+    parent::__construct($plugin, $auth_manager);
+    // Remove all other controllers since we only need to GET a list.
+    $this->setControllers(array(
+      '' => array(
+        'get' => 'getList',
+      )
+    ));
+  }
+
+  /**
+   * Overrides \RestfulEntityBase::publicFields().
+   */
+  public function getPublicFields() {
+    $public_fields = parent::getPublicFields();
+    $public_fields['type'] = array(
+      'property' => 'type',
+    );
+    $public_fields['roles'] = array(
+      'property' => 'author',
+      'sub_property' => 'roles',
+      'wrapper_method' => 'label',
+    );
+    return $public_fields;
+  }
+
+  /**
+   * Overrides \RestfulEntityBase::alterQueryForList().
+   */
+  public function alterQueryForList(\EntityFieldQuery $query) {
+    // Get the configured roles.
+    $options = $this->getPluginInfo('options');
+
+    // Get a list of role ids for the configured roles.
+    $roles_list = user_roles();
+    $selected_rids = array();
+    foreach ($roles_list as $rid => $role) {
+      if (in_array($role, $options['roles'])) {
+        $selected_rids[] = $rid;
+      }
+    }
+    if (empty($selected_rids)) {
+      return;
+    }
+
+    // Get the list of user ids belonging to the selected roles.
+    $uids = db_query('SELECT uid FROM {users_roles} WHERE rid IN (:rids)', array(
+      ':rids' => $selected_rids,
+    ))->fetchAllAssoc('uid');
+
+    // Restrict the list of entities to the nodes authored by any user on the
+    // list of users with the administrator role.
+    if (!empty($uids)) {
+      $query->propertyCondition('uid', array_keys($uids), 'IN');
+    }
+  }
+
+}
--- a/modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
+++ b/modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
@@ -1,0 +1,16 @@
+<?php
+
+$plugin = array(
+  'label' => t('Administrator'),
+  'resource' => 'administrator',
+  'name' => 'administrator',
+  'authentication types' => array(
+    'cookie',
+  ),
+  'entity_type' => 'node',
+  'options' => array(
+    'roles' => array('administrator'),
+  ),
+  'description' => t('Get a list of all the nodes authored by users with the administration role.'),
+  'class' => 'RestfulExampleRoleResource',
+);
--- a/modules/restful_example/plugins/restful/node/administrators/1.0/administrators__1_0.inc
+++ b/modules/restful_example/plugins/restful/node/administrators/1.0/administrators__1_0.inc
@@ -2,14 +2,18 @@
 
 $plugin = array(
   'label' => t('Administrator'),
-  'resource' => 'administrator',
-  'name' => 'administrator',
+  'resource' => 'administrators',
+  'name' => 'administrators__1_0',
   'authentication types' => array(
     'cookie',
   ),
   'entity_type' => 'node',
   'options' => array(
     'roles' => array('administrator'),
+  ),
+  'bundles' => array(
+    'article',
+    'page',
   ),
   'description' => t('Get a list of all the nodes authored by users with the administration role.'),
   'class' => 'RestfulExampleRoleResource',
--- a/modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
+++ b/modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
@@ -4,7 +4,7 @@
  * Contains RestfulExampleRoleResource.
  */
 
-class RestfulExampleRoleResource extends \RestfulEntityBaseMultipleBundles implements \RestfulEntityInterface {
+class RestfulExampleRoleResource extends \RestfulEntityBase implements \RestfulEntityInterface {
 
   /**
    * Overrides \RestfulEntityBase::__construct().
--- a/modules/restful_example/plugins/restful/node/per_role_content/1.0/per_role_content__1_0.inc
+++ b/modules/restful_example/plugins/restful/node/per_role_content/1.0/per_role_content__1_0.inc
@@ -1,20 +1,19 @@
 <?php
 
 $plugin = array(
-  'label' => t('Administrator'),
-  'resource' => 'administrators',
-  'name' => 'administrators__1_0',
+  'label' => t('Content per role'),
+  'resource' => 'per_role_content',
+  'name' => 'per_role_content__1_0',
   'authentication types' => array(
     'cookie',
   ),
   'entity_type' => 'node',
   'options' => array(
-    'roles' => array('administrator'),
+    'roles' => array(
+      'administrator',
+    ),
   ),
-  'bundles' => array(
-    'article',
-    'page',
-  ),
+  'bundle' => 'article',
   'description' => t('Get a list of all the nodes authored by users with the administration role.'),
   'class' => 'RestfulExampleRoleResource',
 );
--- a/modules/restful_example/plugins/restful/node/per_role_content/1.0/per_role_content__1_0.inc
+++ b/modules/restful_example/plugins/restful/node/per_role_content/1.0/per_role_content__1_0.inc
@@ -4,17 +4,32 @@ $plugin = array(
   'label' => t('Content per role'),
   'resource' => 'per_role_content',
   'name' => 'per_role_content__1_0',
+  // Authentication is optional, many resources are available to anonymous
+  // users. Current authentication types are:
+  //  - 'cookie': Uses regular Drupal session.
+  //  - 'basic_auth': Uses basic auth credentials.
+  //  - 'token': Use any of the methods above to get a valid token. Once you have a valid token use it to authenticate.
   'authentication types' => array(
     'cookie',
   ),
   'entity_type' => 'node',
   'options' => array(
+    // This will be populated automatically in the 'get children' callback for
+    // every child plugin.
     'roles' => array(),
   ),
   'bundle' => 'article',
   'description' => t('Get a list of all the nodes authored by users with the administration role.'),
   'class' => 'RestfulExampleRoleResource',
+  // Callback function that will return modified instances of this plugin. Use
+  // this as a way to provide a basic plugin in code and generate as many
+  // plugins as you need. This is mostly used for plugins that store some of
+  // their information in the database, such as views, blocks or exportable
+  // custom versions of plugins.
   'get children' => 'restful_restful_per_role_content_get_children',
+  // To implement, each plugin can have a 'get child' and 'get children'
+  // callback. Both of these should be implemented for performance reasons,
+  // since it is best to avoid getting all children if necessary.
   'get child' => 'restful_restful_per_role_content_get_child',
 );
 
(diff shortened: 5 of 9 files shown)


amitaibu:
We actually already have `\RestfulEntityBase::getQueryForList()` for this, so you can override by

``` php
public function getQueryForList($request, stdClass $account = NULL) {
    $query = parent::getQueryForList($request, $account);

    // Do changes
}
```

modules/restful_example/plugins/restful/node/administrator/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -0,0 +1,68 @@
...
+ * Contains RestfulExampleRoleResource.
+ */
+
+class RestfulExampleRoleResource extends \RestfulEntityBase implements \RestfulEntityInterface {
+
+  /**
+   * Overrides \RestfulEntityBase::__construct().
+   */
+  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL) {
+    parent::__construct($plugin, $auth_manager);
```
amitaibu: To override the controllers you can override the variable itself, which is more convenient:

``` php
  protected $controllers = array(
    '' => array(
      // GET returns a list of entities.
      'get' => 'getList',
    ),
  );
```

modules/restful_example/plugins/restful/node/administrator/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -0,0 +1,68 @@
...
+      'sub_property' => 'roles',
+      'wrapper_method' => 'label',
+    );
+    return $public_fields;
+  }
+
+  /**
+   * Overrides \RestfulEntityBase::alterQueryForList().
+   */
+  public function alterQueryForList(\EntityFieldQuery $query) {
```
amitaibu: this can be moved to \RestfulEntityBase::getQueryForList()``

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
```diff
@@ -0,0 +1,16 @@
+<?php
+
+$plugin = array(
```
amitaibu: this plugin doesn't expose a bundle / bundles which isn't right (in fact I believe it won't return any node, since we explicitly query for the bundle(s) mentioned in the plugin (where each bundle is mapped to a restful resource to be used for `entityView()`).

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
```diff
@@ -0,0 +1,16 @@
...
+  'label' => t('Administrator'),
+  'resource' => 'administrator',
```
amitaibu: The resource should be in plural.

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
```diff
@@ -0,0 +1,16 @@
...
+  'name' => 'administrator',
```
amitaibu: The name should include the major and minor versions (`administrators__1_0`)

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
```diff
@@ -0,0 +1,16 @@
...
+  'authentication types' => array(
+    'cookie',
```
amitaibu: How nice we already have authentication layer in! :)

plugins/restful/RestfulEntityBase.php
```diff
@@ -110,11 +110,21 @@
...
     return $this->controllers;
   }
 
   /**
+   * Set the controllers.
+   *
+   * @param array $controllers
+   *   The controllers to set.
+   */
+  public function setControllers($controllers) {
```
amitaibu: I think we can avoid this, since we change the variable itself. agree?

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
amitaibu: also, maybe change the logic of the resource a bit:
Resource name would be "content", and to get it filtered by role you pass it in the $request?

modules/restful_example/plugins/restful/node/administrator/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -0,0 +1,68 @@
+<?php
+/**
+ * Contains RestfulExampleRoleResource.
+ */
+
+class RestfulExampleRoleResource extends \RestfulEntityBase implements \RestfulEntityInterface {
```
amitaibu: why not just `class RestfulExampleRoleResource extends \RestfulEntityBase`?

amitaibu: also you probably need to extend `\RestfulEntityBaseMultipleBundles`

e0ipso: > why not just class RestfulExampleRoleResource extends \RestfulEntityBase?

Being explicit may serve an education purpose.

> also you probably need to extend \RestfulEntityBaseMultipleBundles

Yeah, let me try that.

plugins/restful/RestfulEntityBase.php
e0ipso: Ok with that. Having getters and setters is nice in many cases though. We just need to make sure that there's no use case to alter the controllers outside of a protected scope.

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
e0ipso: > in fact I believe it won't return any node

But it does…

We're not forcing it [here](https://github.com/Gizra/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L815-L818), nor [here](https://github.com/Gizra/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L373-L375).

> to get it filtered by role you pass it in the $request?

I wanted to show how you can build a list for a custom query (maybe eventually not limited within a single entity type) since that is an _actual_ use case. Being exclusively rest is OK for academic purposes, but inconvenient for real use cases. Think about an endpoint listing a label and URI of all the entities that have _field_dummy_\checkbox_ turned on.

Right now the most abstract case we have is `EFQ` and `entityreference` based, and that has its own limitations, but I don't want to get into that now.

I guess my point here was trying to say: «Hey people, there absolute freedom if you play with the settings in the plugin and override the correct methods.», rather than «you can filter a default listing by a parameter». Maybe we can find a better example (or we can provide both).

modules/restful_example/plugins/restful/node/administrator/1.0/administrator.inc
e0ipso: :+1:

modules/restful_example/plugins/restful/node/administrator/1.0/RestfulExampleRoleResource.class.php
e0ipso: You only get an error for not specifying your bundles if you extend from `RestfulEntityBaseMultipleBundles`.

Isn't there a way to say _I want all bundles defined now and in the future for this entity type_?

amitaibu: > Isn't there a way to say I want all bundles defined now and in the future for this entity type?
> Add a line note

Question is do you have a _real_ use case where you need it?

e0ipso: Sure! Imagine an endpoint where you list video entities _video_ bundle, and then you add a new bundle to support vimeo generated entities, or you add a shallow file entity with complete editorial control handled to the editors.

I've encountered many cases where I was in the need (due to shifting requirements) to create a brand new bundle for a certain entity that I wanted to be treated like the other bundle.

Does that make any sense?

modules/restful_example/plugins/restful/node/administrators/1.0/administrators__1_0.inc
```diff
@@ -0,0 +1,20 @@
+<?php
+
+$plugin = array(
+  'label' => t('Administrator'),
```
amitaibu: Maybe we should change the resource to be `own_content` - show nodes per user? I think it would be closer to life example.

modules/restful_example/plugins/restful/node/administrators/1.0/administrators__1_0.inc
```diff
@@ -0,0 +1,20 @@
...
+  'name' => 'administrators__1_0',
+  'authentication types' => array(
+    'cookie',
+  ),
+  'entity_type' => 'node',
+  'options' => array(
+    'roles' => array('administrator'),
+  ),
+  'bundles' => array(
+    'article',
```
amitaibu: note that the key is the bundle name, and the value should be the plugin name of the resource to be used for rendering the entity see [example](https://github.com/Gizra/restful/blob/7.x-1.x/tests/modules/restful_test/plugins/restful/entity_test/entity_tests/1.0/entity_tests__1_0.inc#L9-L12)

e0ipso:
@amitaibu this last commit introduces _plugin derivatives_ that will declare one plugin for each role.

Please, tell me if I'm taking this too far. I'm exploring what I can do with this myself while trying to provide a good example plugin. I think that overriding `getQueryForListing` and providing children plugins is **really** flexible and developer oriented. I know I would appreciate someone throwing me this example in my face :stuck_out_tongue:

Usage:

``` sh
curl http://127.0.01:8080/api/v1/anonymous-user
curl http://127.0.01:8080/api/v1/authenticated-user
curl http://127.0.01:8080/api/v1/administrator
```

amitaibu:
That's pretty crazy! Sounds like a good _advanced_ example. You are now forced to blog post about this ;)

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -0,0 +1,72 @@
...
+
+class RestfulExampleRoleResource extends \RestfulEntityBase implements \RestfulEntityInterface {
+
+  /**
+   * Overrides \RestfulEntityBase::__construct().
+   */
+  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL) {
+    parent::__construct($plugin, $auth_manager);
+    // Remove all other controllers since we only need to GET a list.
+    $this->controllers = array(
```
amitaibu, line 15: move to `$controllers`

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -0,0 +1,72 @@
...
+      )
+    );
+  }
+
+  /**
+   * Overrides \RestfulEntityBase::publicFields().
+   */
+  public function getPublicFields() {
+    $public_fields = parent::getPublicFields();
+    $public_fields['type'] = array(
```
amitaibu, line 27: is this needed for the example?

modules/restful_example/plugins/restful/node/per_role_content/1.0/per_role_content__1_0.inc
```diff
@@ -0,0 +1,45 @@
...
+    'cookie',
+  ),
+  'entity_type' => 'node',
+  'options' => array(
+    'roles' => array(),
+  ),
+  'bundle' => 'article',
+  'description' => t('Get a list of all the nodes authored by users with the administration role.'),
+  'class' => 'RestfulExampleRoleResource',
+  'get children' => 'restful_restful_per_role_content_get_children',
```
amitaibu, line 29: worth adding a comment about this.

modules/restful_example/plugins/restful/node/per_role_content/1.0/per_role_content__1_0.inc
```diff
@@ -0,0 +1,45 @@
...
+ */
+function restful_restful_per_role_content_get_children($plugin, $parent) {
+  $plugins = array();
+  foreach (user_roles() as $rid => $role_name) {
+    $plugins[$parent . ':' . $role_name] = $plugin;
+    $plugins[$parent . ':' . $role_name]['options'] = array(
+      'roles' => array($role_name),
+    );
+    // Create endpoints like api/v1/administrator, api/v1/authenticated-user ...
+    $plugins[$parent . ':' . $role_name]['resource'] = preg_replace('/\s+/', '-', drupal_strtolower($role_name));
```
amitaibu: can we `str_repalce()` instead of `preg_replace()`?

tests/RestfulListTestCase.test
```diff
@@ -147,5 +147,27 @@ class RestfulListTestCase extends DrupalWebTestCase {
     catch (RestfulBadRequestException $e) {
       $this->pass('Exception thrown on illegal sort property, descending.');
     }
+
+    // Test the administrator's content listing.
+    $role_name = 'administrator';
+    $plugin = ctools_get_plugins('restful', 'restful', 'per_role_content__1_0:' . $role_name);
```
amitaibu: maybe add `restful_handler_get_handler_by_name()`

amitaibu:
@mateu-aguilo-bosch ping me once it's ready. This is going to be a really nice addition.

e0ipso:
@amitaibu yeah, I think it's pretty commented now and we have the `restful_handler_get_handler_by_name` helper function.

So feel free to merge/review.

amitaibu:
In that case there's just those two minor comments:
https://github.com/Gizra/restful/pull/45#discussion_r13064135
https://github.com/Gizra/restful/pull/45#discussion_r13064144

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
e0ipso, line 15: Not sure I follow, I want to disable all endpoints that are not `''` and only allow _get_ in that one. Shouldn't I do that in `$this->controllers`?.

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
e0ipso, line 27: Not strictly needed, but I thought I would be nice to show how to export the node type. Do you agree?

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
amitaibu, line 15: https://github.com/Gizra/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBaseMultipleBundles.php#L18-L23

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
amitaibu, line 27: ok, possible, in that case we can also add the wrapper method like [here](https://github.com/Gizra/restful/blob/f091a81cb4227fc9e0f1a338cc5c5b95ac660ea3/tests/modules/restful_test/plugins/restful/entity_test/tests/1.0/RestfulTestTestsResource.class.php#L17-L18)

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
e0ipso, line 15: We are [only listing one bundle](https://github.com/Gizra/restful/pull/45/files#diff-e7637b486bb4b82d23be52dbe07352a0R7), and https://github.com/Gizra/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L76-L89

amitaibu, line 15: I this code will have the same effect as what you have currently in the `__construct()`, no?

``` php
  protected $controllers = array(
    '' => array(
      // GET returns a list of entities.
      'get' => 'getList',
    ),
  );
```

amitaibu:
Cleaned up a bit and merged, thanks!

PR #48, opened 2014-05-28
Via drupal default authentication

e0ipso:


e0ipso:
@amitaibu now we have a centralized way of detecting if the request was originated _via Drupal_ or _via CURL_.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -60,7 +59,15 @@ public function getAccount($request = NULL) {
 
       // If the account could not be authenticated default to the global user.
       // Most of the cases the cookie provider will do this for us.
-      $account = $user->uid ? user_load($user->uid) : drupal_anonymous_user();
+      $account = drupal_anonymous_user();
+
+      if (empty($request['rest_call']) && !in_array('cookie', array_keys($this->getArrayCopy()))) {
```
amitaibu: Just to confirm: so if I call an handler via Drupal as a logged in user, which has a cookie authentication, then it will result with anon user returned?

e0ipso: /me facepulms

facepalm

Let me fix that.

e0ipso: On second thoughts I see that this is actually OK. I will make a last min change though since we won't get here if the `cookie` provider was used (the _Bad Credentials_ exception).

e0ipso: To summarize: if you get here via Drupal as a logged in user then it will result in the currently logged in user. If you get here via CURL it will result in an anonymous user.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -60,7 +59,15 @@ public function getAccount($request = NULL) {
 
       // If the account could not be authenticated default to the global user.
       // Most of the cases the cookie provider will do this for us.
-      $account = $user->uid ? user_load($user->uid) : drupal_anonymous_user();
+      $account = drupal_anonymous_user();
+
+      if (empty($request['rest_call'])) {
+        global $user;
```
amitaibu, line 65: normally `global $user` is on top of the function, so I'll re-move this up before merge.

Issue #50, opened 2014-05-28
Add pagination support to getList()


amitaibu:
This will possibly also be the first time to introduce Hateoas - where we'll add links for the previous/ next links

PR #51, opened 2014-05-28
Add pagination support to getList()

amitaibu:
#50


amitaibu:
local_d7_dev_api_v1_articles_page_2

plugins/restful/RestfulEntityBase.php
```diff
@@ -404,13 +434,65 @@ public function getQueryForList($request, stdClass $account = NULL) {
...
+    $range = $this->getRange() + 1;
+    $offset = ($page - 1) * $range;
+
+    $query->range($offset, $range);
 
     return $query;
   }
 
   /**
+   * Add HATEOS links to list of item.
```
e0ipso: I believe is called _HATEOAS_.

plugins/restful/RestfulEntityBase.php
```diff
@@ -404,13 +434,65 @@ public function getQueryForList($request, stdClass $account = NULL) {
...
+   *   The array that will be returned from \RestfulEntityBase::getList().
+Passed by reference, as this will add a "_links" property to that array.
+   * @param $ids
+   *   Array of entity IDs retrieved for the list. Passed by reference, so we
+   *   can check if there is an extra item, thus know there is a "next" page.
+   * @param $request
+   *   The request array.
+   */
+  public function getListAddHateos(&$return, &$ids, $request){
+    $return['_links'] = array();
```
e0ipso, line 469: Do you think we should be using `Content-Type: application/hal+json` after this move and start being more _HAL_ friendly?

plugins/restful/RestfulEntityBase.php
```diff
@@ -404,13 +434,65 @@ public function getQueryForList($request, stdClass $account = NULL) {
...
 
     return $query;
   }
 
   /**
+   * Add HATEOS links to list of item.
+   *
+   * @param $return
+   *   The array that will be returned from \RestfulEntityBase::getList().
+Passed by reference, as this will add a "_links" property to that array.
```
e0ipso: This got uncommented or wrapped beyond 80 lines somehow.

plugins/restful/RestfulEntityBase.php
```diff
@@ -404,13 +434,65 @@ public function getQueryForList($request, stdClass $account = NULL) {
...
+   * @param $request
+   *   The request array.
+   */
+  public function getListAddHateos(&$return, &$ids, $request){
+    $return['_links'] = array();
+    $page = !empty($request['page']) ? $request['page'] : 1;
+
+    $resource_url = $this->getPluginInfo('menu_item');
+    $dummy_request = $request;
+    unset($dummy_request['rest_call'], $dummy_request['q']);
```
e0ipso: Should we have `rest_call` and `q` declared somewhere as special request parameters to remove when building urls? Perhaps a method in the base class?

restful.module
```diff
@@ -30,6 +30,12 @@ function restful_plugin_process(&$plugin, $info) {
       'hook_menu' => TRUE,
     );
 
+    if ($plugin['hook_menu'] && empty($plugin['menu_item'])) {
```
e0ipso, line 33: I'm really excited to have more flexibility when it comes to URL structure. I think this is a right move, but for consistency I would not like to see the `menu_item` being configurable in the plugin definition. That should be done via url aliases or 301 redirections, the resource URI structures should be reliable.

Thoughts?

e0ipso:
Good stuff! Two interesting conversations:
- Move towards `hal+json`? (I'm inclined to say yes, but can be talked out of it).
- Have _menu_item_ functionality as 301s.

plugins/restful/RestfulEntityBase.php
amitaibu, line 469: yeah, sounds a good idea. We should advance in baby steps here so we don't add too much complexity.

plugins/restful/RestfulEntityBase.php
amitaibu: I agree, I'll fix it.

restful.module
amitaibu, line 33: >  the resource URI structures should be reliable

I don't agree that 301/ aliases should be the approach. For example the token-auth: even though it's a resource (`api/v1/token_auth`) I'd like to expose it _only_ in `api/login`. 

Also I believe that 99% of the plugins will not touch the `menu_item`. It's just for the edge cases.

e0ipso, line 33: I'm good with that. Let smart developers abuse `menu_item` under their own risk.

amitaibu:
@mateu-aguilo-bosch tests are green, ready for review

plugins/restful/RestfulEntityBase.php
```diff
@@ -908,4 +982,32 @@ public function setAccount(\stdClass $account) {
...
+    unset($request['q'], $request['rest_call']);
+
+    // By default set URL to be absolute.
+    $options += array(
+      'absolute' => TRUE,
+      'query' => array(),
+    );
+
+    // Add the request as query strings.
+    $options['query'] += $request;
```
e0ipso: Not sure if this makes sense, but `$request` can be populated with the POST body. In that scenario I'm not sure we want to do this (imagine someone posting a file here). I know this helper method is only called from a GET context, but I can easily be used in other contexts.

I think we need to check the method or add a warning to the docs to only use it in GET context.

restful.module
```diff
@@ -101,13 +106,13 @@ function restful_menu() {
...
 
-    $resource = $plugin['resource'];
-    $items[$base_path . '/' . $resource] = array(
+    $items[$plugin['menu_item']] = array(
       'title' => $plugin['name'],
       'access callback' => 'restful_menu_access_callback',
       'access arguments' => array(1, 2),
       'page callback' => 'restful_menu_process_callback',
       'page arguments' => array(1, 2),
+      'delivery callback' => 'drupal_json_output',
```
e0ipso, line 115: :clap: well spotted.

e0ipso:
Ship it!

Merge it

amitaibu:
Merged.

PR #53, opened 2014-05-29
Added rendered entity cache.

e0ipso:
@amitaibu please review.


e0ipso:
This PR will add a cache layer on the rendered entity. With all other caches in place this one fills the gap of shortcutting the `viewEntity` process which can be costy (depending on the entity type).

According to some benchmarks I've done for a certain configuration (not very useful data, I know) this cache made requests about 15% quicker with a standard DB cache. One would expect memcache making a difference here.

plugins/restful/RestfulEntityBase.php
```diff
@@ -99,6 +99,13 @@
   protected $httpHeaders = array();
 
   /**
+   * Cache controller object.
+   *
+   * @var \DrupalCacheInterface
+   */
+  protected $cacheController;
```
amitaibu, line 113: we should add getter and setter functions.

plugins/restful/RestfulEntityBase.php
```diff
@@ -908,4 +934,128 @@ public function setAccount(\stdClass $account) {
...
+   *
+   * @see \RestfulEntityInterface::viewEntity().
+   */
+  protected function setRenderedEntityCache($data, $entity_id, $uid, $request) {
+    if (!$this->getPluginInfo('cache_render')) {
+      return;
+    }
+
+    $cid = $this->generateCacheId($entity_id, $uid, $request);
+    $this->cacheController->set($cid, $data, $this->getPluginInfo('cache_expiration'));
```
amitaibu: with the getter we could `$this->getCacheController()->set()`

restful.module
```diff
@@ -319,3 +323,55 @@ function restful_menu_process_callback($major_version, $resource_name) {
...
+  $cid = $type . '::ei' . $entity_id . '::uu';
+  _restful_invalidate_entity_cache($cid);
+}
+
+/**
+ * Implements hook_entity_delete().
+ */
+function restful_entity_delete($entity, $type) {
+  list($entity_id,,) = entity_extract_ids($type, $entity);
+  $cid = $type . '::ei' . $entity_id . '::uu';
```
amitaibu: what are the `ei` and `uu`?

restful.module
```diff
@@ -28,6 +28,10 @@ function restful_plugin_process(&$plugin, $info) {
       'bundle' => FALSE,
       'authentication types' => array(),
       'hook_menu' => TRUE,
+      'cache_render' => TRUE,
+      'cache_class' => NULL,
+      'cache_bin' => 'cache',
+      'cache_expiration' => CACHE_PERMANENT,
```
amitaibu: @bricel had a nice suggestion -- lets add an option `'cache_validate' => TRUE`. When disabled, no invalidation will happen -- it will be up to the implementor to do it.

restful.module
```diff
@@ -319,3 +323,55 @@ function restful_menu_process_callback($major_version, $resource_name) {
...
+function restful_entity_delete($entity, $type) {
+  list($entity_id,,) = entity_extract_ids($type, $entity);
+  $cid = $type . '::ei' . $entity_id . '::uu';
+  _restful_invalidate_entity_cache($cid);
+}
+
+/**
+ * Implements hook_user_update().
+ */
+function restful_user_update(&$edit, $account, $category) {
```
amitaibu, line 401: why do we need this User specific hook?

e0ipso, line 401: When a user is updated the access may be modified, we don't want to grant/deny access based on old conditions.

Also, when a user is deleted we don't want to keel stale records in the cache.

restful.module
e0ipso: Short indications for __e_ntity _i_d_ and __u_ser _u_id_

restful.module
```diff
@@ -28,6 +28,10 @@ function restful_plugin_process(&$plugin, $info) {
       'bundle' => FALSE,
       'authentication types' => array(),
       'hook_menu' => TRUE,
+      'cache_render' => TRUE,
+      'cache_class' => NULL,
+      'cache_bin' => 'cache',
```
e0ipso: We need a custom default cache bin to avoid cache stampede.

plugins/restful/RestfulEntityBase.php
e0ipso: I'm going to do magic methods to add setters and getters automatically for protected properties, this way we can get rid of a lot of boilerplate code. Agree?

plugins/restful/RestfulEntityBase.php
e0ipso, line 113: See https://github.com/Gizra/restful/pull/53/files#discussion_r13181324

e0ipso:
Tests back to green. Yeah!

plugins/restful/RestfulEntityBase.php
amitaibu: magic methods isn't very IDE friendly + not a pattern I've seen much, I'm ok with being verbose.
Unless you think I'm missing the point :)

plugins/restful/RestfulEntityBase.php
```diff
@@ -908,4 +934,129 @@ public function setAccount(\stdClass $account) {
...
+   * @param array $request
+   *   The request array to match the condition how cached entity was generated.
+   *
+   * @return string
+   *   The cache identifier.
+   */
+  protected function generateCacheId($entity_id, $uid, $request) {
+    // Get the cache ID from the selected params.
+    $version = $this->getVersion();
+    $cid = 'v' . $version['major'] . '.' . $version['minor'] . '::et' . $this->getEntityType() . '::ei' . $entity_id . '::uu' . $uid . '::pa';
```
amitaibu: Do we have to use the shorthand, because it looks like a WTF.
also, what's `pa` standing for? ;)

plugins/restful/RestfulEntityBase.php
```diff
@@ -908,4 +934,129 @@ public function setAccount(\stdClass $account) {
...
+   *   The entity ID.
+   * @param int $uid
+   *   The uid for the authenticated user.
+   * @param array $request
+   *   The request array to match the condition how cached entity was generated.
+   *
+   * @return string
+   *   The cache identifier.
+   */
+  protected function generateCacheId($entity_id, $uid, $request) {
```
amitaibu: maybe we can avoid the $uid, and get it with `$this->getAccount();`?

plugins/restful/RestfulEntityBase.php
```diff
@@ -908,4 +934,129 @@ public function setAccount(\stdClass $account) {
...
+    $cid = 'v' . $version['major'] . '.' . $version['minor'] . '::et' . $this->getEntityType() . '::ei' . $entity_id . '::uu' . $uid . '::pa';
+    $cid_params = array();
+    $request = $request ? $request : array();
+    foreach ($request as $param => $value) {
+      // Some request parameters don't affect how the entity is rendered, this
+      // means that we should skip them for the cache ID generation.
+      if (in_array($param, array('page', 'sort', 'q', 'rest_call'))) {
+        continue;
+      }
+      $cid_params[] = substr($param, 0, 2) . ':' . $value;
```
amitaibu, line 1190: instead of the substr(), we could md5():

``` php
unset($params['q'] ...)
sort($params)
$cid = md5(serialize($params));
```

restful.install
```diff
@@ -6,9 +6,26 @@
...
  * Implements hook_uninstall().
  */
 function restful_uninstall() {
   variable_del('restful_hook_menu_base_path');
 }
 
+/**
+ * Adds cache_restful cache table.
+ */
+function restful_update_7100(&$sandbox) {
```
amitaibu: I think it's  bit early for update hooks. Or are you already using it in production? ;)

plugins/restful/RestfulEntityBase.php
amitaibu, line 1190: oh, I see what your doing, you keep those initials for a smarter invalidation. Worth adding a comment about the technique + docs what each initial stands for.

restful.install
e0ipso: Dunno… force of habit? :stuck_out_tongue:

I'll kill it.

restful.module
```diff
@@ -319,3 +323,59 @@ function restful_menu_process_callback($major_version, $resource_name) {
 
   return drupal_json_output($result);
 }
+
+/**
+ * Implements hook_entity_update().
+ */
+function restful_entity_update($entity, $type) {
+  list($entity_id,,) = entity_extract_ids($type, $entity);
+  $cid = 'et' . $type . '::ei' . $entity_id . '::uu';
```
amitaibu, line 385: you would probably need to check `if ($type != 'user')` to avoid double invalidation

restful.module
amitaibu, line 401: > When a user is updated the access may be modified

Can you give an example?

> Due to the limitations for cid matching on clearing caches

Explain? invalidation the entire cache whenever a user is saved sounds quite harsh.

plugins/restful/RestfulEntityBase.php
e0ipso: I'll do it, if you don't feel comfortable with it we'll dump it. Deal?

amitaibu: cool.

restful.module
e0ipso, line 385: Actually, you may want to keep both. One thing is the render cache of a rendered user with uid = x, and the other one is the rendered entity as presented to the user with uid = x.

We need to segment the cache per uid since different uid may yield different rendering. I'm thinking about entity translation based on user language, field level access control, …

restful.module
e0ipso, line 401: > Can you give an example?

Imagin a user with _administrator_ role. The guy can access every protected resource in the site with that role, and therefore a cache starts to build up. We have to assume worst case scenario, permanent lifetime cache. That means that if that particular user is demoted to be _authenticated user_ he would still have access to the protected resources because they are in his cache –as long as the entity doesn't get updated causing cache invalidation–.

Another example would be a user updating his preferences changing the language negotiation for an entity translated api.

> Explain? invalidation the entire cache whenever a user is saved.

We have a `DrupalCacheInterface` that we use to interact with the different cache backends. One of the methods on that is `DrupalCacheInterface::clear($cid, $wildcard)`. According to the docs you can easily remove every cache record whose cache id that starts with a string `$cid`. Note the **starting with** part. There is no way to tell the cache implementation _hey you! bust all cache entries that **contain** this particular string_. Or I have never been able to find one (if you don't hack into backend specific implementations).

> sounds quite harsh.

Agreed. Should we scan the cache bin manually to compare cache ids using PHP? Sounds costy but it won't run very often…

plugins/restful/RestfulEntityBase.php
e0ipso, line 1190: Gotcha. :+1:

plugins/restful/RestfulEntityBase.php
e0ipso: Check 3d549e3.

e0ipso:
Multiple changes:
- Adds magic methods for automatic getters and setters.
- Cleans unneeded getter and setter definitions.
- Adds README.md documentation.
- Disables render cache for the token resource.
- Uses getter for cacheController access.

restful.module
amitaibu, line 401: I think the docs are misleading. Looking at `\DrupalDatabaseCache::clear()` you can pass a string with `wildcard = TRUE` and the it will\

``` php
db_delete($this->bin)
  ->condition('cid', db_like($cid) . '%', 'LIKE')
  ->execute();
```

Looking at memcache module, If I understand correctly `\MemCacheDrupal::wildcards` will do the same.

restful.module
```diff
@@ -28,6 +28,11 @@ function restful_plugin_process(&$plugin, $info) {
       'bundle' => FALSE,
       'authentication types' => array(),
       'hook_menu' => TRUE,
+      'cache_render' => TRUE,
+      'cache_class' => NULL,
+      'cache_bin' => 'cache_restful',
+      'cache_expiration' => CACHE_PERMANENT,
```
amitaibu: maybe better `cache_expire`

restful.module
```diff
@@ -28,6 +28,11 @@ function restful_plugin_process(&$plugin, $info) {
...
+      'cache_invalidation' => TRUE,
```
amitaibu: `cache_invalidate`

restful.module
```diff
@@ -28,6 +28,11 @@ function restful_plugin_process(&$plugin, $info) {
       'bundle' => FALSE,
       'authentication types' => array(),
       'hook_menu' => TRUE,
+      'cache_render' => TRUE,
```
amitaibu: may concern with defaulting to cache = TRUE, is that with improper validation there might be access issues. For example, if a permission was changed we don't invalidate, and that's fine, because RESTful can't and doesn't need to take care of all the scenarios.

restful.module
amitaibu: or maybe better `cache_simple_invalidate` -- to make it clearer it won't take care of all your business logic?

plugins/restful/RestfulInterface.php
```diff
@@ -13,8 +13,12 @@
    *
    * @param $plugin
    *   The restful plugin object.
+   * @param RestfulAuthenticationManager $auth_manager
+   *   Injected authentication manager.
+   * @param DrupalCacheInterface $cache_controller
+   *   Injected cache controller.
    */
-  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL);
+  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL);
```
amitaibu, line 21: maybe we should move the definition of which class to use to the plugin?

``` php
function restful_plugin_process(&$plugin, $info) {
    if ($info['type'] == 'restful') {
      $plugin += array(
        'description' => '',
        'options' => array(),
        'auth_manager' => '\RestfulAuthenticationManager',
        'cache_controller' => NULL
```

Anyway, it can be in a follow up PR

plugins/restful/RestfulEntityBase.php
```diff
@@ -905,7 +883,221 @@ public function getAccount($request = NULL) {
...
+    $camel = substr($method, 3);
+    // Make the first letter lower case.
+    $camel[0] = strtolower($camel[0]);
+    return $camel;
+  }
+
+  /**
+   * Implement the magic __call method to catch calls to undefined methods.
+   */
+  public function __call($method, $args) {
```
amitaibu: It's a cool pattern, however my concern is that it is not inline with D8's classes. I also prefer the verbose option, which is also more IDE friendly.

tests/RestfulEntityAndPropertyAccessTestCase.test
```diff
@@ -172,6 +172,8 @@ class RestfulEntityAndPropertyAccessTestCase extends DrupalWebTestCase {
     // Privileged user, with limited access to property.
     restful_test_deny_access_field();
     $handler->setAccount($user1);
+    // Remove all caches for rendered nodes.
+    $handler->cacheInvalidate('v1.0::etnode::ei');
```
amitaibu: if cache would be disabled by default will probably won't need to do it.

plugins/restful/RestfulEntityBase.php
```diff
@@ -905,7 +883,221 @@ public function getAccount($request = NULL) {
...
+   *   The cache identifier.
+   */
+  protected function generateCacheId($entity_id, $uid, $request) {
+    // Get the cache ID from the selected params. We will use a complex cache ID
+    // for smarter invalidation. The cache id will be like:
+    // v<major version>.<minor version>::et<entity type>::ei<entity id>::uu<user uid>::pa<params array>
+    // The code before every bit is a 2 letter representation of the label. For
+    // instance, the params array will be something like:
+    // fi:id,title::re:admin
+    // When the request has ?fields=id,title&restrict=admin
```
amitaibu, line 1173: Good stuff!

restful.module
amitaibu, line 385: I think this calls for a good test coverage, not only it would make it clear to me, it's also a super important point to get right. Would it be ok to ask for tests? :)

amitaibu:
Ping me when ready for review.
btw, are you in DrupalCon?

e0ipso:
No Drupalcon for me this time.

I'm working on some other cool authentication stuff for Restful, I'll
finish this one in a moment.

On Mon, Jun 2, 2014 at 8:46 AM, Amitai Burstein notifications@github.com
wrote:

> Ping me when ready for review.
> btw, are you in DrupalCon?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/Gizra/restful/pull/53#issuecomment-44806211.

## 

Mateu Aguiló Bosch.
Developer.

w. lb.cm/e0ipso
e. mateu@lullabot.com
skype. mateu.aguilo.bosch
irc. e0ipso

restful.module
e0ipso, line 401: Instead of:

``` php
->condition('cid', db_like($cid) . '%', 'LIKE')
```

We need:

``` php
->condition('cid', '%' . db_like($cid) . '%', 'LIKE')
```

plugins/restful/RestfulEntityBase.php
e0ipso: Yeah. I added the param to the signature since it is in the `viewEntity` signature. Maybe we can have a followup to fix that too.

plugins/restful/RestfulInterface.php
e0ipso, line 21: Yeah, it makes sense. Let's have that issue ready asap.

e0ipso:
@amitaibu changes made. Is there any manual way to kick off the Travis job?

amitaibu:
> Is there any manual way to kick off the Travis job?

It will probably work after you re-roll, I see there are merge conflict
added_rendered_entity_cache _by_mateu-aguilo-bosch__pull_request__53__gizra_restful

restful.module
```diff
@@ -324,3 +329,59 @@ function restful_menu_process_callback($major_version, $resource_name) {
...
+  list($entity_id,,) = entity_extract_ids($type, $entity);
+  $cid = 'et' . $type . '::ei' . $entity_id . '::uu';
+  _restful_invalidate_entity_cache($cid);
+}
+
+/**
+ * Implements hook_entity_delete().
+ */
+function restful_entity_delete($entity, $type) {
+  list($entity_id,,) = entity_extract_ids($type, $entity);
```
amitaibu: => `list($entity_id)`

README.md
```diff
@@ -203,6 +203,34 @@ Will result with an HTTP code 400, and the following JSON:
...
+cache layer. This means that no entity will need to be loaded if it was rendered
+in the past under the same conditions.
+
+Developers have absolute control where the cache is stored and the expiration
+for every resource, meaning that very volatile resources can skip cache entirely
+while other resources can have its cache in MemCached or the database. To
+configure this developers just have to specify the following keys in their
+_restful_ plugin definition:
+
+  - `cache_render`: Set it to `FALSE` to disable the render cache completely
```
amitaibu: I think this can be moved as PHPdocs under `restful_plugin_process()` (I'd like to keep the README not too detailed, and just give an overview of the features/ concepts)

plugins/restful/RestfulEntityBase.php
```diff
@@ -905,7 +893,250 @@ public function getAccount($request = NULL) {
...
+  public function getEntityType() {
+    return $this->entityType;
+  }
+
+  /**
+   * Setter for $plugin.
+   *
+   * @param array $plugin
+   */
+  public function setPlugin($plugin) {
```
amitaibu: do we need this get/set?

plugins/restful/RestfulEntityBase.php
```diff
@@ -905,7 +893,250 @@ public function getAccount($request = NULL) {
...
+  public function getAuthenticationManager() {
+    return $this->authenticationManager;
+  }
+
+  /**
+   * Setter for $bundle.
+   *
+   * @param string $bundle
+   */
+  public function setBundle($bundle) {
```
amitaibu: is this used? I don't think we need it, as we have the info in `$this->plugin['bundle']`

plugins/restful/RestfulInterface.php
```diff
@@ -41,16 +43,6 @@ public function process($path = '', $request = NULL, $method = 'get');
...
 
   /**
-   * Return array keyed by the header property, and the value.
-   *
-   * This can be used for example to change the "Status" code of the HTTP
-   * response, or to add a "Location" property.
-   *
-   * @return array
-   */
-  public function getHttpHeaders();
```
amitaibu, line 51: why remove form the interface?

plugins/restful/RestfulEntityBase.php
e0ipso: I made PHPStorm generate all the getters and setters, just in case we ever need them.

plugins/restful/RestfulInterface.php
e0ipso, line 51: > why remove form the interface?

This was from back when I moved to magic methods.

e0ipso:
> I see there are merge conflict

Only people with merge permissions can see that there are merge conflicts. :-1:

e0ipso:
@amitaibu this should be good to merge now.

BTW: I re-enabled the UpdateEntity tests and fixed them.

amitaibu:
Did minor cleanup and merged, thanks!

PR #57, opened 2014-06-05
Oauth (Part I)

e0ipso:
This is a **FULLY** decoupled approach to an OAuth authentication/authorization workflow. See how other alternatives require the drupal backend to be accessible by end users in order to:
- Register an application key & secret.
- Defining authorization levels.
- Defining authorization contexts. Just a bunch of authorization levels.
- Loging in the user to grant access to a token.

They're also tightly coupled to Form API :wink: 

This approach assumes that there is no possible interaction to the Drupal installation beyond the RESTful API. Only the user1 can have access to it.

Also, everything is implemented as entities and therefore it can be easily turned into an endpoint. This way an implementor can easily build an administration frontend (if needed).

Since this is going to be a big PR, I want to break it in different parts.


e0ipso:
The current code goes as far as to [getting a token request](http://oauth.net/core/1.0a/#request_grant) in the OAuth flow.

amitaibu:
@bojanz this is the part you step in! :)

bojanz:
Oh, so you wrote an entire OAuth implementation.

My idea was to provide a way to integrate oauth2_server (which is just a Drupal UI around the oauth2-server-php library), since my APIs all use OAuth 2.

amitaibu:
It would indeed be nicer if we'd leverage another module and reduce the code

e0ipso:
Actually all this does is leveraging the PECL OAuth library to check tokens.

Everything else is declaring a couple of entity types and 2 RESTful endpoints to interact with them. All the OAuth work is done by the external library.

All that other modules provide are custom tables. Then we would need to wrap around into an entity, and provide an endpoint to interact with them. Also we would have a lot of surrounding unused code for UI and forms that we **should** not be using anyway.

The code is focused on OAuth 1.0a so far, but AFAIK the library can handle 2LeggedAuthentication and complex workflows. Besides, if there is a crazy use case since our code is based on plugins and object oriented polymorphism anyone can create a version –or fork– on our `request_token` endpoint and just override. According to http://oauth.net/code/

> There is an extension to php that supports OAuth. It was written by John Jawed. An example can be found in http://php.net/oauth and a walkthrough is also available. This pecl package is considered the de facto standard by Rasmus Lerdorf.

The only non-standar bit is the Cipher class, that will add an extra bit of security for replay injection attacks making token requests only valid for a natural day –besides request signature and timestamp limitations of 2h–. The nonce endpoint is needed to keep the hash keys on the server. This part could go away, but it makes me feel safe.

Maybe this can be moved to another contrib, and have an OAuth2 implementation (which has some security concerns).

amitaibu:
If I understand correctly @bojanz recommend Oauth2 for decoupled apps.
@bojanz maybe you can comment here + are you going to work on an integration with Oauth2?

bojanz:
I've just tagged oauth2_server 1.1.
Integrating it with this module is next on my todo list.
It should be enough to call oauth2_server_verify_access() at the right moment.

OAuth2 gets a bit of a bad reputation because the specification makes security optional (the presence of a state parameter, the validation of the redirect_uri parameter that is behind 99% of published OAuth2 exploits, etc). However, our library (oauth2-server-php) and module (oauth2_server) take a strict approach to security, with strict requirements, safe defaults and documentation for possible pitfalls. 
It is possible to write an insecure login implementation using OAuth2 (which doesn't really concern this module), which is why we also support OpenID Connect, a thin layer on top of OAuth2 (endpoint for UserInfo + additional id_token).

I do prefer OAuth2 because it supports a wider range of use cases. 
We use:
- the implicit flow for decoupled apps (authenticating our AngularJS app - platform.sh)
- The password grant type for our CLI app.
- The authorization code flow and client credentials grant types for two different APIs
- The authorization code flow with OpenID Connect enabled for our single sign on solution (Drupal sites).
  That's Drupal, AngularJS, Symfony Console, and various APIs, all using the same codebase. 

Of course, that doesn't mean I'm against the efforts in this PR. Just providing my point of view.

amitaibu:
@mateu-aguilo-bosch I personally would prefer to have the Oauth2, because it means Bojanz will have to maintain it ;)

What do you think? Should we provide both, or would Oauth2 be enough - I'd prefer to select the one we'd consider as "best practice" and that can "simply work".

e0ipso:
I'm OK dropping this. The only thing I regret is some buried improvements that I had made along this PR, but I'm sure we'll get them eventually.

e0ipso:
@amitaibu please, don't remove the branch.

PR #67, opened 2014-06-12
Create better unique token

snirs90:
This creates a better token with more security of the token and more reliable for unique token.

### DIFF
--- a/modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/1.0/RestfulTokenAuthentication.class.php
+++ b/modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/1.0/RestfulTokenAuthentication.class.php
@@ -8,8 +8,8 @@
 class RestfulTokenAuthentication extends \RestfulEntityBase {
 
   /**
-   * A secret token prefix for better security and uniqueness
-   * .
+   * A secret token prefix for better security and uniqueness.
+   * 
    * @var string
    */
   protected $secretTokenPrefix = 'A secret token prefix for better encryption Ax%@GFbZS3a@';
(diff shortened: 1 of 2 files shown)


modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/1.0/RestfulTokenAuthentication.class.php
```diff
@@ -69,7 +76,7 @@ public function getOrCreateToken($request = NULL, stdClass $account = NULL) {
         'type' => 'restful_token_auth',
         'created' => REQUEST_TIME,
         'name' => 'self',
-        'token' => md5(time()),
+        'token' => $this->createToken($account),
```
amitaibu: we can probably use `drupal_random_key()` instead.

modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/1.0/RestfulTokenAuthentication.class.php
```diff
@@ -78,4 +85,22 @@ public function getOrCreateToken($request = NULL, stdClass $account = NULL) {
...
+   * @param stdClass $account
+   *  The user account object.
+   *
+   * @return string
+   *  Return a unique token.
+   */
+  protected function createToken(stdClass $account = NULL) {
+    $unique = uniqid(NULL, TRUE);
+
+    $token = sha1($this->secretTokenPrefix) . md5(sha1($account->uid) . time()) . sha1($unique);
```
e0ipso: I know that every token generation is different, but if we're going to use SHA we probably want to use `drupal_hmac_base64` to ensure encoding compatibility.

``` php
$uid = $account ? $account->uid : '';
return drupal_hmac_base64($account->uid . drupal_get_hash_salt(), $unique);
```

Also note that since the default for `$account` is `NULL` we need extra checks. We can also dump the `secretTokenPrefix` in favor of Drupal's salt.

amitaibu:
Merged, thanks.

PR #68, opened 2014-06-13
Login user and respond with JSON along with auth cookie

amitaibu:


amitaibu:
Still needs tests.

amitaibu:
@mateu-aguilo-bosch any objection? :)

amitaibu:
fyi, I've started working on [this](https://github.com/Gizra/angular-restful-auth) Angular example app (not making it too much public yet, as I want to style it + write a blog post about it)

And it's actually already working! :)

e0ipso:
Count me in on that too.

amitaibu:
> Count me in on that too.

You mean on the Angular app? 

Also, I believe this PR is ready to ship

amitaibu:
Merged.

plugins/restful/user/login_cookie/1.0/login_cookie__1_0.inc
```diff
@@ -0,0 +1,15 @@
+<?php
+
+$plugin = array(
+  'label' => t('Login'),
+  'description' => t('Login a user and return a JSON along with the authentication cookie..'),
```
e0ipso, line 5: Extra trailing '.'

e0ipso:
Sorry for commenting late, but I did not have the time earlier.

I don't see any tests for `login_cookie`, is that intentional?

e0ipso:
Also, maybe it deserves a little documentation on how to set the cookie jar to capture the returned cookie. Perhaps an example using curl.

amitaibu:
> I don't see any tests for login_cookie, is that intentional?

No, just an oversight, I'll write ones soon.

> Perhaps an example using curl.

I actually didn't check with CURL, but it's a good point. Maybe also point in the readme to the Angular app.

e0ipso:
I like curl because you can brainlessly c&p to your terminal, but the Angular app is a real world case scenario. Up to you :smiley:

Issue #69, opened 2014-06-13
When access is denied restful_json_output() gets wrong output


amitaibu:
If the access callback of the menu item returns FALSE, then the output would be 3 (which is actually `MENU_ACCESS_DENIED`) with status code 200.

We should call `drupal_deliver_html_page()` in case of int values.

PR #70, opened 2014-06-13
Properly handle integer menu status constant in restful_json_output()

amitaibu:
#69


amitaibu:
Actually, instead of calling `drupal_deliver_html_page()` we should return a proper json.

amitaibu:
Ok, this is now converted to JSON, and sends correct response for `MENU_ACCESS_DENIED` and `MENU_SITE_OFFLINE`

Still needs tests.

amitaibu:
Merged.

PR #72, opened 2014-06-18
Add wrapper object as a param to the callback property

snirs90:
Now the callback property function get the wrapper object to it can be used inside the callback.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -50,7 +50,7 @@ abstract class RestfulEntityBase implements RestfulEntityInterface {
    *    (e.g. $wrapper->field_reference->label()). Defaults to FALSE.
    *  - "callback": A callable callback to get a computed value. Defaults To
    *    FALSE.
-   *    The callback function receive as first argument the node
+   *    The callback function receive as first argument the entity
    *    EntityMetadataWrapper object.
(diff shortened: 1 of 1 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -583,7 +583,7 @@ public function viewEntity($entity_id, $request, stdClass $account) {
           throw new Exception(format_string('Process callback function: @callback does not exists.', array('@callback' => $callback_name)));
         }
 
-        $value = call_user_func($info['callback']);
+        $value = call_user_func($info['callback'], $wrapper);
```
amitaibu, line 588: Add the info that we pass the wrapper [here](https://github.com/Gizra/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L51)

snirs90, line 588: Added

plugins/restful/RestfulEntityBase.php
```diff
@@ -50,6 +50,8 @@
    *    (e.g. $wrapper->field_reference->label()). Defaults to FALSE.
    *  - "callback": A callable callback to get a computed value. Defaults To
    *    FALSE.
+   *    The callback function receive as first argument the node
```
amitaibu: node => entity

amitaibu:
Merged (with correct credit).

PR #74, opened 2014-06-18
Fix setting values for entity reference and text fields on POST

snirs90:
When creating a new entity via POST request, now the fields of entity reference type with multiple values is supported.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -881,6 +881,13 @@ abstract class RestfulEntityBase implements RestfulEntityInterface {
     $original_request = $request;
 
     foreach ($this->getPublicFields() as $public_property => $info) {
+      if (empty($info['property'])) {
+        // We may have for example an entity with no label property, but with a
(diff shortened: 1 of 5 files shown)


amitaibu:
@snirs90 please continue with the tests in your daily contribution hour.

plugins/restful/RestfulEntityBase.php
```diff
@@ -912,6 +924,64 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, $
...
+      if ($field_info['cardinality'] == 1) {
+        // Single value.
+        if (!$instance['settings']['text_processing']) {
+          return $value;
+        }
+
+        return array (
+          'value' => $value,
+          // @todo: How to get the correct format value?
+          'format' => 'filtered_html',
```
amitaibu, line 960: @mateu-aguilo-bosch do you have a better idea on what to do here? We could move this info into the getPublicFields(), but the format may change from user to user.

amitaibu:
Added #86 for follow up, and since it's fixing WSOD on POST and comes with tests -- merged.

PR #78, opened 2014-06-19
Allow using all auth types by default, and also making authentication optional

amitaibu:


plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -14,6 +14,37 @@ class RestfulAuthenticationManager extends \ArrayObject {
...
+   *
+   * @param boolean $is_optional
+   *   Determines if the authentication is optional.
+   */
+  public function setIsOptional($is_optional) {
+    $this->isOptional = $is_optional;
+  }
+
+  /**
+   * Set the authentications' "optional" flag.
```
snirs90: Set => Get

amitaibu:
@mateu-aguilo-bosch in case you missed the PR ^^ :)

restful.module
```diff
@@ -263,11 +269,23 @@ function restful_get_restful_handler_by_name($plugin_name) {
...
-  foreach ($plugin['authentication_types'] as $auth_plugin_name) {
+  foreach ($auth_plugins as $auth_plugin_name) {
     $auth_handler = restful_get_authentication_handler($auth_plugin_name);
     $handler->getAuthenticationManager()->addAuthenticationProvider($auth_handler);
+    $has_auth = TRUE;
   }
+
+  if ($has_auth) {
+    // Set the "optional" flag of the authentication manager.
+    $handler->getAuthenticationManager()->setIsOptional($plugin['authentication_optional']);
```
e0ipso: Can we refactor all the `$has_auth` logic into:

``` php
if (!empty($auth_plugins)) {
  // Set the "optional" flag of the authentication manager.
  $handler->getAuthenticationManager()->setIsOptional($plugin['authentication_optional']);
}
```

But I don't get why we need the check, couldn't we just set it anyways? Maybe even by the authentication manager constructor and avoid the extra logic?

e0ipso:
I was missing this feature just today! One minor detail, other than that this is a go.

amitaibu:
Merged.

PR #71, opened 2014-06-23
Implement rate-limiting

e0ipso:
http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#rate-limiting


e0ipso:
This PR will introduce the rate limits to the restful endpoints. You have to define the rate limits per role at the plugin level.

Also, there is the concept of _event_ since you may want to limit the API usage on costy operations differently. For instance you could have a _posting_ event to limit the API usage for authenticated users when using the POST, PATCH and PUT HTTP methods. Events are plugins declaring a class name that implements a method called `isRequestedEvent`. A default `rate_limit` plugin is shipped for the basic request (event that is true for every request).

The rate limits are non-fieldable entities for operation convenience with a doped entity class for common entity operations.

For a follow-up: Improve documentation (maybe an advanced_help page) about how to use the rate limit configuration array in the restful plugin. Also documentation about how to create a new event can be added.

modules/restful_example/plugins/restful/node/articles/1.4/articles__1_4.inc
```diff
@@ -0,0 +1,26 @@
...
+  'minor_version' => 4,
+  'rate_limit' => array(
+    // The 'request' event is the basic event. You can declare your own events.
+    'request' => array(
+      'event' => 'request',
+      // Rate limit is cleared every day.
+      'period' => new \DateInterval('P1D'),
+      'limits' => array(
+        'authenticated user' => 3,
+        'anonymous user' => 2,
```
amitaibu, line 21: Maybe as a follow up we could move the flood control of the [base_auth](https://github.com/Gizra/restful/blob/b6c9419814210bd1386ce069aa83ca00f4f3478d/plugins/authentication/RestfulAuthenticationBasic.class.php#L31) to use rate-limit, so we consolidate it to a single place

modules/restful_example/plugins/restful/node/articles/1.4/articles__1_4.inc
```diff
@@ -0,0 +1,26 @@
...
+        'administrator' => \RestfulRateLimitManager::INFINITE_RATE_LIMIT,
```
amitaibu: minor but maybe INFINITE => UNLIMITED

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+        if ($rate_limit_entity->hits >= $limit) {
+          throw new \RestfulFloodException('Rate limit reached');
+        }
+      }
+      // Save a new hit after generating the exception to mitigate DoS attacks.
+      $rate_limit_entity->hit();
+
+      // Add the limit headers to the response.
+      $remaining = $limit == static::INFINITE_RATE_LIMIT ? 'unlimited' : $limit - ($rate_limit_entity->hits + 1);
+      drupal_add_http_header('X-Rate-Limit-Limit', $limit, TRUE);
```
amitaibu, line 138: not totally sure about this, but maybe we should inject the restful handler into the class. So get/set account would be `$this->getRestfulHandler()->getAuthenticationManager...`

Advantage: is that we could use  `$this->getRestfulHandler()->addHttpHeader()` to set the header instead of directly calling `drupal_add_http_header()`

Disadvantage: More coupled with the Restful handler, although I can't think of a use case where the rate limit would be by itself.

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+   * Set the account.
+   *
+   * @param \stdClass $account
+   */
+  public function setAccount($account) {
+    $this->account = $account;
+  }
+
+  /**
+   * @param string $resource
```
amitaibu, line 58: missing one liner

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+   *   Resource name being checked.
+   * @param array $plugin_info
+   *   The plugin info array for the rate limit.
+   * @param \stdClass $account
+   *   The identified user account for the request.
+   */
+  public function __construct($resource, $plugin_info, $account = NULL) {
```
amitaibu, line 65: I don't think account is really NULL if called from the restful handler, it's always going to be anon or auth.

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+   * The identified user account for the request.
+   * @var \stdClass
+   */
+  protected $account;
+
+  /**
+   * The plugin info array for the rate limit.
+   * @var array
+   */
+  protected $pluginInfo;
```
amitaibu, line 20: I think it's worth adding getter/ setter

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+   */
+  public function checkRateLimit($request) {
+    $now = new \DateTime();
+    $now->setTimestamp(REQUEST_TIME);
+    // Check all rate limits configured for this handler.
+    foreach ($this->pluginInfo as $event_name => $info) {
+      // If the limit is infinite then skip everything.
+      $limit = $this->rateLimit($info);
+      if ($limit == static::INFINITE_RATE_LIMIT) {
+        continue;
```
amitaibu, line 94: Add comment why we continue. (i.e. `User has unlimited access to the resources.`)

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+      }
+      // Check if there is a rate_limit plugin for the event.
+      // There are no error checks on purpose, let the exceptions bubble up.
+      $rate_limit_plugin = restful_get_rate_limit_plugin($info['event']);
+      $rate_limit_class = ctools_plugin_get_class($rate_limit_plugin, 'class');
+      /** @var \RestfulRateLimitInterface $rl */
```
amitaibu: remove.

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+      $rl = new $rate_limit_class();
```
amitaibu: maybe $rl => $handler?

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+      if (!$rate_limit_entity = $this->loadRateLimitEntity($event_name)) {
+        // If there is no entity, then create one.
+        $rate_limit_entity = entity_create('rate_limit', array(
+          'timestamp' => REQUEST_TIME,
+          'expiration' => $now->add($info['period'])->format('U'),
+          'hits' => 0,
+          'event' => $event_name,
+          'identifier' => $this->generateIdentifier(),
+        ));
+        // We don't need to save it since it will be saved upon hit.
```
amitaibu: move this comment above `$rate_limit_entity = entity_create..`

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+  /**
+   * Gets the limit for the identified user.
+   *
+   * @param array $event_plugin_info
+   *   The array containing the plugin info for the current event.
+   *
+   * @return int
+   *   The limit.
+   */
+  protected function rateLimit($event_plugin_info) {
```
amitaibu, line 190: => getRateLimit
event_plugin_info => $plugin_info

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+    // If the user is anonymous.
+    if (empty($this->account->roles)) {
```
amitaibu, line 192: I think better check $this->account->uid

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+      return empty($event_plugin_info['limits']['anonymous user']) ? 0 : $event_plugin_info['limits']['anonymous user'];
```
amitaibu: we could assign default `$event_plugin_info['limits']['anonymous user'] = 0` on the plugin process to avoid doing it here.

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -0,0 +1,191 @@
...
+    }
+    // If the user is logged then return the best limit for all the roles the
+    // user has.
+    $max_limit = 0;
+    foreach ($this->account->roles as $rid => $role) {
+      if (
+        isset($event_plugin_info['limits'][$role]) &&
```
amitaibu: for readability lets break these lines to smaller ones that we can return (continue) early, we a comment above each one to explain the logic.

plugins/rate_limit/RestfulRateLimitRequest.class.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulRateLimitRequest.
+ */
+
+class RestfulRateLimitRequest implements RestfulRateLimitInterface {
```
amitaibu, line 8: => `RestfulRateLimitRequestBase` ?

plugins/restful/RestfulEntityBase.php
```diff
@@ -122,6 +122,13 @@
   protected $authenticationManager;
 
   /**
+   * Rate limit manager.
+   *
+   * @var \RestfulRateLimitManager
+   */
+  protected $limitManager;
```
amitaibu: => `$rateLimitManager`

plugins/rate_limit/request.inc
```diff
@@ -0,0 +1,8 @@
+<?php
+
+$plugin = array(
+  'label' => t('Request'),
```
amitaibu: => `Any request`

plugins/rate_limit/request.inc
```diff
@@ -0,0 +1,8 @@
...
+  'description' => t('The basic rate limit plugin. It is always met.'),
```
amitaibu: Maybe `The basic rate limit plugin. Every call to a resource is counted.`?

plugins/restful/RestfulEntityBase.php
amitaibu: - setter/ getter

plugins/restful/RestfulEntityBase.php
```diff
@@ -1025,7 +1039,13 @@ public function getPluginInfo($key = NULL) {
    *   The user object.
    */
   public function getAccount($request = NULL) {
-    return $this->getAuthenticationManager()->getAccount($request);
+    $account = $this->getAuthenticationManager()->getAccount($request);
+
+    // If the limit rate is enabled for the current plugin then set the account.
+    if (!empty($this->limitManager)) {
+      $this->limitManager->setAccount($account);
```
amitaibu: this should also happen in setAccount(), unless we'll inject the restful handler to the rate limit manager, so we'll always deal with the same user.

restful.info
```diff
@@ -8,12 +8,17 @@ dependencies[] = entity
...
+files[] = plugins/rate_limit/RestfulRateLimitInterface.php
+files[] = plugins/rate_limit/RestfulRateLimitManager.php
 files[] = plugins/restful/RestfulEntityBase.php
 files[] = plugins/restful/RestfulEntityBaseNode.php
 files[] = plugins/restful/RestfulEntityBaseMultipleBundles.php
 files[] = plugins/restful/RestfulEntityInterface.php
 files[] = plugins/restful/RestfulInterface.php
 
+; Includes
+files[] = includes/RestfulRateLimit.php
```
amitaibu, line 20: I think it should be like the others under `plugins/rate_limit`

restful.module
```diff
@@ -54,12 +54,32 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
  *   responsible to invalidate caches in this scenario. Defaults to TRUE.
+ * - rate_limit: Associative array keyed by event name. Every item contains the
+ *   rate limiting options.
  *
  * Properties for the "authentication" plugin type:
  * - description: The description of the authentication provider. Defaults to
  *   empty string.
  * - settings: Array with the settings needed for the plugin. Defaults to empty
  *   array.
+ *
```
amitaibu, line 65: as a follow up PR, we should probably create a plugin process function for each plugin type. Would make doc look nicer.

restful.module
```diff
@@ -54,12 +54,32 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
+ * Properties for every item of the rate_limit:
+ *   - period: A \DateInterval object representing the period on which the rate
```
amitaibu, line 67: 'event' property is missing

restful.module
```diff
@@ -453,6 +514,47 @@ function _restful_get_json_from_menu_status(&$var) {
...
+/**
+ * Implements hook_entity_info().
+ */
+function restful_entity_info() {
+  $items['rate_limit'] = array(
+    'label' => t('Rate limit'),
+    'entity class' => 'RestfulRateLimit',
+    'controller class' => 'EntityAPIController',
+    'base table' => 'restful_rate_limit',
+    'fieldable' => FALSE,
```
amitaibu: is this for performance reasons?

tests/RestfulRateLimitTestCase.test
```diff
@@ -0,0 +1,83 @@
...
+      ->propertyCondition('identifier', 'articles::' . $account->uid)
+      ->execute();
+    $rl = entity_load_single('rate_limit', key($results['rate_limit']));
+    $rl->timestamp = REQUEST_TIME - 2;
+    $rl->expiration = REQUEST_TIME - 1;
+    $rl->save();
+    $this->roleExecute($account, 3);
+  }
+
+  protected function roleExecute($account, $limit) {
```
amitaibu, line 69: PHPdocs.

tests/RestfulRateLimitTestCase.test
```diff
@@ -0,0 +1,83 @@
...
+      'efg',
+    );
+    foreach ($titles as $title) {
+      $settings['title'] = $title;
+      $node = $this->drupalCreateNode($settings);
+      $nodes[$title] = $node->nid;
+    }
+  }
+
+  public function testLimits() {
```
amitaibu, line 37: PHPdocs.

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso, line 65: The problem is if we instantiate the class before calling the authentication providers. We shouldn't, but there is a chance.

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso, line 94: There was a comment about that 2 lines above

> // If the limit is infinite then skip everything.

Anyways, it won't hurt.

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso: You don't like IDE hinting :smile:

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso, line 138: I don't see the real value on that. `RestfulBaseEntity::setHttpHeader()` will only store key-values in an array to ultimately use `drupal_add_http_header()` outside the scope of the class (in restful.module). So the only thing we would be reusing is the storage array.

> Disadvantage: More coupled with the Restful handler, although I can't think of a use case where the rate limit would be by itself.

But we may want the rate limit to apply to different restful handlers without having to maintain one inner restful plugin.

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso, line 190: > event_plugin_info => $plugin_info

I'd rather keep it as `$event_plugin_info`. Actually I had `$plugin_info` in my code and changed it to avoid confusion with the _rate_limit_ plugin info.

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso, line 192: `empty($this->account->roles` is an early return for error check on [L#177](https://github.com/Gizra/restful/pull/71/files#diff-55e7308550589d57aff8649137fc29ecR177).

plugins/rate_limit/RestfulRateLimitRequest.class.php
e0ipso, line 8: I was thinking of `RestfulRateLimitRequest` as `<ModuelNamespace><PluginType><EventName>`. I'd rather see

``` php
class RestfulRateLimitPosting implements RestfulRateLimitInterface { … }
```

than

``` php
class RestfulRateLimitPosting extends RestfulRateLimitEventBase { … }
```

Since `RestfulRateLimitEventBase` as simple as it would almost be duplicating the interface.

restful.info
e0ipso, line 20: I think it's confusing to have entity classes in the plugins directory.

e0ipso, line 20: Actually I see that the `restful_token_auth` module also uses the _includes_ approach.

restful.module
e0ipso: No strong feelings about this. I think entity cache will do a good job if someone gets crazy and starts adding fields to this entity. I just could not imagine a use case where custom fields could be useful.

modules/restful_example/plugins/restful/node/articles/1.4/articles__1_4.inc
e0ipso, line 21: Done.

restful.info
amitaibu, line 20: oh sorry, I missed that it's the Entity class.

restful.module
```diff
@@ -399,6 +438,34 @@ function restful_menu_process_callback($major_version, $resource_name) {
 }
 
 /**
+ * @return array
```
amitaibu, line 445: missing docs.

e0ipso:
@amitaibu, tests are back to green.

amitaibu:
Good stuff! Merged, thanks.

amitaibu:
Oh no, I've committed by mistake without proper credit :(
Let me see if I can fix that...

amitaibu:
ok, fixed :)

PR #87, opened 2014-06-25
Allow changing the entity just before it's saved via POST

amitaibu:
For example, setting the author of the node entity.


amitaibu:
Tests are green.

PR #89, opened 2014-06-25
Add support to attach a file to a field with file id

snirs90:
Because we are uploading files via the API of "file-upload" end-point, we are receiving the file ID,
so we need to attach a file into a file field with the file id that recieved.


snirs90:
Note:
I think we need to add support for passing the file object(s) as well and not just the file IDs
when we want to use programmatically and pass an array of file object(s).
or we can just leave it like this and the user will must to pass the file ID(s) when using programmatically as well.
@amitaibu what do you say?

snirs90:
Closed in favor of #90

PR #90, opened 2014-06-25
Fix creation of different field types

amitaibu:
Also split preprocess per field type methods, for POST - this would make it easier to extend according to field types.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -922,7 +922,7 @@ abstract class RestfulEntityBase implements RestfulEntityInterface {
 
     // Allow changing the entity just before it's saved. For example, setting
     // the author of the node entity.
-    $this->createEntityPreInsert($wrapper->value(), $request, $account);
+    $this->entityPreSave($wrapper->value(), $request, $account);
 
     $wrapper->save();
   }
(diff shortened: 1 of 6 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -926,58 +926,130 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, $
...
+   * @param array $field_info
+   *   The field info array.
+   *
+   * @return mix
+   *   The value to set using the wrapped property.
+   */
+  protected function propertyValuesPreprocessFile($property_name, $value, $field_info) {
+    if ($field_info['cardinality'] == 1) {
+      // Single value.
+      return array('fid' => $value);
```
snirs90: As far as I've checked and tested, Need to add "display" property.
`display => 1`
if it's not set, it's throws an SQL error:
`PDOException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'field_file_display' cannot be null`

Or maybe I'm doing something wrong? (tried without display property):

``` php
$wrapper = entity_metadata_wrapper('node', 1);
$wrapper->field_file->set(array('fid' => 2));
$wrapper->save();
```

plugins/restful/RestfulEntityBase.php
```diff
@@ -926,58 +926,130 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, $
...
     }
 
-    // Return the value as is.
-    return $value;
+    $return = array();
+    foreach ($value as $delta => $single_value) {
+      $return[$delta]['fid'] = $single_value;
```
snirs90: Same here about the display property:
`$return[$delta]['display'] = 1;`

plugins/restful/RestfulEntityBase.php
```diff
@@ -926,58 +926,130 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, $
   /**
    * Massage the value to set according to the format expected by the wrapper.
    *
-   * @param $property_name
+   * @param string  $property_name
```
snirs90: One space too much after `string`

plugins/restful/RestfulEntityBase.php
```diff
@@ -926,58 +926,130 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, $
...
+  protected function propertyValuesPreprocessFile($property_name, $value, $field_info) {
+    if ($field_info['cardinality'] == 1) {
+      // Single value.
+      return array('fid' => $value);
     }
 
-    // Return the value as is.
-    return $value;
+    $return = array();
+    foreach ($value as $delta => $single_value) {
```
snirs90, line 1057: I think we should add `explode()` to commas same as in entity-reference case if we want to pass few file IDs with comma separated through a POST request.

amitaibu:
POST now works (+tested) for text fields (wit and without text-processing), entity reference, taxonomy term reference, file field and image field.

amitaibu:
@mateu-aguilo-bosch ^^

amitaibu:
Merged.

Issue #92, opened 2014-06-29
Validate the entity with the entity validate module 


RoySegall:
We already using the [Entity validate](https://github.com/gizra/entity_validator) module in OpenScholar. I think this can be good addition. Dependent PR: https://github.com/Gizra/entity_validator/pull/8

PR #93, opened 2014-06-29
Using entity validator for validate the entity before saving it to the DB.

RoySegall:
https://github.com/Gizra/restful/issues/92


amitaibu:
Closed in favor of #94

PR #94, opened 2014-06-30
Integrate with entity validator module

amitaibu:


amitaibu:
@mateu-aguilo-bosch this integration allows us to move away from Drupal forms into Angular.

plugins/restful/RestfulEntityBase.php
```diff
@@ -924,6 +924,8 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $request, s
     // the author of the node entity.
     $this->entityPreSave($wrapper->value(), $request, $account);
 
+    $this->entityValidate($wrapper);
```
amitaibu, line 927: @mateu-aguilo-bosch we are passing the wrapped entity here for convenience. I'd like to do the same for the above `$this->entityPreSave()` -- makes sense?

amitaibu, line 927: opened a follow up issue -- #97

amitaibu:
Merged.

Issue #98, opened 2014-07-03
CSRF security vulnerabilty on write operations with session cookies


klausi:
Since this module does not have a stable release on drupal.org yet I think it is fine to report security issues in public (see our policy https://www.drupal.org/security-advisory-policy ).

I have not looked to closely at the code and have not tried to exploit this yet, but I think there could be a CSRF vulnerability on write operations such as POST, PUT and DELETE requests when using session cookie authentication.

This happened in the past to other Drupal REST server solutions:
RESTWS: https://www.drupal.org/node/1890222 and http://cgit.drupalcode.org/restws/commit/?id=dae7e2e
Services: https://www.drupal.org/node/2012982
Drupal 8 core: https://www.drupal.org/node/1891052

Maybe I overlooked something, do you already handle this case somehow?

amitaibu:
Thanks @klausi indeed we are not doing this yet. Started work on #99

amitaibu:
Merged, thanks @klausi :)

PR #99, opened 2014-07-04
Add CSRF token for write operations

amitaibu:
#98

### DIFF
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -9,7 +9,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function applies($request = NULL) {
+  public function applies($request = NULL, $method = 'get') {
     $key_name = !empty($this->plugin['options']['param_name']) ? $this->plugin['options']['param_name'] : 'access_token';
     return !empty($request[$key_name]);
   }
@@ -17,7 +17,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function authenticate($request = array()) {
+  public function authenticate($request = NULL, $method = 'get') {
     $key_name = !empty($this->plugin['options']['param_name']) ? $this->plugin['options']['param_name'] : 'access_token';
 
     // Check if there is a token that did not expire yet.
--- a/plugins/authentication/RestfulAuthenticationCookie.class.php
+++ b/plugins/authentication/RestfulAuthenticationCookie.class.php
@@ -9,12 +9,30 @@ class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements R
   /**
    * Implements RestfulAuthenticationInterface::authenticate().
    */
-  public function authenticate($request = NULL) {
+  public function authenticate($request = NULL, $method = 'get') {
     if (!drupal_session_started() && !$this->isCli()) {
       return;
     }
+
     global $user;
-    return user_load($user->uid);
+    $account = user_load($user->uid);
+
+    if (in_array($method, array('get', 'head', 'options', 'trace')) || empty($request['__application']['rest_call'])) {
+      // Request is done via API not CURL, or not a write operation, so we don't
+      // need to check for a CSRF token.
+      return $account;
+    }
+
+    if (empty($request['__application']['csrf_token'])) {
+      throw new \RestfulBadRequestException('No CSRF token passed in the HTTP header.');
+    }
+
+    if (!drupal_valid_token($request['__application']['csrf_token'], 'rest')) {
+      throw new \RestfulForbiddenException('CSRF token validation failed.');
+    }
+
+    // CSRF validation passed.
+    return $account;
   }
 
   /**
--- a/restful.module
+++ b/restful.module
@@ -24,7 +24,7 @@ function restful_ctools_plugin_directory($module, $plugin) {
  *   "articles" resource will result with "api/v2/articles" as the URL. Defaults
  *   to 1.
  * - minor_version: The minor version of the resource. Setting the minor version
- *   via CURL is done by setting HTTP_RESTFUL_MINOR_VERSION in the HTTP headers.
+ *   via CURL is done by setting HTTP_X_RESTFUL_MINOR_VERSION in the HTTP headers.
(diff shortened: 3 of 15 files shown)


modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
```diff
@@ -9,7 +9,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function applies($request = NULL) {
+  public function applies($request = NULL, $method = 'get') {
```
e0ipso, line 12: Maybe we could start referencing to methods by a constant to avoid case edge cases.

plugins/authentication/RestfulAuthenticationCookie.class.php
```diff
@@ -9,12 +9,30 @@ class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements R
...
+      // Request is done via API not CURL, or not a write operation, so we don't
+      // need to check for a CSRF token.
+      return $account;
+    }
+
+    if (empty($request['__application']['csrf_token'])) {
+      throw new \RestfulBadRequestException('No CSRF token passed in the HTTP header.');
+    }
+
+    if (!drupal_valid_token($request['__application']['csrf_token'], 'rest')) {
```
e0ipso, line 30:
``` php
drupal_valid_token($request['__application']['csrf_token'], 'rest')
```

Lets get rid of the magic string `'rest'`, extract it somewhere (possibly in a constant) and document it.

restful.module
```diff
@@ -426,7 +432,7 @@ function restful_menu_access_callback($major_version, $resource_name) {
  */
 function restful_menu_process_callback($major_version, $resource_name) {
   $major_version = intval(str_replace('v', '', $major_version));
-  $minor_version = !empty($_SERVER['HTTP_RESTFUL_MINOR_VERSION']) && is_numeric($_SERVER['HTTP_RESTFUL_MINOR_VERSION']) ? $_SERVER['HTTP_RESTFUL_MINOR_VERSION'] : 0;
+  $minor_version = !empty($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) && is_numeric($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) ? $_SERVER['HTTP_X_RESTFUL_MINOR_VERSION'] : 0;
```
e0ipso, line 436: Does this work? `$_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']` should be a string (?) (even if it's a numeric string) and `is_int("3")` will return `FALSE`.

restful.module
```diff
@@ -677,3 +686,12 @@ function restful_file_upload_access() {
...
   return variable_get('restful_file_upload_allow_anonymous_user', FALSE) || $user->uid;
 }
+
+/**
+ * Page callback: returns a session token for the currently active user.
+ */
+function restful_csrf_session_token() {
+  drupal_add_http_header('Content-Type', 'text/plain');
+  print drupal_get_token('rest');
+  drupal_exit();
```
e0ipso: Is there a reason not to use the `restful_json_output` delivery callback and do:

``` php
function restful_csrf_session_token() {
  drupal_add_http_header('Content-Type', 'text/plain');
  return drupal_get_token('rest');
}
```

It seems more consistent with the rest of the module's output. It will add some wrapping `"` though.

tests/RestfulCsrfTokenTestCase.test
```diff
@@ -0,0 +1,45 @@
...
+  /**
+   * Test the validation of a CSRF token for write operations.
+   */
+  function testCsrfToken() {
+    $user1 = $this->drupalCreateUser(array('create article content'));
+    $this->drupalLogin($user1);
+
+    $request = array('label' => $this->randomName());
+
+    // No CSRF token.
```
e0ipso: We could iterate over the list of write methods. It will be 3 times slower but increases coverage.

tests/RestfulCsrfTokenTestCase.test
```diff
@@ -0,0 +1,45 @@
...
+
+    // Invalid CSRF token.
+    $result = $this->httpRequest('api/v1/articles', 'POST', $request, array('X-CSRF-Token: invalidToken'), NULL, FALSE);
+    $this->assertEqual($result['code'], 403, '403 error on POST with invalid CSRF token.');
+
+    // Valid CSRF token.
+    $result = $this->httpRequest('api/v1/articles', 'POST', $request);
+    $this->assertEqual($result['code'], 200, 'Entity was created on POST with CSRF token.');
+    debug($result);
+  }
```
e0ipso, line 135: We should also make sure that non-write operations work without token or even with invalid tokens.

tests/RestfulCurlBaseTestCase.test
```diff
@@ -78,8 +81,14 @@ class RestfulCurlBaseTestCase extends DrupalWebTestCase {
         break;
     }
 
+    if ($this->loggedInUser && $method != 'GET' && $use_token) {
```
e0ipso:
``` php
if ($this->loggedInUser && !in_array($method, array('GET', 'OPTIONS', 'HEAD', 'TRACE')) && $use_token) {
```

modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
amitaibu, line 12: what kind of edge cases?

amitaibu, line 12: ...and thanks for the review :)

restful.module
amitaibu, line 436: I've re-confirmed that this works:

```
curl https://example.com/api/v1/articles/1 \
  -H "X-Restful-Minor-Version: 1"
```

tests/RestfulCurlBaseTestCase.test
amitaibu: nice catch, thanks.

amitaibu:
Update: Tests are still failing, as re-logging a user in simpleTest + doing a Curl request doesn't work properly.

amitaibu:
Finally, tests are green - wasn't easy ;)
Lets deal with the constants in #100

amitaibu:
@mateu-aguilo-bosch I think it's ready to go -- any other comments?

e0ipso:
Looking good!

amitaibu:
Merged.

PR #100, opened 2014-07-05
Add constants for token value and HTTP operations.

e0ipso:
This will also introduce a `RestfulBase` class to help (in time) alleviate the `RestfulEntityBase` class of all non-entity methods/functionality.


plugins/restful/RestfulInterface.php
```diff
@@ -9,6 +9,19 @@
 interface RestfulInterface {
 
   /**
+   * HTTP methods.
+   */
+  const GET = 'get';
```
amitaibu, line 14: what's the advantage of having these constants?  it looks more complex to write `\RestfulInterface::GET` then `get` -- and the result is the same.

e0ipso, line 14: If you mistype 'GET' (instead of lower case) you will have unexpected bugs hard to track. With a constant you will get an error.

Also, all these 'get' are not independent from one another, so they should be centralized somewhere under a single_

e0ipso, line 14: A single _something_ (variable or constant)

plugins/restful/RestfulBase.php
```diff
@@ -0,0 +1,42 @@
...
+    return in_array($method, array(
+      \RestfulInterface::GET,
+      \RestfulInterface::HEAD,
+      \RestfulInterface::OPTIONS,
+      \RestfulInterface::TRACE,
+      \RestfulInterface::CONNECT,
+    ));
+  }
+
+}
```
e0ipso: Line break.

plugins/restful/RestfulInterface.php
amitaibu, line 14: > If you mistype 'GET' (instead of lower case) you will have unexpected bugs hard to track. With a constant you will get an error.

Ok, this argument convinced me :)

amitaibu:
Can you please re-roll.

e0ipso:
Conflicts resolved.

amitaibu:
Merged, thanks.

Issue #104, opened 2014-07-12
Inject file validators and scheme from the plugin


amitaibu:
https://github.com/Gizra/restful/pull/79#discussion_r14813306

PR #106, opened 2014-07-12
Inject file validators and scheme from the plugin

amitaibu:
#104

### DIFF
--- a/plugins/restful/file/files_upload/1.0/RestfulFilesUpload.class.php
+++ b/plugins/restful/file/files_upload/1.0/RestfulFilesUpload.class.php
@@ -17,6 +17,31 @@ class RestfulFilesUpload extends \RestfulEntityBase {
   );
 
   /**
+   * Overrides \RestfulEntityBase::__construct()
+   *
+   * Set the "options" key from the plugin info, specific for file upload, with
+   * the following keys:
+   * - "validators": By default no validation is done on the file extensions or
+   *   file size.
+   * - "scheme": By default the default scheme (e.g. public, private) is used.
+   */
+  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
+    parent::__construct($plugin, $auth_manager, $cache_controller);
+
+    $options = $this->getPluginInfo('options');
+
+    $default_values = array(
+      'validators' => array(
+        'file_validate_extensions' => array(),
+        'file_validate_size' => array(),
+      ),
+      'scheme' => file_default_scheme(),
+    );
+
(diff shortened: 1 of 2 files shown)


plugins/restful/file/files_upload/1.0/RestfulFilesUpload.class.php
```diff
@@ -68,4 +65,27 @@ public function createEntity($request = NULL, stdClass $account = NULL) {
...
+  /**
+   * Return the "options" key from the plugin info.
+   *
+   * @return array
+   *   Array with the assigned or default values.
+   *   "validators": By default no validation is done on the file extensions or
+   *   file size.
+   *   "scheme": By defaults the scheme scheme (e.g. public, private) is used.
+   */
+  public function getPluginInfoOptions() {
```
e0ipso: Not sure I get why a new dedicated method for the options array. This breaks the pattern we have so far.

Usually we would set the defaults for the plug-in in the process function and getting them with getPluginInfo.

amitaibu: I wasn't sure either if to do a method or not.
Question is do we want to set those default values to _all_ restful plugins? So my reasoning is that I was assuming other file-upload plugins will extend `RestfulFilesUpload`.

amitaibu: maybe we should add a process callback for setting defaults by entity_type?

e0ipso: :+1:

e0ipso: What about setting the defaults for the specific plug-in type in the constructor?

amitaibu: You are right, I'll go with the constructor.

plugins/restful/file/files_upload/1.0/RestfulFilesUpload.class.php
```diff
@@ -48,7 +70,7 @@ public function createEntity($request = NULL, stdClass $account = NULL) {
         $_FILES['files'][$key][$name] = $value;
       }
 
-      if (!$file = file_save_upload($name, $validators, file_default_scheme() . "://")) {
+      if (!$file = file_save_upload($name, $options['validators'], $options['scheme'] . "://")) {
```
snirs90, line 73: Maybe let's call it: `file_validators`, `file_schema` ?

amitaibu, line 73: since those options are specific to file upload I think it's ok to be less verbose in the property name.

amitaibu:
Merged.

PR #108, opened 2014-07-13
Fix access token, and checking access for correct account

amitaibu:
- We need to check access for the passed $account not the active user
- We need to allow POSTing using access-token, thus moving it to the http header (otherwise it will be part of the request, and then it will throw an exception)
- Allow browser to send preflight OPTIONS before POST

### DIFF
--- a/plugins/restful/user/login_cookie/1.0/RestfulUserLoginCookie.class.php
+++ b/plugins/restful/user/login_cookie/1.0/RestfulUserLoginCookie.class.php
@@ -36,7 +36,9 @@ class RestfulUserLoginCookie extends \RestfulEntityBase {
     $version = $this->getVersion();
     $handler = restful_get_restful_handler('users', $version['major'], $version['minor']);
 
-    return $handler->viewEntity($account->uid, $request, $account);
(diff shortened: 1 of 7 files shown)


plugins/restful/user/login_cookie/1.0/RestfulUserLoginCookie.class.php
```diff
@@ -36,7 +36,9 @@ public function loginAndRespondWithCookie($request = NULL, stdClass $account = N
     $version = $this->getVersion();
     $handler = restful_get_restful_handler('users', $version['major'], $version['minor']);
 
-    return $handler->viewEntity($account->uid, $request, $account);
+    $output = $handler->viewEntity($account->uid, $request, $account);
+    $output += restful_csrf_session_token();
```
amitaibu, line 40: @mateu-aguilo-bosch what do you think about this one -- along with the user info when you login, you also get the CSRF -- so we save a request.

amitaibu:

amitaibu:
btw, here's a sneak preview to the Angular app that needs this PR:

screenflow 1

amitaibu:
Merged.

PR #110, opened 2014-07-14
Upload files testing & improvements

e0ipso:


plugins/restful/RestfulFilesUpload.php
```diff
@@ -90,4 +90,16 @@ public function createEntity($request = NULL, stdClass $account = NULL) {
 
     return $return;
   }
+
+  /**
+   * Overrides RestfulEntityBase::access().
+   */
+  public function access() {
+    global $user;
```
amitaibu: this should use the injected $account, so it works with token auth as-well.

plugins/restful/RestfulBase.php
```diff
@@ -12,12 +12,20 @@
    *
    * @param string $method
    *   The method name.
+   * @param boolean $strict
+   *   TRUE if the comparisons are case sensitive.
```
amitaibu, line 16: can't we always cast it to lower case?

tests/RestfulCreateEntityTestCase.test
```diff
@@ -153,4 +164,56 @@ class RestfulCreateEntityTestCase extends DrupalWebTestCase {
     $this->assertEqual($result['image_multiple'], $ids, 'Created entity with comma separated multiple image reference.');
 
   }
+
+  /**
+   * Test access for file upload.
+   */
+  public function testFileUploadAccess() {
```
amitaibu, line 179: I'm so happy you did it :)

plugins/restful/RestfulFilesUpload.php
e0ipso: Yeah, you're right. I just copied this from the custom access function in `restful.module`.

plugins/restful/RestfulBase.php
e0ipso, line 16: Actually I made it to be uppercase by default so it's the same format the curl wrapper SimpleTest expects. Since we're using constants now capitalization shouldn't be a problem.

This is optional to make it fail in those rare occasions we want to make sure the capitalization is correct (for instance the curl example above).

plugins/restful/RestfulFilesUpload.php
amitaibu: will you have time to fix it, or prefer me or Snir to do it?

e0ipso:
@amitaibu, @snirs90 sorry it took so long for a small change.

e0ipso:
Tests should come back green now.

plugins/restful/RestfulFilesUpload.php
```diff
@@ -90,4 +90,21 @@ public function createEntity($request = NULL, stdClass $account = NULL) {
...
+   */
+  public function access() {
+    try {
+      $account = $this->getAccount();
+    }
+    catch (\RestfulUnauthorizedException $e) {
+      $account = drupal_anonymous_user();
+    }
+    if (module_exists('file_entity')) {
+      return user_access('bypass file access') || user_access('create files');
```
amitaibu: you need to pass the `$account` to `user_access()`

plugins/restful/RestfulFilesUpload.php
```diff
@@ -90,4 +90,21 @@ public function createEntity($request = NULL, stdClass $account = NULL) {
 
     return $return;
   }
+
+  /**
+   * Overrides RestfulEntityBase::access().
+   */
+  public function access() {
+    try {
```
amitaibu, line 101: Maybe worth adding a comment why we do this, as I'm not totally clear.

plugins/restful/RestfulBase.php
```diff
@@ -39,4 +50,19 @@ public static function isReadMethod($method) {
...
+   *
+   * @param string $method
+   *   The method name.
+   * @param boolean $strict
+   *   TRUE if the comparisons are case sensitive.
+   *
+   * @return boolean
+   *   TRUE if it is a known method. FALSE otherwise.
+   */
+  public static function isKnownMethod($method, $strict = TRUE) {
```
amitaibu: maybe `isValidMethod`?

amitaibu:
Also needs a re-roll.

e0ipso:
This came back green after the feedback implementation.

amitaibu:
Merged, thanks!

Issue #109, opened 2014-07-14
Allow creating sub-resources in a single request


amitaibu:
For example, if we have a node that reference a new term, currently you need 2 requests: one to create the term, and the other to create the node that references it.

Since RESTful already has an option to map a resource to an entity-reference field, we could save that extra request.

This one is a little hard, as if one entity isn't valid, it should fail the others, so we might opt for a `db_transaction()` here.

@mateu-aguilo-bosch thoughts?

PR #113, opened 2014-07-16
Allow creating referenced sub-resources in a single request

amitaibu:
#109

``` php
  $request = array(
    'label' => 'parent',
    'body' => 'Gizra1',
    'er' => array(
      'label' => 'child',
      'body' => 'Gizra2',
    ),
  );
  $result = $handler->post('', $request);
```

site-install

Validation is also working, since we are recursing into the referenced sub-resources, so nothing is saved if one of entities is invalid!

Note that the PR also moved request, account, path and method to properties of the handler, to stop passing so many arguments around from function to function.

### DIFF
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -9,7 +9,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function applies($request = NULL, $method = 'get') {
+  public function applies(array $request = array(), $method = \RestfulBase::GET) {
     $key_name = !empty($this->plugin['options']['param_name']) ? $this->plugin['options']['param_name'] : 'access_token';
     return !empty($request[$key_name]);
   }
@@ -17,7 +17,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function authenticate($request = NULL, $method = 'get') {
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
     $key_name = !empty($this->plugin['options']['param_name']) ? $this->plugin['options']['param_name'] : 'access_token';
 
     // Check if there is a token that did not expire yet.
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -9,7 +9,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function applies(array $request = array(), $method = \RestfulBase::GET) {
+  public function applies(array $request = array(), $method = \RestfulInterface::GET) {
     $key_name = !empty($this->plugin['options']['param_name']) ? $this->plugin['options']['param_name'] : 'access_token';
     return !empty($request[$key_name]);
   }
@@ -17,7 +17,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
+  public function authenticate(array $request = array(), $method = \RestfulInterface::GET) {
     $key_name = !empty($this->plugin['options']['param_name']) ? $this->plugin['options']['param_name'] : 'access_token';
 
     // Check if there is a token that did not expire yet.
--- a/plugins/authentication/RestfulAuthenticationBasic.class.php
+++ b/plugins/authentication/RestfulAuthenticationBasic.class.php
@@ -10,7 +10,7 @@ class RestfulAuthenticationBasic extends RestfulAuthenticationBase implements Re
   /**
    * {@inheritdoc}
    */
-  public function applies(array $request = array(), $method = \RestfulBase::GET) {
+  public function applies(array $request = array(), $method = \RestfulInterface::GET) {
     list($username, $password) = $this->getCredentials();
     return isset($username) && isset($password);
   }
@@ -20,7 +20,7 @@ class RestfulAuthenticationBasic extends RestfulAuthenticationBase implements Re
    *
    * @see user_login_authenticate_validate().
    */
-  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
+  public function authenticate(array $request = array(), $method = \RestfulInterface::GET) {
     list($username, $password) = $this->getCredentials();
 
     // Do not allow any login from the current user's IP if the limit has been
--- a/plugins/authentication/RestfulAuthenticationBasic.class.php
+++ b/plugins/authentication/RestfulAuthenticationBasic.class.php
@@ -10,7 +10,7 @@ class RestfulAuthenticationBasic extends RestfulAuthenticationBase implements Re
   /**
    * {@inheritdoc}
    */
-  public function applies($request = NULL, $method = 'get') {
+  public function applies(array $request = array(), $method = \RestfulBase::GET) {
     list($username, $password) = $this->getCredentials();
     return isset($username) && isset($password);
   }
@@ -20,7 +20,7 @@ class RestfulAuthenticationBasic extends RestfulAuthenticationBase implements Re
    *
    * @see user_login_authenticate_validate().
    */
-  public function authenticate($request = NULL, $method = 'get') {
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
     list($username, $password) = $this->getCredentials();
 
     // Do not allow any login from the current user's IP if the limit has been
--- a/plugins/authentication/RestfulAuthenticationCookie.class.php
+++ b/plugins/authentication/RestfulAuthenticationCookie.class.php
@@ -9,7 +9,7 @@ class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements R
   /**
    * Implements RestfulAuthenticationInterface::authenticate().
    */
-  public function authenticate($request = NULL, $method = 'get') {
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
     if (!drupal_session_started() && !$this->isCli()) {
       return;
     }
--- a/plugins/authentication/RestfulAuthenticationCookie.class.php
+++ b/plugins/authentication/RestfulAuthenticationCookie.class.php
@@ -9,7 +9,7 @@ class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements R
   /**
    * Implements RestfulAuthenticationInterface::authenticate().
    */
-  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
+  public function authenticate(array $request = array(), $method = \RestfulInterface::GET) {
     if (!drupal_session_started() && !$this->isCli()) {
       return;
     }
--- a/plugins/authentication/RestfulAuthenticationInterface.php
+++ b/plugins/authentication/RestfulAuthenticationInterface.php
@@ -18,14 +18,14 @@ interface RestfulAuthenticationInterface {
    * @return \stdClass|null
    *   The user object.
    */
-  public function authenticate($request = NULL, $method = 'get');
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET);
 
   /**
    * Determines if the request can be checked for authentication. For example,
    * when authenticating with HTTP header, return FALSE if the header values do
    * not exist.
    *
-   * @param $request
+   * @param array $request
    *   The request.
    * @param string $method
    *   The HTTP method. Defaults to "get".
@@ -33,7 +33,7 @@ interface RestfulAuthenticationInterface {
    * @return bool
    *   TRUE if the request can be checked for authentication, FALSE otherwise.
    */
-  public function applies($request = NULL, $method = 'get');
+  public function applies(array $request = array(), $method = 'get');
 
   /**
    * Get the name of the authentication plugin.
--- a/plugins/authentication/RestfulAuthenticationInterface.php
+++ b/plugins/authentication/RestfulAuthenticationInterface.php
@@ -18,7 +18,7 @@ interface RestfulAuthenticationInterface {
    * @return \stdClass|null
    *   The user object.
    */
-  public function authenticate(array $request = array(), $method = \RestfulBase::GET);
+  public function authenticate(array $request = array(), $method = \RestfulInterface::GET);
 
   /**
    * Determines if the request can be checked for authentication. For example,
--- a/plugins/authentication/RestfulAuthenticationInterface.php
+++ b/plugins/authentication/RestfulAuthenticationInterface.php
@@ -33,7 +33,7 @@ interface RestfulAuthenticationInterface {
    * @return bool
    *   TRUE if the request can be checked for authentication, FALSE otherwise.
    */
-  public function applies(array $request = array(), $method = 'get');
+  public function applies(array $request = array(), $method = \RestfulInterface::GET);
 
   /**
    * Get the name of the authentication plugin.
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -66,7 +66,7 @@ class RestfulAuthenticationManager extends \ArrayObject {
    * @return \stdClass
    *   The user object.
    */
-  public function getAccount(array $request = array(), $method = \RestfulBase::GET) {
+  public function getAccount(array $request = array(), $method = \RestfulInterface::GET) {
     global $user;
     // Return the previously resolved user, if any.
     if (!empty($this->account)) {
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -66,7 +66,7 @@ class RestfulAuthenticationManager extends \ArrayObject {
    * @return \stdClass
    *   The user object.
    */
-  public function getAccount($request = NULL, $method = 'get') {
+  public function getAccount($request = NULL, $method = \RestfulBase::GET) {
     global $user;
     // Return the previously resolved user, if any.
     if (!empty($this->account)) {
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -58,7 +58,7 @@ class RestfulAuthenticationManager extends \ArrayObject {
   /**
    * Get the user account for the request.
    *
-   * @param $request
+   * @param array $request
    *   The request.
    * @param string $method
    *   The HTTP method.
@@ -66,7 +66,7 @@ class RestfulAuthenticationManager extends \ArrayObject {
    * @return \stdClass
    *   The user object.
    */
-  public function getAccount($request = NULL, $method = \RestfulBase::GET) {
+  public function getAccount(array $request = array(), $method = \RestfulBase::GET) {
     global $user;
     // Return the previously resolved user, if any.
     if (!empty($this->account)) {
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1280,7 +1280,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
 
     if (empty($params['@fields'])) {
       // There was a validation error, but on non-public fields, so we need to
-      // throw an exception, but can say on which fields it coccured.
+      // throw an exception, but can't say on which fields it occurred.
       throw new \RestfulBadRequestException('Invalid value(s) sent with the request.');
     }
 
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -471,7 +471,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
       $this->getRateLimitManager()->checkRateLimit($request);
     }
 
-    return $this->{$method_name}();
+    return $this->{$method_name}($path);
   }
 
   /**
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1093,9 +1093,6 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
     if ($field_info['cardinality'] == 1 && !is_array($value)) {
       return $value;
     }
-    elseif ($field_info['cardinality'] != 1 && !is_array($value[0])) {
-      return $value;
-    }
 
     // In case we have multiple bundles, we opt for the first one.
     $resource_name = reset($public_fields[$public_field_name]['resource']);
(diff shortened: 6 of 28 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -991,21 +1054,79 @@ public function propertyValuesPreprocess($property_name, $value) {
...
+    }
+    elseif ($field_info['cardinality'] != 1 && !is_array($value[0])) {
+      return $value;
+    }
+
+    // In case we have multiple bundles, we opt for the first one.
+    $resource_name = reset($public_fields[$public_field_name]['resource']);
+
+    $version = $this->getVersion();
+    $handler = restful_get_restful_handler($resource_name, $version['major'], $version['minor']);
```
snirs90, line 1102: Maybe to let the user to pass the resource version in case the user wants other version than the "parent" resource?
And in case the user didn't defined any version it will take by default the "parent" resource version.

amitaibu, line 1102: Opened #114

e0ipso:
> Note that the PR also moved request, account, path and method to properties of the handler, to stop passing so many arguments around from function to function.

:clap:

modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
```diff
@@ -9,7 +9,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   /**
    * {@inheritdoc}
    */
-  public function applies($request = NULL, $method = 'get') {
+  public function applies(array $request = array(), $method = \RestfulBase::GET) {
```
e0ipso: I believe we want:

``` php
public function applies(array $request = array(), $method = \RestfulInterface::GET)
```

modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
```diff
@@ -17,7 +17,7 @@ public function applies($request = NULL, $method = 'get') {
   /**
    * {@inheritdoc}
    */
-  public function authenticate($request = NULL, $method = 'get') {
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
```
e0ipso: I believe we want:

``` php
public function authenticate(array $request = array(), $method = \RestfulInterface::GET)
```

plugins/authentication/RestfulAuthenticationBasic.class.php
```diff
@@ -10,7 +10,7 @@ class RestfulAuthenticationBasic extends RestfulAuthenticationBase implements Re
   /**
    * {@inheritdoc}
    */
-  public function applies($request = NULL, $method = 'get') {
+  public function applies(array $request = array(), $method = \RestfulBase::GET) {
```
e0ipso: Same here (`\RestfulInterface::GET`).

plugins/authentication/RestfulAuthenticationBasic.class.php
```diff
@@ -20,7 +20,7 @@ public function applies($request = NULL, $method = 'get') {
    *
    * @see user_login_authenticate_validate().
    */
-  public function authenticate($request = NULL, $method = 'get') {
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
```
e0ipso: Same here (`\RestfulInterface::GET`).

plugins/authentication/RestfulAuthenticationCookie.class.php
```diff
@@ -9,7 +9,7 @@ class RestfulAuthenticationCookie extends RestfulAuthenticationBase implements R
   /**
    * Implements RestfulAuthenticationInterface::authenticate().
    */
-  public function authenticate($request = NULL, $method = 'get') {
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET) {
```
e0ipso: Same here (`\RestfulInterface::GET`).

plugins/authentication/RestfulAuthenticationInterface.php
```diff
@@ -18,14 +18,14 @@
    * @return \stdClass|null
    *   The user object.
    */
-  public function authenticate($request = NULL, $method = 'get');
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET);
```
e0ipso: Same here (`\RestfulInterface::GET`).

plugins/authentication/RestfulAuthenticationInterface.php
```diff
@@ -33,7 +33,7 @@ public function authenticate($request = NULL, $method = 'get');
    * @return bool
    *   TRUE if the request can be checked for authentication, FALSE otherwise.
    */
-  public function applies($request = NULL, $method = 'get');
+  public function applies(array $request = array(), $method = 'get');
```
e0ipso: This one is not using the constant.

plugins/authentication/RestfulAuthenticationInterface.php
```diff
@@ -18,14 +18,14 @@
...
-  public function authenticate($request = NULL, $method = 'get');
+  public function authenticate(array $request = array(), $method = \RestfulBase::GET);
 
   /**
    * Determines if the request can be checked for authentication. For example,
    * when authenticating with HTTP header, return FALSE if the header values do
    * not exist.
    *
-   * @param $request
+   * @param array $request
```
e0ipso, line 28: I array type hinting compatible with all supported PHP versions?

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -66,7 +66,7 @@ public function addAuthenticationProvider(RestfulAuthenticationInterface $provid
    * @return \stdClass
    *   The user object.
    */
-  public function getAccount($request = NULL, $method = 'get') {
+  public function getAccount(array $request = array(), $method = \RestfulBase::GET) {
```
e0ipso: Same here (`\RestfulInterface::GET`).

plugins/restful/RestfulEntityBase.php
```diff
@@ -348,54 +436,49 @@ public function patch($path = '', $request = NULL) {
...
     return $this->process($path, $request, \RestfulInterface::DELETE);
   }
 
   /**
    * {@inheritdoc}
    */
-  public function process($path = '', $request = NULL, $method = \RestfulInterface::GET) {
-    $account = $this->getAccount($request, $method);
+  public function process($path = '', array $request = array(), $method = \RestfulInterface::GET, $check_rate_limit = TRUE) {
+    $this->setMethod($method);
```
e0ipso, line 453: I am worried that we are turning the `process` method into a mash of _initialization_ and _request processing_ functionality.

What if we move this up to the constructor? If needed we can alter the request and such with the public getters and setters before calling `$handler->get()`.

plugins/restful/RestfulEntityBase.php
```diff
@@ -289,13 +377,13 @@ public function getVersion() {
...
    *   (optional) The path.
-   * @param null $request
+   * @param array $request
    *   (optional) The request.
    *
    * @return mixed
    *   The return value can depend on the controller for the get method.
    */
-  public function get($path = '', $request = NULL) {
+  public function get($path = '', array $request = array()) {
```
e0ipso, line 386: If we move the `$path` and `$request` to the constructor then this functions won't need any arguments. See https://github.com/Gizra/restful/pull/113/files#discussion_r15106369

We can keep the docblock docs turning them into `@internal` if needed.

plugins/restful/RestfulEntityBase.php
```diff
@@ -348,54 +436,49 @@ public function patch($path = '', $request = NULL) {
...
     }
 
-    if (!$path) {
-      // If $path is empty we don't need to pass it along.
-      return $this->{$method_name}($request, $account);
-    }
-    else {
-      return $this->{$method_name}($path, $request, $account);
-    }
+    return $this->{$method_name}($path);
```
e0ipso, line 466: Why are we keeping the `$path` here? Can't we just use `$this->getPath()`?

tests/RestfulSubResourcesCreateEntityTestCase.test
```diff
@@ -0,0 +1,141 @@
...
+ * @file
+ * Contains RestfulSubResourcesCreateEntityTestCase.
+ */
+
+class RestfulSubResourcesCreateEntityTestCase extends DrupalWebTestCase {
+
+  /**
+   * Hold the RESTful handler.
+   *
+   * @var RestfulInterface | NULL
```
e0ipso: I think that `NULL` is optional when using the `|`

tests/RestfulSubResourcesCreateEntityTestCase.test
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  /**
+   * Test creating an entity with sub-resources.
+   */
+  function testCreateEntity() {
+    $handler = $this->handler;
+    $request = array(
+      'label' => 'parent',
+      'body' => 'Gizra',
```
e0ipso: Maybe make it less branded? I'm sure the attention will make it back to _Gizra_ anyways.

Don't take me wrong, **I could not care less**, but there's people (potential contributors) that may be stirred away because of these things.

tests/RestfulSubResourcesCreateEntityTestCase.test
```diff
@@ -0,0 +1,141 @@
...
+          'label' => 'child2',
+          'body' => 'Gizra2',
+        ),
+        array(
+          'label' => 'child3',
+          'body' => 'Gizra3',
+        ),
+      ),
+    );
+    $result = $handler->post('', $request);
```
e0ipso: Random thought: What if we created a wrapper method on our `DrupalTestCase` subclass that would test `post` via Drupal and via Curl.

Maybe open a feature request?

plugins/restful/RestfulEntityBase.php
```diff
@@ -991,21 +1046,79 @@ public function propertyValuesPreprocess($property_name, $value) {
...
+    if (empty($public_fields[$public_field_name]['resource'])) {
+      // Field is not defined as "resource", which means it only accepts an
+      // integer as a valid value.
+      return $value;
+    }
+
+    if ($field_info['cardinality'] == 1 && !is_array($value)) {
+      return $value;
+    }
+    elseif ($field_info['cardinality'] != 1 && !is_array($value[0])) {
```
e0ipso: Can we have a comment on the logic here? Explaining what means that `$value[0]` is not an array.

plugins/restful/RestfulEntityBase.php
```diff
@@ -991,21 +1046,79 @@ public function propertyValuesPreprocess($property_name, $value) {
...
+    if ($field_info['cardinality'] == 1) {
+      $result = $handler->process($this->getPath(), $value, $this->getMethod(), FALSE);
+      // Single value.
+      return $result['id'];
+    }
+
+    // Multiple values.
+    $return = array();
+    foreach ($value as $value_item) {
+      $result = $handler->process($this->getPath(), $value_item, $this->getMethod(), FALSE);
```
e0ipso: Not sure I get this.

Scenario: I'm issuing a PATCH request to update the title of the article and to create a new taxonomy term.

With the current code I would be:
1. Limited to update operations (PATCH) for the taxonomy term.
2. Updating the taxonomy term with the ID that corresponds to the article that we want to update. Therefore overwriting an unrelated taxonomy term with the data that we wanted for the new term.

I think that if we go down this path (and we should), we need to have a request that looks like this:

``` php`
$request = array(
  'label' => 'parent',
  'body' => 'Gizra1',
  'er' => array(
    'endpoint' => 'api/v1/tag', // This would need to match the resource in public fields or fail.
    'method' => 'POST',
    'headers' => array(
      'X-Restful-Minor-Version' => 2,
    ),
    'request' => array(
      array(
        'label' => 'child',
        'body' => 'Gizra2',
      ),
      array(
        'label' => 'child',
        'body' => 'Gizra3',
      ),
    ),
  ),
);
$result = $handler->post('', $request);
```

Then we could be able to combine several requests as one.

e0ipso: Of course we could provide sensible defaults to make the `$request` less bloated.

plugins/restful/RestfulEntityBase.php
amitaibu, line 466: this is convinent where for example when we create a content using `$handler->post('', $request)` -- but the response is actually `$handler->get($entity_id);` - so we don't want to change $this->path

plugins/restful/RestfulEntityBase.php
amitaibu: I see the logic in what you suggest, however I would omit the  `endpoint` and `headers` as they I think should be declared in the getPublicFields. maybe something like this:

``` php
$field['foo'] = array(
  'property' => 'field_foo_reference',
  'resource' => array(
    // shothand form
    'article' => 'articles'
   'page' => array(
      'name' => 'pages'
      'major_version' => 1,
      'minor_version' => 1,
   ),
  ),
); 
```

tests/RestfulSubResourcesCreateEntityTestCase.test
```diff
@@ -0,0 +1,141 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulSubResourcesCreateEntityTestCase.
+ */
+
+class RestfulSubResourcesCreateEntityTestCase extends DrupalWebTestCase {
```
amitaibu, line 8: test is still WIP, as we need to check PUT/ PATCH as-well.

tests/RestfulSubResourcesCreateEntityTestCase.test
amitaibu: You are very right, sorry about that, Already fixed entity validator. Replaced it with "Drupal" ;)

tests/RestfulSubResourcesCreateEntityTestCase.test
amitaibu: Good idea, but It would be a little complicated, as we'll need to mock the calling user? But probably worth a shot.

plugins/authentication/RestfulAuthenticationInterface.php
amitaibu, line 28: I'm not aware of any issues with PHP5.3+

plugins/restful/RestfulEntityBase.php
amitaibu, line 386: If I understand correctly what you mean, it will change this:

``` php
$handler = restful_get_restful_handler('test_articles', 1, 2);
$handler->post('', $request);
$handler->get($entity_id);
```

Into this:

``` php
$handler = restful_get_restful_handler('test_articles', 1, 2, 'post', $request);
// POST
$result = $handler->post();

// GET, with the same handler.
$handler->setRequest(NULL);
$handler->setPath($entity_id);
$result = $handler->get();
```

The latter seems a bit less readable to me, but I'm open to discussion if you feel strongly about it.

plugins/restful/RestfulEntityBase.php
amitaibu, line 453: Lets discuss this on this comment -- https://github.com/Gizra/restful/pull/113/files#discussion_r15127459

plugins/restful/RestfulEntityBase.php
e0ipso: We are still missing the IDs for the sub-resouces and the HTTP method for them.

plugins/authentication/RestfulAuthenticationInterface.php
e0ipso, line 28: I'm perfectly happy to ditch PHP5.2 but we should at least declare that in the info file (if that is the case).

plugins/restful/RestfulEntityBase.php
amitaibu: indeed, the ID should be given in the request, otherwise it would be a POST. I'll work on a logic, and ping once it's ready.

plugins/restful/RestfulEntityBase.php
e0ipso, line 386: > The latter seems a bit less readable to me

Hrm, it seems a bit more readable to me. I guess it's a matter of preference. However, do you think that there are many use cases where we are getting a handler to do `post + get` in the user implementations?

I'm OK either way. Or even defaulting `$path` and `$request` to the class state only if they're not provided in the function argument.

amitaibu, line 386: I guess mostly in tests, so we can change to your suggestion. But do you think we need to require this in the constructor? - I'd prefer avoiding something like this - `$handler = restful_get_restful_handler('test_articles', 1, 2, 'post', $request);`

e0ipso:
@amitaibu I think this PR is crucial. Doing this right will result in a very good user experience for the client implementations.

amitaibu:
> I think this PR is crucial. Doing this right will result in a very good user experience for the client implementations.

I agree :)

plugins/restful/RestfulEntityBase.php
e0ipso, line 386: Hmm, since we're already reading from PHP globals we could read them in the constructor, then? Not totally sold on this idea, just brainstorming here.

amitaibu, line 386: not sure exactly what you mean. here's the code I think clients should do:

``` php
$handler = restful_get_restful_handler('test_articles', 1, 2);
$handler->setRequest(array('label' => 'foo');
$handler->setPath($entity_id);
$result = $handler->get();
```

do you agree?

e0ipso, line 386: Exactly, that's when using the API via Drupal. When using it via CURL we'll have the request object in `php://input` so we can read it in the constructor if necessary.

amitaibu, line 386: > When using it via CURL we'll have the request object in php://input so we can read it in the constructor if necessary.

I don't think it will be needed, since the handler is executed from the page callback (`restful_menu_process_callback()`) - so the handler works the same for Drupal API/ CURL.

amitaibu, line 386: actually, we could keep the code used by the client as-is, and just change the wrapper functions, for example:

``` php
public function get($path = '', array $request = array()) {
    $this->setMethod(\RestfulInterface::GET);
    $this->setPath($path);
    $this->setRequest($request);

    return $this->process();
  }
```

amitaibu, line 386: actually2, I think we should keep process() getting all those arguments, so the same handler could be called with different params, when dealing with sub-resources.

plugins/restful/RestfulEntityBase.php
amitaibu: yeah, the logic is wrong here ;) I'll fix it -- as one might pass multiple values (e.g. existing ID, entity for POST, entity for PATCH)

plugins/restful/RestfulEntityBase.php
```diff
@@ -991,21 +1049,109 @@ public function propertyValuesPreprocess($property_name, $value) {
...
+    }
+
+    // In case we have multiple bundles, we opt for the first one.
+    $resource_name = reset($public_fields[$public_field_name]['resource']);
+
+    $version = $this->getVersion();
+    $handler = restful_get_restful_handler($resource_name, $version['major'], $version['minor']);
+
+    // Return the entity ID that was created.
+    if ($field_info['cardinality'] == 1) {
```
amitaibu, line 1105: @mateu-aguilo-bosch create sub-resource logic is now working, and tested.

amitaibu:
Update: I'll need to revert a few wrong commits, and re-fix the tests.

amitaibu:
@mateu-aguilo-bosch tests are green. ready for review.

plugins/authentication/RestfulAuthenticationInterface.php
snirs90, line 28: Array hinting is compatible  from PHP5.1+
http://php.net/manual/en/language.oop5.typehinting.php
`arrays (since PHP 5.1)`

e0ipso, line 28: Thanks @snirs90 I did not know that.

e0ipso:
Starting review.

plugins/restful/RestfulEntityBase.php
e0ipso, line 386: Maybe worth dropping a comment in `process()` code?

``` php
// The class level variables are not used for convenience when dealing with sub-resources.
```

plugins/restful/RestfulEntityBase.php
```diff
@@ -991,21 +1049,109 @@ public function propertyValuesPreprocess($property_name, $value) {
...
+   *
+   * @param int | array $value
+   *   The entity ID, or array to POST, PATCH, or PUT entity from.
+   * @param \RestfulInterface $handler
+   *   The RESTful handler.
+   *
+   * @return int
+   *   The saved entity ID.
+   */
+  protected function createOrUpdateSubResourceItem($value, $handler) {
```
e0ipso, line 1130: What about `get`? I think it would be nice to cover the following scenario:

I want to create an article with title _Title_ body _Lorem ipsum_, tagged by _alpha, beta, gamma_ and associated to the entity with label _Series: A_

That's why I was suggesting the use of [pseudo-API calls](https://github.com/Gizra/restful/pull/113#discussion_r15111791) within the sub-resources. That way I think the code is way more maintainable since we will just have a single pattern to think of. We can have wicked functionality for free that is not supported anywhere else like authenticating different users in different sub-resources if necessary, having unlimited levels of sub-resources (create an article that references another article which has this image that in turn is tagged with a term…).

If we do custom logic for sub-resources at level-1 we will probably need to maintain those separately. I guess that the major concern is about DX in the client side, but maybe we can address that somehow later.

Other than that, this looks good.

plugins/restful/RestfulEntityBase.php
```diff
@@ -991,21 +1049,109 @@ public function propertyValuesPreprocess($property_name, $value) {
...
+
+    // Figure the method that should be used.
+    if (empty($value['id'])) {
+      $method_name = \RestfulInterface::POST;
+      $path = '';
+    }
+    else {
+      // Use PATCH by default, unless client has explicitly set the method in
+      // the sub-resource.
+      $method_name = !empty($value['__application']['method']) ? strtoupper($value['__application']['method']) : \RestfulInterface::PATCH;
```
e0ipso: We will probably need some docs about the use of `$value['__application'][…]`. I assume that the idea is to make it work like `$request['__application']`.

plugins/restful/RestfulEntityBase.php
amitaibu, line 1130: > I want to create an article with title Title body Lorem ipsum, tagged by alpha, beta, gamma and associated to the entity with label Series: A

Not clear about what you say -- can you show how your `$request` would look like?

> anywhere else like authenticating different users in different sub-resources if necessary

IMO, Sounds like something that has no real use case ;)

> If we do custom logic for sub-resources at level-1 we will probably need to maintain those separately. 

Since this is a recursive function, you could nest those references. Maybe worth adding a test to assert it.

amitaibu:
Merged.

plugins/restful/RestfulEntityBase.php
e0ipso, line 1130: Article:
- title: _Title_
- body: _Lorem ispum_
- tags: _alpha, beta, gamma_
- series_reference:
  - label: _Reference A_ (find it or create it)

> IMO, Sounds like something that has no real use case ;)

True, that's why I added the second example. This was just to point out that we will be creating a flexibility that we haven't even thought yet (think about contrib adding plugins and getting the ability to make them different for every subresource).

The following one sounds very realistic:

Article:
- title: _Title_
- body: _Lorem ispum_
- related: (Article: you may want create, update, search, update or create, search or create)
  - title: _Related title_
  - image: (File entity: you may want create, update, search, update or create, search or create)
    - label: _Filename_
    - tags: (term: you may want create, update, search, update or create, search or create)
      - name: 'term1'
      - description: 'Lorem ipsum'
      - …

You can do this if you're able to treat each sub-resource as an original resource. Error handling becomes really simple, you just have to bubble up the error from the sub-resources.

Besides, we know that if you want to add an entity reference the user can do it with no problem with 2 requests:
- 1st: Create related node doing: POST, `api/v2/related`, with headers `X-Restful-Minor-Version: 3`, …
  - 2nd: Create the article doing: PATCH, `api/v1/articles`, with headers `X-Restful-Minor-Version: 1`, …, with the node id from the step above.

My opinion is that if we have to create sub-resource operations with have to do it with the more intuitive way for the user: let them think that is just like doing multiple API calls, but save them the HTTP round trip by making them from within the backend server.

amitaibu, line 1130: apart of the "search", the scenario you describe + the entity validation is indeed already happening.

PR #122, opened 2014-07-28
Add a generic entity access tag to \RestfulEntityBase::getQueryForList()

amitaibu:
Make sure we filter out un-accessible items, for the correct account.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -617,6 +617,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
     $query->addTag($entity_type . '_access');
     $query->addTag('restful');
     $query->addMetaData('restful_handler', $this);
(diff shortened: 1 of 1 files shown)


amitaibu:
@mateu-aguilo-bosch ^^

e0ipso:
This looks good. Merge it! :shipit: 

However, we're starting to have several access controls:
- Hook menu access (that calls the `access` method in the handler).
- Entity access, for the returned entities.
- Property access for the requested properties on the entities.
- Query tags for access purposes.
- Rate limits.
- Authentication rules (if the resource requires authentication but the request cannot resolve an account).
- Other?

Is this worth compiling and adding a section in the README?

amitaibu:
Merged.

> Is this worth compiling and adding a section in the README?

I think apart of the rate limit the rest are "expected" access. Maybe we can add a comment that it's recommend to test endpoint with anon + a regular non-admin user.

PR #125, opened 2014-08-07
added few more plugin configurations entity_id, entity_count, entity_lis...

marcogiacomassi:
In my project I needed a way to easily set entity ID and list name to a custom name.
For example:

``` javascript
{
    "products": [
        {
            "pid": 310,
            "tnid": "309",
            "title": "product name",
            "status": "1",
            "description": "product descr"
                }
        ]
}
```

I think this could be done from plugin configuration without the needs to extend RestfulEntityBaseNode class.

These are the new parameters added to plugin config

``` php
'entity_id' => 'pid',
'entity_list_name' => 'products',
'entity_count' => TRUE,
```

This also adds a count of the returned items.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -524,12 +524,11 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
       // Return autocomplete list.
       return $this->getListForAutocomplete();
     }
-
+    
     $entity_type = $this->entityType;
     $result = $this
       ->getQueryForList()
       ->execute();
-
 
     if (empty($result[$entity_type])) {
       return array();
@@ -551,9 +550,11 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
       $return[$this->entityListName][] = $this->viewEntity($id);
     }
     
-    // add count if requested in plugin config
+    // add count if requested in plugin config.
     if ($this->entityCount) {
-        $return['count'] = count($ids);
+        $count_query = $this->getQueryForList()->count();
+        unset($count_query->range);
+        $return['count'] = $count_query->execute();
     }
     return $return;
   }
@@ -629,7 +630,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
     $query->addMetaData('account', $this->getAccount());
 
     $query->range($offset, $range);
-
+    
     return $query;
   }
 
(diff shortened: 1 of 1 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -343,6 +343,11 @@ public function __construct($plugin, \RestfulAuthenticationManager $auth_manager
     $this->bundle = $plugin['bundle'];
     $this->authenticationManager = $auth_manager ? $auth_manager : new \RestfulAuthenticationManager();
     $this->cacheController = $cache_controller ? $cache_controller : $this->newCacheObject();
+    
+    $this->entityId = !empty($plugin['entity_id']) ? $plugin['entity_id'] : 'id';
```
e0ipso, line 347: I think that this is confusing. Maybe something like `entityIdKey`?

plugins/restful/RestfulEntityBase.php
```diff
@@ -343,6 +343,11 @@ public function __construct($plugin, \RestfulAuthenticationManager $auth_manager
...
+    $this->entityCount = !empty($plugin['entity_count']) ? $plugin['entity_count'] : FALSE;
```
e0ipso, line 348: How does this account pagination? Is this the total number? Is this the number of items per page?

plugins/restful/RestfulEntityBase.php
```diff
@@ -538,14 +543,18 @@ public function getList() {
...
+    $return = array($this->entityListName => array());
 
     $this->getListAddHateoas($return, $ids);
 
     foreach ($ids as $id) {
-      $return['list'][] = $this->viewEntity($id);
+      $return[$this->entityListName][] = $this->viewEntity($id);
+    }
+    
+    // add count if requested in plugin config
```
e0ipso: `// add count if requested in plugin config`
to
`// Add count if requested in plugin config.`

plugins/restful/RestfulEntityBase.php
e0ipso, line 347: Actually I'd go with whatever the name is in core/entity API module.

plugins/restful/RestfulEntityBase.php
marcogiacomassi, line 348: You are right, this should be the the total number

plugins/restful/RestfulEntityBase.php
marcogiacomassi, line 347: I think entity_id_key and entityIdKey are good! In core/entity API module are named entity keys

amitaibu:
Sorry, in vacation. I'll review once I'm back :)

amitaibu:
Thank you for the contribution, but I'm not a fan of this PR. Those changes can be pretty easily be done by overriding some methods. 
I'd like to keep the plugin configuration as much as possible out of the final data structure.

If you feel I'm missing something, please re-open.

PR #126, opened 2014-08-26
Query multiple ids

e0ipso:
For performance reasons we should allow clients to requests multiple entities at once.

Also our output structure should allow adding metadata to the response without polluting the response scope itself, hence wrapping it into a "data" object even if there is only one resource.


plugins/restful/RestfulEntityBase.php
```diff
@@ -82,8 +82,8 @@
       // POST
       \RestfulInterface::POST => 'createEntity',
     ),
-    '\d+' => array(
-      \RestfulInterface::GET => 'viewEntity',
+    '^(\d+,)*\d+$' => array(
```
amitaibu, line 85: maybe worth adding a comment with examples (e.g. `api/v1/articles/1 or api/v1/articles/1,2`)

plugins/restful/RestfulEntityBase.php
```diff
@@ -765,6 +794,7 @@ public function getListAddHateoas(&$return, &$ids){
       // "next" page.
       array_pop($ids);
     }
+    $return['count'] = count($ids);
```
amitaibu: Do we really need this count -- isn't it just a count of $result['data'], which can be achieved client side?

amitaibu:
I really like this PR. Lets get the minor comments, and get it in. I hope people already using RESTful won't be too angry at us for changing the structure ;)

e0ipso:
@amitaibu people already using RESTful are going to **love** what I'm working on right in #127. It'll give overriding capabilities on the output format.

plugins/restful/RestfulEntityBase.php
e0ipso: Nah, we don't. We should be putting the real count here (not the paginated count). But we can have a separate issue for that.

e0ipso: I will remove this.

e0ipso:
@amitaibu feedback implemented. Thanks for your suggestions.

amitaibu:
Merged, thanks.

PR #129, opened 2014-08-29
Output format plugin

e0ipso:
This is early work. I just wanted you to have a peek to the direction this is turning into.


amitaibu:
The approach makes sense. I saw some minor stuff, but I'll wait to review once it's ready.

One thing though, is that I think the FormatterBase class could be our Json+HAL, instead of an unused one.

e0ipso:
Makes sense

e0ipso:
@amitaibu this is ready to be reviewed now. I realized in the way that dealing with the special case of one item not being an array is a pain of if/else both in the back-end and in the front-end. So everything is an array now inside of `'data'`.

e0ipso:
Also a nice side effect of this PR is that we now get the `count` of assets (not limited by pagination).

modules/restful_example/plugins/formatter/RestfulFormatterHalXml.class.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulFormatterHalJson.
+ */
+
+class RestfulFormatterHalXml extends \RestfulFormatterHalJson implements \RestfulFormatterInterface {
```
amitaibu, line 8: get idea, kudos!

modules/restful_example/plugins/restful/node/articles/1.6/articles__1_6.inc
```diff
@@ -0,0 +1,15 @@
...
+  'resource' => 'articles',
+  'name' => 'articles__1_6',
+  'entity_type' => 'node',
+  'bundle' => 'article',
+  'description' => t('Export the article content type in XML format.'),
+  'class' => 'RestfulExampleArticlesResource__1_6',
+  'authentication_types' => TRUE,
+  'authentication_optional' => TRUE,
+  'minor_version' => 6,
+  'formatter' => 'hal_xml',
```
amitaibu, line 15: Maybe add a comment above, so it's clear on a quick look what we do here.

plugins/formatter/RestfulFormatterBase.php
```diff
@@ -0,0 +1,51 @@
...
+  public function format(array $data) {
+    return $this->render($this->massage($data));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getContentTypeHeader() {
+    // Default to the most general content type.
+    return 'application/hal+json; charset=utf-8';
```
amitaibu, line 48: shouldn't this be return `$this->contentType`?

plugins/formatter/RestfulFormatterHalJson.php
```diff
@@ -0,0 +1,85 @@
...
+   * Content Type
+   *
+   * @var string
+   */
+  protected $contentType = 'application/hal+json; charset=utf-8';
+
+  /**
+   * {@inheritdoc}
+   */
+  public function massage(array $data) {
```
amitaibu: maybe `massage` => `prepare`?

plugins/formatter/RestfulFormatterBase.php
e0ipso, line 48: `$this->contentType` only exists in `RestfulFormatterHalJson`. This is just a dummy implementation, more complex stuff should be delegated to children classes.

plugins/formatter/RestfulFormatterHalJson.php
e0ipso: :+1:

plugins/formatter/RestfulFormatterHalJson.php
```diff
@@ -0,0 +1,85 @@
...
+    // If we're returning an error then set the content type to
+    // 'application/problem+json; charset=utf-8'.
+    if (!empty($data['status']) && floor($data['status'] / 100) != 2) {
```
amitaibu, line 22: I think we need to generalize this, so we can: 
- `protected $contentTypeOnError = 'application/problem+json; charset=utf-8'`
- Move `if (!empty($data['status'])` logic to a central place
- For simplicity maybe we can just `floor($data['status']) != 200`?

plugins/formatter/RestfulFormatterHalJson.php
```diff
@@ -0,0 +1,85 @@
...
+      $this->contentType = 'application/problem+json; charset=utf-8';
+      return $data;
+    }
+    // Here we get the data after calling the backend storage for the resources.
+    $output = array('data' => $data);
+
+    if (!empty($this->handler)) {
```
amitaibu, line 29: what's the case of not having an handler?

plugins/formatter/RestfulFormatterHalJson.php
e0ipso, line 22: > For simplicity maybe we can just floor($data['status']) != 200?

Actually `floor(204) == 204 != 200`

> Move if (!empty($data['status']) logic to a central place

Any suggestion?

plugins/formatter/RestfulFormatterHalJson.php
e0ipso, line 29: Using the `restful_delivery` with anything that is not a resource. Right now we're using it for CSRF token requests.

plugins/formatter/RestfulFormatterHalJson.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    // We know that there are more pages if the total count is bigger than the
+    // number of items of the current request plus the number of items in
+    // previous pages.
+    $items_per_page = $this->handler->getRange();
+    $previous_items = ($page - 1) * $items_per_page;
+    if ($data['count'] > count($data['data']) + $previous_items) {
+      $request['page'] = $page + 1;
+      $data['_links']['next'] = $this->handler->getUrl();
```
amitaibu: This seems wrong (not because of you -- because of my previous change). getUrl() should get the $request

plugins/restful/RestfulBase.php
```diff
@@ -65,4 +86,34 @@ public static function isValidMethod($method, $strict = TRUE) {
...
+   */
+  public function getPluginInfo($key = NULL) {
+    if (isset($key)) {
+      return empty($this->plugin[$key]) ? NULL : $this->plugin[$key];
+    }
+    return $this->plugin;
+  }
+
+  /**
+   * Call the output format on the passed in array.
```
amitaibu: => `Call the output format on the given data.`

plugins/restful/RestfulBase.php
```diff
@@ -65,4 +86,34 @@ public static function isValidMethod($method, $strict = TRUE) {
...
+   *
+   * @param array $data
+   *   The array of data to format.
+   *
+   * @return string
+   *   The output string to return.
```
amitaibu: => `The formatted output.`

plugins/restful/RestfulEntityBase.php
```diff
@@ -654,6 +632,63 @@ public function getQueryForList() {
...
+   *
+   * @return int
+   *   The total number of results without including pagination.
+   */
+  public function getTotalCount() {
+    $count_results = $this
+      ->getQueryCount()
+      ->execute();
+    $entity_type = $this->getEntityType();
+    if (!empty($count_results[$entity_type])) {
```
amitaibu: count() will return 0 or the number, so this can be 

``` php
return $this
  ->getQueryCount()
  ->execute();
```

In that case maybe we can deprecate getQueryCount() and move logic here.

plugins/restful/RestfulEntityBase.php
```diff
@@ -654,6 +632,63 @@ public function getQueryForList() {
...
+      return count($count_results[$entity_type]);
+    }
+    return 0;
+  }
+
+  /**
+   * Adds tags and metadata to the EntityFieldQuery.
```
amitaibu: Adds tags => Add query tags

plugins/restful/RestfulEntityBase.php
```diff
@@ -1834,4 +1853,18 @@ public function cacheInvalidate($cid) {
     $this->getCacheController()->clear($cid, TRUE);
   }
 
+  /**
+   * Helper method to know if the current request is for a list of entities.
+   *
+   * @return boolean
+   *   TRUE if the request is for a list. FALSE otherwise.
+   */
+  public function isListRequest() {
```
amitaibu, line 1827: Is this needed?

restful.module
```diff
@@ -587,7 +638,26 @@ function restful_json_output($var = NULL) {
     drupal_add_http_header('Content-Type', 'application/problem+json; charset=utf-8');
   }
 
-  echo drupal_json_encode($var);
+  // Get the formatter for the current resource.
+  if ($restful_handler = restful_get_restful_handler_for_path()) {
```
amitaibu, line 645: I'm not clear on `restful_get_restful_handler_for_path`, and why do we need an `else` for it?

restful.module
```diff
@@ -587,7 +638,26 @@ function restful_json_output($var = NULL) {
...
+    $formatter_name = $restful_handler->getPluginInfo('formatter');
+    // Allow the handler to change the HTTP headers.
+    foreach ($restful_handler->getHttpHeaders() as $key => $value) {
```
amitaibu, line 648: Now that we have a formatter, maybe we can move logic there?

restful.module
```diff
@@ -587,7 +638,26 @@ function restful_json_output($var = NULL) {
...
+  else {
+    $formatter_name = variable_get('restful_default_output_formatter', 'hal_json');
+  }
+  $formatter_handler = restful_get_formatter_handler($formatter_name, $restful_handler);
+  $output = $formatter_handler->format($var);
+
+  // The content type header is modified after the massaging if there is an
+  // error code. Therefore we need to set the content type header after
+  // formatting the output.
+  drupal_add_http_header('Content-Type', $formatter_handler->getContentTypeHeader());
```
amitaibu, line 661: Now that we have a formatter, maybe we can move logic there?

tests/RestfulSubResourcesCreateEntityTestCase.test
```diff
@@ -190,7 +191,10 @@ class RestfulSubResourcesCreateEntityTestCase extends DrupalWebTestCase {
 
 
     // Validation fail on the multiple resource.
-    if (is_array($request['entity_reference_multiple'][0]['label'])) {
+    if (
```
amitaibu, line 194: matter of taste, but can this be in a single line

tests/RestfulViewEntityTestCase.test
```diff
@@ -168,5 +168,36 @@ class RestfulViewEntityTestCase extends DrupalWebTestCase {
...
+
+    $response = $this->httpRequest('api/v1/articles', \RestfulInterface::GET, NULL, array(
+      'X-Restful-Minor-Version' => 6,
+    ));
+    // Returns something like:
+    // <api>
+    //   <data>
+    //     <item0>
+    //       <id>1</id>
+    //       <label>Humo</label>
```
amitaibu: Is that "Foo" in Spanish? :)

tests/RestfulViewEntityTestCase.test
```diff
@@ -168,5 +168,36 @@ class RestfulViewEntityTestCase extends DrupalWebTestCase {
...
+    //       <self>http://restful.local/node/1</self>
```
amitaibu: `https://example.com/node/1`

tests/RestfulViewEntityTestCase.test
e0ipso: LOL, it's just bad copy & paste.

amitaibu:
Really great PR, fun to review! :)

plugins/formatter/RestfulFormatterHalJson.php
e0ipso: I'll look into it. I should have missed it.

Maybe we should add a ticket to add test coverage to it?

plugins/formatter/RestfulFormatterHalJson.php
amitaibu, line 22: to the render itself maybe?

e0ipso, line 22: > to the render itself maybe?

The problem with that is that since we have an error we don't want to add HATEOAS and use the handler anymore (because it's throwing exceptions), this is a way to return early and abort.

plugins/formatter/RestfulFormatterHalJson.php
e0ipso: Hmm, I don't see it in https://github.com/Gizra/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L1670 Is it possible that this does not need any changes?

amitaibu: I think `getUrl()` should get the `$request` in the signature as it previously did.

plugins/formatter/RestfulFormatterHalJson.php
amitaibu, line 29: > Right now we're using it for CSRF token requests.

So the CSRF is also returning inside `data`? that sounds wrong to me.

plugins/formatter/RestfulFormatterHalJson.php
amitaibu, line 22: ok, lets leave as is for now.

plugins/formatter/RestfulFormatterBase.php
amitaibu, line 48: But I see it also exists [here](https://github.com/Gizra/restful/pull/129/files#diff-4541236b53ea9c32a255545b56e3ba2cR14).

plugins/formatter/RestfulFormatterHalJson.php
e0ipso, line 29: It is returning inside `data`. I think we want to use the same output format, that means that we have to use `restful_delivery`… Maybe we should have something that determines if HAL applies. I'll try to come up with something.

plugins/formatter/RestfulFormatterHalJson.php
e0ipso: I really like what you did of maintaining the state in the class property instead of passing stuff around. I'm going to suggest:

``` php
  protected function addHateoas(array &$data) {
    if (!$this->handler) {
      return;
    }
    $original_request = $this->handler->getRequest();

    $data['_links'] = array();
    $page = !empty($original_request['page']) ? $original_request['page'] : 1;

    if ($page > 1) {
      $request = $original_request;
      $request['page'] = $page - 1;
      // Set the modified request with the updated page so getUrl returns the
      // correct value.
      $this->handler->setRequest($request);
      $data['_links']['previous'] = $this->handler->getUrl();
    }

    // We know that there are more pages if the total count is bigger than the
    // number of items of the current request plus the number of items in
    // previous pages.
    $items_per_page = $this->handler->getRange();
    $previous_items = ($page - 1) * $items_per_page;
    if ($data['count'] > count($data['data']) + $previous_items) {
      $request = $original_request;
      $request['page'] = $page + 1;
      // Set the modified request with the updated page so getUrl returns the
      // correct value.
      $this->handler->setRequest($request);
      $data['_links']['next'] = $this->handler->getUrl();
    }
    // Put the original request back.
    $this->handler->setRequest($original_request);
  }
```

plugins/restful/RestfulEntityBase.php
e0ipso, line 1827: It's a convenient helper function. Imagine a format plugin implementor that needs this to format data depending on this… It can go away since it's not needed, but it sure won't hurt. Your call.

plugins/formatter/RestfulFormatterHalJson.php
amitaibu: I think that since getUrl() is a helper function, it's "legit" to pass it the request. Re-Setting and reverting the request sounds a little harsh for what we want to accomplish here.

restful.module
e0ipso, line 648: That's my goal. But that will require to refactor `RestfulEntityBase` and move all that is not `entity` related to `RestfulBase`. I wanted to keep this PR _as small as possible_.

amitaibu, line 648: I'm not sure I totally follow what you mean, but a small PR sounds good :P

e0ipso, line 648: Opened https://github.com/Gizra/restful/issues/132 and https://github.com/Gizra/restful/issues/133 for this.

restful.module
e0ipso, line 645: The same as before, CSRF calls and such that require a format but don't load a handler.

tests/RestfulSubResourcesCreateEntityTestCase.test
e0ipso, line 194: It gets too long for a non-horizontal scroll reading. I trust this is standard formatting in this cases, no?

plugins/formatter/RestfulFormatterBase.php
e0ipso, line 48: > But I see it also exists here.

Right, that one should go away.

plugins/formatter/RestfulFormatterHalJson.php
e0ipso: Done.

tests/RestfulSubResourcesCreateEntityTestCase.test
amitaibu, line 194: I usually prefer a one liner. Anyway, not a deal breaker ;)

e0ipso:
Tests are green again.

modules/restful_example/plugins/restful/node/articles/1.6/RestfulExampleArticlesResource__1_6.class.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulExampleArticlesResource__1_5.
```
amitaibu: RestfulExampleArticlesResource__1_5 => RestfulExampleArticlesResource__1_6

plugins/formatter/RestfulFormatterHalJson.php
amitaibu, line 29: >  Maybe we should have something that determines if HAL applies. I'll try to come up with something.

Maybe we could (as you did initially) have a JSON base class, that is used by default for non-handlers? In that case the CSRF would use that formatter which will not have the token under `data[0]` - which is weird ;)

tests/RestfulViewEntityTestCase.test
```diff
@@ -168,5 +169,36 @@ class RestfulViewEntityTestCase extends DrupalWebTestCase {
...
+    $response = $this->httpRequest('api/v1/articles', \RestfulInterface::GET, NULL, array(
+      'X-Restful-Minor-Version' => 6,
+    ));
+    // Returns something like:
+    // <api>
+    //   <data>
+    //     <item0>
+    //       <id>1</id>
+    //       <label>Foo</label>
+    //       <self>http://example.com/node/1</self>
```
amitaibu, line 188: minor, but lets change to `https://` (notice the s), so we advocate using https for RESTful endpoints.

e0ipso:
@amitaibu I think this is ready to be merged.

plugins/formatter/RestfulFormatterHalJson.php
amitaibu, line 29: @mateu-aguilo-bosch ^^

restful.module
```diff
@@ -587,7 +638,26 @@ function restful_json_output($var = NULL) {
...
+  // Get the formatter for the current resource.
+  if ($restful_handler = restful_get_restful_handler_for_path()) {
+    $formatter_name = $restful_handler->getPluginInfo('formatter');
+    // Allow the handler to change the HTTP headers.
+    foreach ($restful_handler->getHttpHeaders() as $key => $value) {
+      drupal_add_http_header($key, $value);
+    }
+  }
+  else {
+    $formatter_name = variable_get('restful_default_output_formatter', 'hal_json');
```
amitaibu, line 653: https://github.com/Gizra/restful/pull/129/files#r17127841

Something like this:

``` php
else {
  // Use a simple JSON formatter, for non RESTful handler calls, such as a CSRF token.
  $formatter_name = variable_get('restful_default_output_formatter', 'json');
}
```

amitaibu, line 653: or instead, for simplicity, just keep the delivery callback from CSRF token as it was with `restful_json_output`?

Seems simpler :)

e0ipso, line 653: I'd rather not. It makes no sense to have YAML for everything but csrf tokens. It forces clients to support JSON, and that is against the spirit of this pr.

I'll have a custom delivery that skips the `prepare` method.

amitaibu, line 653: > I'll have a custom delivery that skips the prepare method.

Or maybe in the plugin we can have "formatter_skip_prepare"?

amitaibu:
@mateu-aguilo-bosch Seems ready. Can I merge, or still in work?

e0ipso:
Mergeable.

amitaibu:
@mateu-aguilo-bosch I see something wrong here -- I query for a single entity ID and I get a `count` and `_links`:

local_d7_dev_api_v1_articles_73-3

e0ipso:
Merged by ce02e442435c38429ca76c5d7b856534eef30330.

Created a follow up for the `count` and `_links` issue.

Issue #127, opened 2014-08-28
Create output plugin for output format flexibility.


e0ipso:
I think RESTful should be opinionated about the output format. A variation of [JSON API](http://jsonapi.org/) is my preferred format and I'd like to see the module following that. The only I'd like to have with it is that I think that when getting a specific single item we should not be returning an array with only one item but an object. I also like having `"data"` as a common object key for … well, data.

Having an output plugin and defining our default one (`json_api_custom`) will allow module integrators to easily have APIs in YML, other JSON structures, or `$deity` forbids… XML.

amitaibu:
>  I think that when getting a specific single item we should not be returning an array with only one item but an object. I also like having "data" as a common object key for … well, data.

Indeed, that's what #126 is doing, and I think it's ok to "force" it. If someone insists on changing them they can override the needed methods.

> Having an output plugin and defining our default one (json_api_custom) will allow module integrators to easily have APIs in YML, other JSON structures, or $deity forbids… XML.

Let's avoid this for now. Or put differently, lets consider this when someone asks for it, for a _real_ use case :)

Please re-open if you feel strongly about it.

amitaibu:
Oh, I saw in your comment you are already working on this, so re-opening. I might be missing your point?

amitaibu:
What's your use case for this?

e0ipso:
@amitaibu imagine an EmberJS that gets you very far in client side if your API is returning JSON in a specific format. Imagine some dev that has to implement an API that has already a test suite built around a specific output format, if we don't allow overriding they may need to look somewhere else for the API.

amitaibu:
I can only imagine AngularJs ;)
Ok, as always, your point is valid! :)

e0ipso:
hahaha. AngularJS is the best.

e0ipso:
Closed in favor of https://github.com/Gizra/restful/pull/129

PR #130, opened 2014-08-31
Allow filtering the query for list

amitaibu:
#124


plugins/restful/RestfulEntityBase.php
```diff
@@ -627,10 +642,66 @@ public function getQueryForList() {
...
+   * query range.
+   *
+   * @param \EntityFieldQuery $query
+   *   The query object.
+   *
+   * @throws \RestfulBadRequestException
+   *
+   * @see \RestfulEntityBase::getQueryForList
+   */
+  protected function queryForListPagination(\EntityFieldQuery $query) {
```
e0ipso, line 691: We now (#127) have a method to get the total count. That will allow us to know the number of pages in advance as well as avoiding the `$this->getRange() + 1` trick.

tests/modules/restful_test/plugins/restful/node/test_articles/1.2/RestfulTestArticlesResource__1_2.class.php
```diff
@@ -18,19 +18,37 @@ public function getPublicFields() {
...
-    );
-
-    $public_fields['entity_reference_multiple'] = array(
-      'property' => 'entity_reference_multiple',
-      'resource' => array(
-        'article' => 'test_articles',
-      ),
-    );
+    // By checking that the field exists, we allow re-using this class on
+    // different tests, where different fields exist.
```
e0ipso, line 22: :+1:

e0ipso:
I'm afraid to see this become a query builder. First adding the match for the field contents, then the _contains_, then the _NOT_, then operators with multiple filters, … I feel like we should come with a plan in advance for that.

What's the plan behind this? Do other API projects face this need, how do they solve them?

Is it for user defined filters?

amitaibu:
> What's the plan behind this?

Here's my use case:
- "Company" content type (e.g. puma, adidas)
- "Item" content type (e.g. shirt, pants) that references a company.

`api/v1/items` -- gets all the items
`api/v1/items?filter[company]=1` -- get the items of a single company with ID 1

amitaibu:
@mateu-aguilo-bosch ^^ 
makes sense to you?

e0ipso:
@amitaibu sure. I think that we need to ask ourselves, _Is that the only filtering we'll want to allow?, filtering by exact match without condition composition_.

amitaibu:
> Is that the only filtering we'll want to allow?, filtering by exact match without condition composition.

I've started building it with optionally allowing setting an [operator](https://github.com/Gizra/restful/pull/130/files#diff-3c241a6c3b4e93a88a7da3e91c3e9358R671)

We could also introduce `'allow filter' => TRUE` property, to allow switching it off.

What are your concerns?

e0ipso:
My concerns are that we end up building a _Query Builder_ using EFQ and exposing a dev interface through the URL.
1. If we keep it simple, it makes sense this way.
2. If we go flexible we may need some advanced implementation for this (views does something like that, exposing filters and building queries).

amitaibu:
The (still non-working) PR is currently more or less what I was planning. I'll finish it, and submit it for your review, and we'll figure out if it's too complex, ok?

amitaibu:
This now includes #135.

amitaibu:
@mateu-aguilo-bosch Tests are passing.

(We can wait for #135 and #127 to get it in first)

amitaibu:
PR merged with #129

amitaibu:
Merged.

PR #131, opened 2014-09-01
Fix entityvalidate property condition

snirs90:
This fix the PHP Notice of:
`Notice: Undefined index: property in RestfulEntityBase->entityValidate() (line 1400 of /path/to/project/modules/contrib/restful/plugins/restful/RestfulEntityBase.php).`

When you define a field with just a callback without a field property,
For example:

``` php
$public_fields['phpVersion'] = array(
      'callback' => 'phpversion',
);
```

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1426,7 +1426,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
 
     $map = array();
     foreach ($this->getPublicFields() as $field_name => $value) {
-      if (!$value['property']) {
+      if (!isset($value['property'])) {
         continue;
       }
 
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1426,7 +1426,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
 
     $map = array();
     foreach ($this->getPublicFields() as $field_name => $value) {
-      if (!isset($value['property'])) {
+      if (empty($value['property'])) {
         continue;
       }
 


plugins/restful/RestfulEntityBase.php
```diff
@@ -1426,7 +1426,7 @@ public function entityValidate(\EntityMetadataWrapper $wrapper) {
 
     $map = array();
     foreach ($this->getPublicFields() as $field_name => $value) {
-      if (!$value['property']) {
+      if (!isset($value['property'])) {
```
e0ipso: What about `empty($value['property'])`. This way we're checking for: `NULL`, `''`, `FALSE`, `array()`, …

snirs90: `empty()` is not needed because the entity validation is executed here to validate the entity field data.
The entity validate should throw you an error if the field is required but empty.

e0ipso: I don't follow. `!$value['property']` is basically `empty()` but with PHP warnings for unset values.

Are the PHP warnings the only thing that need to be fixed, or you're also fixing some broken logic?

snirs90: Basically, only the PHP warnings.

e0ipso: Then you need to use `empty()`.

Scenario: `$value['property'] = FALSE`.
- Before: `if (!$value['property'])` returns `TRUE`;
- After: `if (!isset($value['property']))` returns `FALSE`. :red_circle: 
- After: `if (!empty($value['property']))` returns `TRUE`.

snirs90: Sorry, I've got confused with the logic, for some reason I've thought I'm dealing here with the field data.
You're right. thanks for notice.

e0ipso:
Awesome! I just want the module to rock hard! :stuck_out_tongue:

e0ipso:
Thanks @snirs90.

PR #135, opened 2014-09-02
Add \RestfulInterface::publicFieldsInfo() for users to declare their fields

amitaibu:
It's much cleaner that getPublicFields() returns with all the values populated. This also helps with #124, when we need the `$public_field['column']` outside of `entityView()`

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -35,35 +35,39 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
   /**
    * The public fields that are exposed to the API.
    *
-   *  Array with the optional values:
-   *  - "property": The entity property (e.g. "title", "nid").
-   *  - "sub_property": A sub property name of a property to take from it the
-   *    content. This can be used for example on a text field with filtered text
-   *    input format where we would need to do $wrapper->body->value->value().
-   *    Defaults to FALSE.
-   *  - "wrapper_method": The wrapper's method name to perform on the field.
-   *    This can be used for example to get the entity label, by setting the
-   *    value to "label". Defaults to "value".
-   *  - "wrapper_method_on_entity": A Boolean to indicate on what to perform
-   *    the wrapper method. If TRUE the method will perform on the entity (e.g.
-   *    $wrapper->label()) and FALSE on the property or sub property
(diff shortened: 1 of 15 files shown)


amitaibu:
hmm, not sure if the approach is right, need to keep thinking about it, but I'm pretty sure we want `getPublicFields()` to return the 'processed' values.

amitaibu:
@mateu-aguilo-bosch what do you think about [this](https://github.com/Gizra/restful/pull/135/files#diff-3c241a6c3b4e93a88a7da3e91c3e9358R1612) set inside `getPublicFields()`?

e0ipso:
We should probably have 2 separated methods:
- one that the user overrides to add their own fields (they're responsible of calling the parents there). That would return all public field configurations (including defaults like `id`, `self`, …). This one could be called something like `getPublicMappings` (:arrow_left: this name sucks).
- another one that would be used to get all the public fields with the populated values. That method could be `getPublicFields` and it would populate the defaults on the fly or return a cached copy if we already did this. This one would be in charge of the heavy lifting.

e0ipso:
Also, lets take the opportunity here and move `process_callback` to `process_callbacks` and make it an array of callables.

amitaibu:
How about `publicFieldsInfo()` and `getPublicFields()`?

amitaibu:
@mateu-aguilo-bosch I've pushed a new version with your suggestion. Looks cleaner.

> process_callback to process_callbacks and make it an array of callables.

Use case? Also, would you do the same for "callback"?

e0ipso:
Having different modules adding their process callbacks is something typical and handy. For instance one module adds image style URLs to the image fields and another module adds extra metadata gathered from Flickr. Both modules don't know about each other, they just register their process callback.

> Also, would you do the same for "callback"?

I don't see the need of it, but it can be handy (as a way to stack EMW calls one on top of the other?).

plugins/restful/RestfulEntityBase.php
```diff
@@ -1597,6 +1598,55 @@ public function getPublicFields() {
   /**
    * {@inheritdoc}
    */
+  public function getPublicFields() {
+    if ($public_fields = $this->publicFields) {
```
e0ipso: This seems cumbersome. Can we do:

``` php
if ($this->publicFields) {
```

e0ipso:
One minor comment and the `process callbacks` thingy. Then the green button can be clicked.

Issue #137, opened 2014-09-07
Allow "X-HTTP-Method-Override"


amitaibu:
For clients that don't have PATCH/ PUT

amitaibu:
Closed in favor of #138

PR #138, opened 2014-09-07
Allow "X-HTTP-Method-Override" in request.

amitaibu:
#137

Client may send a POST request, but override the method to newer methods (e.g. PATCH, PUT).

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -449,6 +449,16 @@ function restful_menu_access_callback($major_version, $resource_name) {
     return;
   }
 
+  $method = strtoupper($_SERVER['REQUEST_METHOD']);
+  if ($method == \RestfulInterface::POST && !empty($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE'])) {
+    $method = strtoupper($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']);
+  }
+
+  if (!\RestfulBase::isValidMethod($method, FALSE)) {
+    // HTTP method is invalid.
+    return;
+  }
+
   return $handler->access();
 }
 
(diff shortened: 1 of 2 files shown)


amitaibu:

restful.module
```diff
@@ -476,6 +486,10 @@ function restful_menu_process_callback($major_version, $resource_name) {
 
   $method = strtolower($_SERVER['REQUEST_METHOD']);
 
+  if ($method == 'post' && !empty($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE'])) {
```
e0ipso: We probably want to use `\RestfulInterface::POST`.

restful.module
```diff
@@ -476,6 +486,10 @@ function restful_menu_process_callback($major_version, $resource_name) {
...
+    $method = strtolower($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']);
+  }
+
   if ($method == 'options') {
```
e0ipso: We probably want to use `\RestfulInterface::OPTIONS`.

amitaibu: it's because we strtolower() so it can be used later in `$handler-{$method}()`. Do you think the strtolower() should happen only later?

restful.module
```diff
@@ -476,6 +486,10 @@ function restful_menu_process_callback($major_version, $resource_name) {
 
   $method = strtolower($_SERVER['REQUEST_METHOD']);
 
+  if ($method == 'post' && !empty($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE'])) {
+    $method = strtolower($_SERVER['HTTP_X_HTTP_METHOD_OVERRIDE']);
```
e0ipso: Do we want to call `isValidMethod`?

e0ipso:
Minor stuff, then ready to merge.

restful.module
amitaibu: We already do in the access callback -- https://github.com/Gizra/restful/pull/138/files#diff-2dd0d416dd78166fa5d35d3ac72a6878R457

restful.module
amitaibu: @mateu-aguilo-bosch ^^

Issue #139, opened 2014-09-07
Request of a single asset should not yield pagination and count results.


e0ipso:
- [x] Make sure that no count or pagination links are returned.
- [x] Add test coverage.

Related to #129.

e0ipso:
Closed in favor of #140.

PR #140, opened 2014-09-08
Issue #139: Fix count and pagination.

e0ipso:


plugins/restful/RestfulEntityBase.php
```diff
@@ -180,7 +180,7 @@ public function getPath() {
    * @param string $path
    */
   public function setPath($path = '') {
-    $this->path = $path;
+    $this->path = implode(',', array_unique(array_filter(explode(',', $path))));
```
amitaibu, line 183: can you explain the logic here please?

plugins/restful/RestfulEntityBase.php
```diff
@@ -900,7 +910,7 @@ public function viewEntity($entity_id) {
 
       $value = NULL;
 
-      if ($info['callback']) {
+      if (!empty($info['callback'])) {
```
amitaibu: the reason I'm not using `!empty()` on those places, is that it "forces" us to make sure we initialized those values properly in `getPublicFields()` -- otherwise we'll see notices in the tests.

amitaibu:
also, how come this prevents the count and _links?
Maybe worth adding a test to check count doesn't appear in entity/entities view - just in lists?

plugins/restful/RestfulEntityBase.php
e0ipso, line 183: This piece of code removes duplicated ids from a coma encoded list and rencodes it back. This may help to avoid useless processing improving performance a bit and remove duplicate listings.

`1,1,2,3,4,2,3,4,5` :arrow_right: `1,2,3,4,5`.

plugins/restful/RestfulEntityBase.php
e0ipso: Hmmm, I changed this b/c I saw those notices in the tests. Does that mean that we cannot rely on the defaults?

amitaibu: Then the fragile safe guard worked ;) It means getPublicFields() isn't initalizing properly for some reason. Do you want me to check this?

e0ipso: @amitaibu I have no problem tracing it down.

e0ipso:
I think that since we advocated for having an array of a single element (because it made everyone's lifes **way** easier not to code for the special case) here we should make no difference either and also provide `count` and `_links`.

What this PR does is: fixes the counting to limit it to the required entities. Fixes a performance bug where all entities were queried but only the ones requested were output (when dealing with resource/{id} and resource/{id,id,id,…}). It also fixes the problem when duplicated ids are requested.

Please let me know if there is something missing here before we can merge it.

amitaibu:
Ok, explanation makes sense.

I've created https://github.com/mateu-aguilo-bosch/restful/pull/1 -- after that, this can be merged. Thanks :)

e0ipso:
Closing. Committed in 35492ff215477d285bb4dc16b19064819a6faba7

PR #142, opened 2014-09-08
Add utility function to return image URLs based on image styles.

amitaibu:


amitaibu:
(Still untested)

amitaibu:
Tested, waiting for Travis to come back green

e0ipso:
Can we have tests for this somehow?

amitaibu:
We could probably check that the array returns with some values. I'll try.

e0ipso:
Let me have a go at this. Since most client side technologies have problems resizing the images appropriately, I think this is something restful should deal with by default.

amitaibu:
Closed in favor of #145

Issue #132, opened 2014-09-02
Refactor RestfulEntityBase into RestfulBase


e0ipso:
Move everything that is not entity related to `RestfulBase`.

e0ipso:
Closed in favor of #143.

PR #143, opened 2014-09-10
Move non entity related stuff to RestfulBase.

e0ipso:
Issue #132

With this PR code is better structured and it opens the door to non-entity based endpoints (manual queries, views, …).

All that is not directly related to an entity has been moved to the `RestfulBase` class.

@amitaibu it should be pretty easy to review since I only did cut & paste and fixed the coupling with the render cache (and its tests).

### DIFF
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -62,12 +62,14 @@ class RestfulAuthenticationManager extends \ArrayObject {
    *   The request.
    * @param string $method
    *   The HTTP method.
+   * @param boolean $cache
+   *   Boolean indicating if the resolved user should be cached for next calls.
    *
    * @throws RestfulUnauthorizedException
    * @return \stdClass
    *   The user object.
    */
-  public function getAccount(array $request = array(), $method = \RestfulInterface::GET) {
+  public function getAccount(array $request = array(), $method = \RestfulInterface::GET, $cache = TRUE) {
     global $user;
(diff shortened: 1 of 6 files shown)


amitaibu:
Cool, I'll wait for tests to pass.

btw, @mateu-aguilo-bosch I've granted you commit access to https://www.drupal.org/project/restful, which I maintain as a mirror for make files.

e0ipso:
@amitaibu super. How's the replication handled for the d.o module?

amitaibu:
A manual `git push drupal 7.x-1.x` ;)

e0ipso:
@amitaibu off topic. I ripped part of the restful module (it's a total clone) to create https://github.com/mateu-aguilo-bosch/data_mapper which I needed for work. The idea is to have a standard way to get info out of entities, which is handy in many situations (not only for restful).

Would it make sense to have restful depend on that data_mapper and not to worry about the entity mapping part? That would seem a good architecture to me, but it may be a bit harder to maintain.

amitaibu:
Asked differently -- maybe data-mapper can wrap RESTful?

amitaibu:
Also for now (+since things are moving rapidly), I think we should  keep RESTful dependt only on CTools and Entity API.

e0ipso:
> I think we should keep RESTful dependt only on CTools and Entity API.

Makes sense.

e0ipso:
@amitaibu tests are green.

amitaibu:
Merged, thanks.

Issue #134, opened 2014-09-02
Provide image style integration for image fields.


e0ipso:
The idea is to have some extra configuration for the image fields (maybe also think of a generic way to have extra config for fields with nice helper methods) that will list an array of image styles.

That way we can output different image urls for different sizes. I think this is important b/c client apps usually are not great at resizing images and also there's bandwidth to consider.

So when we build the return array for an entity that contains an image field we output the url to the original image and all image styles listed for that field instance for that particular endpoint.

amitaibu:
I've done something like this - see [here](https://github.com/Gizra/garmentbox/blob/overhaul-1/garmentbox/modules/custom/gb_restful/plugins/restful/node/item_variants/1.0/GbItemVariantsResource.class.php#L29) and [here](https://github.com/Gizra/garmentbox/blob/overhaul-1/garmentbox/modules/custom/gb_restful/gb_restful.module#L46)

Do you think we should have something similar to `gb_restful_get_images_styles` in restful.module?

e0ipso:
@amitaibu that approach makes a lot of sense. I'm good with closing this issue with a documentation PR or something more baked in.

amitaibu:
Closed in favor of #142

PR #145, opened 2014-09-11
Issue #134: Add automatic image derivatives.

e0ipso:
Implements #134 as an alternative to #142.

@amitaibu this also has the nice side effect to allow passing arguments to `'callback'` and `'process_callbacks'`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,12 +60,12 @@ By default the RESTful module will expose the ID, label and URL of the entity.
 You probably want to expose more than that. To do so you will need to implement
 the `publicFieldsInfo` method defining the names in the output array and how
 those are mapped to the queried entity. For instance the following example will
-retrieve the basic fields plus the body, tags and images from (probably) an
-article node. The RESTful module will know to use the `MyRestfulPlugin` class
-because your plugin definition will say so.
+retrieve the basic fields plus the body, tags and images from an article node.
+The RESTful module will know to use the `MyRestfulPlugin` class because your
+plugin definition will say so.
 
 ```php
-class MyRestfulPlugin extends \RestfulEntityBase {
+class MyArticlesResource extends \RestfulEntityBase {
 
   /**
    * Overrides \RestfulEntityBase::publicFieldsInfo().
--- a/README.md
+++ b/README.md
@@ -56,6 +56,50 @@ $handler->patch($id, $request);
 ```
 
 ### View an entity
+By default the RESTful module will expose the ID, label and URL of the entity.
+You probably want to expose more than that. To do so you will need to implement
+the `publicFieldsInfo` method defining the names in the output array and how
+those are mapped to the queried entity. For instance the following example will
+retrieve the basic fields plus the body, tags and images from (probably) an
+article node. The RESTful module will know to use the `MyRestfulPlugin` class
+because your plugin definition will say so.
+
+```php
+class MyRestfulPlugin extends \RestfulEntityBase {
+
+  /**
+   * Overrides \RestfulEntityBase::publicFieldsInfo().
+   */
+  public function publicFieldsInfo() {
+    $public_fields = parent::publicFieldsInfo();
+
+    $public_fields['body'] = array(
+      'property' => 'body',
+      'sub_property' => 'value',
+    );
+
+    $public_fields['tags'] = array(
+      'property' => 'field_tags',
+      'resource' => array(
+        'tags' => 'tags',
+      ),
+    );
+
+    $public_fields['image'] = array(
+      'property' => 'field_image',
+      'process_callbacks' => array(
+        array($this, 'imageProcess'),
+      ),
+      // This will add 3 image variants in the output.
+      'image_styles' => array('thumbnail', 'medium', 'large'),
+    );
+
+    return $public_fields;
+  }
+
+}
+```
+
 ```php
 // Handler v1.0
 $handler = restful_get_restful_handler('articles');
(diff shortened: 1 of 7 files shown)


e0ipso:
@amitaibu I think this is the last feature that we need to support before an stable release. Can you think of any other must have?

e0ipso:
W00t! Tests are green in the first try.

modules/restful_example/plugins/restful/node/articles/1.5/articles__1_5.inc
```diff
@@ -11,4 +11,9 @@ $plugin = array(
   'authentication_types' => TRUE,
   'authentication_optional' => TRUE,
   'minor_version' => 5,
+  'options' => array(
+    'image_styles' => array(
```
amitaibu: shouldn't we move this to the fieldsInfo() array, OR, allow overriding these plugin level ones, with field level ones?

plugins/restful/RestfulBase.php
```diff
@@ -315,6 +315,37 @@ public function format(array $data) {
...
+   *     - An array containing an object and a method name of that object.
+   *       Ex: array($this, 'format').
+   *     - An array containing any of the methods before and an array of
+   *       parameters to pass to the callback.
+   *       Ex: array(array($this, 'processing'), array('param1', 2))
+   * @param array $params
+   *   Array of additional parameters to pass in.
+   *
+   * @return mixed
+   *   Whatever the callback returns.
```
amitaibu: The return value of the callback.

plugins/restful/RestfulBase.php
```diff
@@ -315,6 +315,37 @@ public function format(array $data) {
...
+   *
+   * @throws \RestfulException
+   */
+  public static function executeCallback($callback, array $params = array()) {
+    if (!is_callable($callback)) {
+      if (is_array($callback) && count($callback) == 2 && is_array($callback[1])) {
+        return static::executeCallback($callback[0], array_merge($params, $callback[1]));
```
amitaibu, line 343: I understand it, but worth a doc with example.

plugins/restful/RestfulBase.php
```diff
@@ -315,6 +315,37 @@ public function format(array $data) {
...
+      }
+      $callback_name = is_array($callback) ? $callback[1] : $callback;
+      throw new \RestfulException(format_string('Process callback function: @callback does not exists.', array('@callback' => $callback_name)));
```
amitaibu: `Process callback` should probably be changed to just `callback`

plugins/restful/RestfulEntityBase.php
```diff
@@ -1337,9 +1324,15 @@ public function getPublicFields() {
         'column' => FALSE,
       );
 
+      // Get the image styles.
+      $options = $this->getPluginInfo('options');
       if ($info['property'] && !$info['column'] && $field = field_info_field($info['property'])) {
         // Set the column name.
         $info['column'] = key($field['columns']);
+        // If it's an image check if we need to add image style processing.
```
amitaibu, line 1327: this should be outside of the IF statement (as it won't get in if `$info['column']` was explicitly declared).

plugins/restful/RestfulEntityBase.php
```diff
@@ -1373,4 +1366,20 @@ public function isListRequest() {
     return empty($path) || strpos($path, ',') !== FALSE;
   }
 
+  /**
+   * Get the image variants based on the configured styles.
```
amitaibu: Get the image URLs based on the configured image styles.

e0ipso:
@amitaibu this is ready.

README.md
```diff
@@ -56,6 +56,50 @@ $handler->patch($id, $request);
...
+By default the RESTful module will expose the ID, label and URL of the entity.
+You probably want to expose more than that. To do so you will need to implement
+the `publicFieldsInfo` method defining the names in the output array and how
+those are mapped to the queried entity. For instance the following example will
+retrieve the basic fields plus the body, tags and images from (probably) an
+article node. The RESTful module will know to use the `MyRestfulPlugin` class
+because your plugin definition will say so.
+
+```php
+class MyRestfulPlugin extends \RestfulEntityBase {
```
amitaibu: Maybe `MyRestfulPlugin` => `MyArticlesResource` ("best practice" naming convention)

plugins/restful/RestfulEntityBase.php
```diff
@@ -1373,4 +1365,27 @@ public function isListRequest() {
     return empty($path) || strpos($path, ',') !== FALSE;
   }
 
+  /**
+   * Get the image URLs based on the configured image styles.
+   *
+   * @param array $file_array
+   *   The file array from EMW.
```
amitaibu: => `The file array.`

README.md
```diff
@@ -56,6 +56,50 @@ $handler->patch($id, $request);
 ```
 
 ### View an entity
+By default the RESTful module will expose the ID, label and URL of the entity.
+You probably want to expose more than that. To do so you will need to implement
+the `publicFieldsInfo` method defining the names in the output array and how
+those are mapped to the queried entity. For instance the following example will
+retrieve the basic fields plus the body, tags and images from (probably) an
```
amitaibu: we can probably remove the word "probably" :)

amitaibu:
Minor comments, after that can be merged.

README.md
e0ipso: LOL, you're probably right.

e0ipso:
Merge squashed in 93b0b4db4c2b0f803715468ac464cbcd2e57fd34.

PR #146, opened 2014-09-11
Add CSRF token for login with token auth

amitaibu:
Also change `delivery_callback` to `restful_unprepared_delivery`


e0ipso:
Looks good.

PR #150, opened 2014-09-12
Allow referenced resource to show only entity ID(s)

amitaibu:


amitaibu:
Maybe the default of `full_view` should be FALSE by default?

amitaibu:

e0ipso:
This is good to be merged.

I think sub-resources need a bit more flexibility since someone could want to have the URI as an identifier instead of the internal Drupal ID. Or even more, we could want to have true HAL support for embedded resources. Let's talk about this somewhere.

Hypertext Application Language

e0ipso:
Merged!

PR #157, opened 2014-09-15
Use Vary header for minor versions 

e0ipso:
**Edit**: If a header changes the output, then we need to list that header name in the `Vary` so proxy caches and cdns can do a decent work.

Also, the mobi version of github sucks.

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -540,6 +540,12 @@ function restful_menu_process_callback($major_version, $resource_name) {
   $major_version = intval(str_replace('v', '', $major_version));
   $minor_version = !empty($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) && is_numeric($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) ? $_SERVER['HTTP_X_RESTFUL_MINOR_VERSION'] : 0;
   $handler = restful_get_restful_handler($resource_name, $major_version, $minor_version);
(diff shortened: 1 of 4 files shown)


e0ipso:
@amitaibu ready for review.

restful.module
```diff
@@ -540,6 +540,12 @@ function restful_menu_process_callback($major_version, $resource_name) {
   $major_version = intval(str_replace('v', '', $major_version));
   $minor_version = !empty($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) && is_numeric($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) ? $_SERVER['HTTP_X_RESTFUL_MINOR_VERSION'] : 0;
   $handler = restful_get_restful_handler($resource_name, $major_version, $minor_version);
+  // If there is a minor version we need to Vary the response.
+  if ($minor_version) {
+    $headers = $handler->getHttpHeaders();
+    $vary = empty($headers['Vary']) ? '' : $headers['Vary'];
+    $handler->setHttpHeaders('Vary', $vary . ',X-Restful-Minor-Version');
```
amitaibu, line 547: I know very little about Vary. shouldn't we also add the minor version value?

e0ipso, line 547: Nope. That is something Varnish (or whatever) will track on they own. Only header names here.

amitaibu:
Merged and pushed to d.o as-well.

PR #156, opened 2014-09-15
Implement the HEAD method as GET with an empty body.

e0ipso:
As a sensible default.

### DIFF
--- a/modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
+++ b/modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
@@ -7,13 +7,15 @@
 class RestfulExampleRoleResource extends \RestfulEntityBase implements \RestfulEntityInterface {
 
   /**
-   * Overrides \RestfulEntityBase::controllers.
+   * Overrides \RestfulBase::controllersInfo().
    */
-  protected $controllers = array(
-    '' => array(
-      \RestfulInterface::GET => 'getList',
-    ),
-  );
+  public static function controllersInfo() {
+    return array(
+      '' => array(
+        \RestfulInterface::GET => 'getList',
+      ),
+    );
+  }
 
   /**
    * Overrides \RestfulEntityBase::publicFields().
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -124,22 +124,25 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
     parent::__construct($plugin, $auth_manager, $cache_controller);
     $this->entityType = $plugin['entity_type'];
     $this->bundle = $plugin['bundle'];
-    $this->controllers = array(
-      '' => array(
-        // GET returns a list of entities.
-        \RestfulInterface::GET => 'getList',
-        \RestfulInterface::HEAD => 'getList',
-        // POST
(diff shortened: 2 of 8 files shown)


e0ipso:
@amitaibu ready for review.

plugins/restful/RestfulEntityBase.php
```diff
@@ -268,8 +266,10 @@ protected function queryForListSort(\EntityFieldQuery $query) {
       }
     }
     else {
-      // Sort by default using the entity ID.
-      $sorts['id'] = 'ASC';
+      if (!empty($public_fields['id'])) {
```
e0ipso, line 274: This is the case for endpoints like `token_auth`.

plugins/restful/RestfulEntityBase.php
```diff
@@ -145,6 +124,25 @@ public function __construct($plugin, \RestfulAuthenticationManager $auth_manager
     parent::__construct($plugin, $auth_manager, $cache_controller);
     $this->entityType = $plugin['entity_type'];
     $this->bundle = $plugin['bundle'];
+    // Add default controllers.
+    if (empty($this->controllers)) {
+      $this->controllers = array(
```
amitaibu: maybe like the `publicFieldInfo()` -- we should move this to `controllersInfo()``?

plugins/restful/RestfulEntityBase.php
amitaibu, line 274: Worth adding as a comment.

plugins/restful/RestfulEntityBase.php
e0ipso: good point.

e0ipso:
@amitaibu if tests come don't back :red_circle: this can be reviewed again.

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -7,13 +7,15 @@
...
   /**
-   * Overrides \RestfulEntityBase::controllers.
+   * Overrides \RestfulBase::controllersInfo().
    */
-  protected $controllers = array(
-    '' => array(
-      \RestfulInterface::GET => 'getList',
-    ),
-  );
+  public static function controllersInfo() {
```
amitaibu, line 12: why static?

e0ipso, line 12: Since it does not depend on the state of the object instance I made it static. Static methods are often easier to work with since you don't need to instantiate an object, so I try to make static methods whenever I can.

amitaibu, line 12: in that logic should `publicFieldInfo()` be also static?

e0ipso, line 12: Yep

amitaibu, line 12: Ok, opened #163

amitaibu, line 12: @mateu-aguilo-bosch you can go ahead an merge it (+ kindly also push merged to d.o)

e0ipso:
Merged.

PR #81, opened 2014-09-15
Create a plugin process function for each plugin type.

e0ipso:
And move documentation accordingly.

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -16,6 +16,22 @@ function restful_ctools_plugin_directory($module, $plugin) {
 
 /**
  * Add defaults values to the restful related plugins.
+ */
+function restful_plugin_process(&$plugin, $info) {
+  // Common operations.
+  $plugin += array(
+    'description' => '',
+  );
+
+  // Call the plugin specific process functions.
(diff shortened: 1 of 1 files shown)


e0ipso:
This one should be pretty straight forward.

e0ipso:
@amitaibu ready for review.

restful.module
```diff
@@ -16,6 +16,22 @@ function restful_ctools_plugin_directory($module, $plugin) {
...
  * Add defaults values to the restful related plugins.
+ */
+function restful_plugin_process(&$plugin, $info) {
+  // Common operations.
+  $plugin += array(
+    'description' => '',
+  );
+
+  // Call the plugin specific process functions.
+  $function = 'restful_plugin_process_' . $info['type'];
```
amitaibu, line 27: since we have only 2 plugings that need a process, lets avoid function_exist() and just call those explicitly?

e0ipso, line 27: We will have to check the type of the executing plug-in then.

Even if the function is empty it's the best place to document the options of the plug-in.

amitaibu:
Thanks :)

Issue #159, opened 2014-09-15
Hijack api/* to be under RESTful


amitaibu:
Imagine a client calling a non existing resource (e.g. api/v1/article -- instead of articles).

Currently the result is HTML with a 404.
I would expect RESTful to take control and return a proper JSON response.

e0ipso:
Good idea.

PR #160, opened 2014-09-15
Hijack api/* to provide a better error message in case of 404

amitaibu:
#159

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -801,8 +801,8 @@ function restful_page_delivery_callback_alter(&$callback) {
   $base_path = variable_get('restful_hook_menu_base_path', 'api');
 
 
-  if (strpos($_GET['q'], $base_path . '/') !== 0) {
-    // Page doesn't start with the base path (e.g. "api/").
+  if (strpos($_GET['q'], $base_path . '/') !== 0 && $_GET['q'] != $base_path) {
+    // Page doesn't start with the base path (e.g. "api" or "api/").
     return;
   }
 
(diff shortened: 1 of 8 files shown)


restful.module
```diff
@@ -773,6 +777,40 @@ function _restful_get_json_from_menu_status(&$var) {
...
+ * Hijack api/* to be under RESTful. We make sure that any call to api/* pages
+ * that isn't valid, will still return with a well formatted error, instead of
+ * a 404 HTML page.
+ */
+function restful_page_delivery_callback_alter(&$callback) {
+  if (!variable_get('restful_hijack_api_pages', TRUE)) {
+    return;
+  }
+
+  if (strpos($_GET['q'], 'api/') !== 0) {
```
e0ipso: We should use the contents of `'restful_hook_menu_base_path'` variable. I see that `'api/login'`, `'api/file-upload'` and `'api/session/token'` will also need fixing in this respect.

restful.module
```diff
@@ -773,6 +777,40 @@ function _restful_get_json_from_menu_status(&$var) {
...
+
+/**
+ * Implements hook_page_delivery_callback_alter().
+ *
+ * Hijack api/* to be under RESTful. We make sure that any call to api/* pages
+ * that isn't valid, will still return with a well formatted error, instead of
+ * a 404 HTML page.
+ */
+function restful_page_delivery_callback_alter(&$callback) {
+  if (!variable_get('restful_hijack_api_pages', TRUE)) {
```
e0ipso, line 800: We should document this variable and delete it on uninstall.

e0ipso:
Maybe also worth adding a test.

amitaibu:
Fixed comments and added tests - which are passing.

e0ipso:
Merged via 25ad31e897846a42114cc346d51be28b52cdad09.

Thanks!

PR #161, opened 2014-09-15
Ignore filter on non-list query

amitaibu:
@mateu-aguilo-bosch When there's more comments than code, it shows the solution is a hack. But I didn't think of another way to do it that wouldn't be more hackish.

Got better ideas?

### DIFF
--- a/tests/RestfulListTestCase.test
+++ b/tests/RestfulListTestCase.test
@@ -263,7 +263,7 @@ class RestfulListTestCase extends RestfulCurlBaseTestCase {
     // (e.g. /api/v1/articles/1?filter[label]=foo), as this might be called from
     // a formatter plugin, after RESTful's error handling has finished.
     $query = array('filter' => array('invalid-key' => 'foo'));
-    $result = $this->httpRequest('api/v1/articles/', \RestfulInterface::GET, $query);
+    $result = $this->httpRequest('api/v1/articles/1', \RestfulInterface::GET, $query);
     $this->assertEqual($result['code'], '200', 'Invalid filter key was ignored on non-list query.');
   }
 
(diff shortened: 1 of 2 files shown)


tests/RestfulListTestCase.test
```diff
@@ -258,6 +258,13 @@ class RestfulListTestCase extends DrupalWebTestCase {
     catch (\Exception $e) {
       $this->fail('Incorrect exception was thrown on invalid key for filter.');
     }
+
+    // Assert filtering doesn't apply for non-list request
+    // (e.g. /api/v1/articles/1?filter[label]=foo), as this might be called from
+    // a formatter plugin, after RESTful's error handling has finished.
+    $query = array('filter' => array('invalid-key' => 'foo'));
+    $result = $this->httpRequest('api/v1/articles/', \RestfulInterface::GET, $query);
```
e0ipso: Should this be changed to:
`$result = $this->httpRequest('api/v1/articles/1', \RestfulInterface::GET, $query);`?

e0ipso:
This is one of those things that it can be considered a bug and a feature. You ask for an entity of an ID and matching some filters: then there is one or none. That would be OK too.

Anyways, if we want to pursue this further we need to make sure we strip the filter from the `$request` otherwise the cache layer will have an extra useless bit in the `cid` that will split the cache unnecessarily.

amitaibu:
> This is one of those things that it can be considered a bug and a feature.

Indeed, what I'm fixing over here, is that the exception wasn't caught -- so the response wasn't in JSON.

If the client send wrong data (like I did by mistake), it's fine if the cache isn't optimized.

tests/RestfulListTestCase.test
amitaibu: correct, I'll fix :)

amitaibu:
Merged.

PR #162, opened 2014-09-16
move declaration of publicFieldsInfo and getFieldInfo into RestFulEntityInterface

stmh:
Hi, 

I am trying to use restful to implement non-entity-based rest-handlers. I think the interface-declaration is a bit misleading, as publicFields and getFieldInfo look like an implementation-detail of RestFulEntityInterfaces.

Hope, you'll find this pull request valuable.

cheers,

Stephan

### DIFF
--- a/plugins/restful/RestfulEntityInterface.php
+++ b/plugins/restful/RestfulEntityInterface.php
@@ -7,4 +7,21 @@
  */
 
 interface RestfulEntityInterface extends RestfulInterface {
+
+  /**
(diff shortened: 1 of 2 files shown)


e0ipso:
@stmh that is a good point. It's in a gray zone right now since we will need to get fields in the response (JSON) for non Entity backed resources, but "Field" is a very Entity-esc terminology.

Let's wait on this until we see if there is a use case for non-entity resources.

PR #166, opened 2014-09-22
Pass the wrapped entity to a callback method

amitaibu:
I think this was already in the past, no?


amitaibu:
@mateu-aguilo-bosch , ready for review.

e0ipso:
Nice!

Also, I had that fixed in https://github.com/mateu-aguilo-bosch/shapeshifter/blob/7.x-1.x/plugin_types/mapper/ShapeshifterMapperBase.php#L158 That's why I think having the two modules maintained separately and dependent make sense. I will have to maintain Shapeshifter for some client work anyways.

PR #167, opened 2014-09-24
Fix \RestfulEntityBase::getImageUris to work also for multiple value images

e0ipso:
If the field is multiple then we get a notice, as the $file_array is actually an array of arrays.
We should probably add a wrapper method `\RestfulEntityBase::getImageUrisMultiple` that calls `\RestfulEntityBase::getImageUris`

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1420,10 +1420,38 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
     if (empty($image_styles)) {
       return $file_array;
     }
+    // If $file_array is an array of file arrays. Then call recursively for each
+    // item and return the result.
+    if (static::isArrayNumeric($file_array)) {
+      $output = array();
+      foreach ($file_array as $item) {
+        $output[] = $this->getImageUris($item, $image_styles);
+      }
+      return $output;
+    }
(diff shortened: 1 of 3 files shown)


e0ipso:
@amitaibu tests came back green (including the new ones) for this.

plugins/restful/RestfulEntityBase.php
```diff
@@ -1420,10 +1420,38 @@ public function getImageUris(array $file_array, $image_styles) {
...
+  /**
+   * Helper method to determine if an array is numeric.
+   *
+   * @param array $input
+   *   The input array.
+   *
+   * @return boolean
+   *   TRUE if the array is numeric, false otherwise.
+   */
+  protected final static function isArrayNumeric(array $input) {
```
amitaibu, line 1448: Maybe `isArrayNumeric` could be replaced by a simpler `!empty($value['fid'])` in `imageProcess()`?

e0ipso, line 1448: I like the `isArrayNumeric` because it is a general, yet simple, solution. Looking for a specific key is bound to a specific use case. What if we need to act on lists of nodes later on, for instance.

amitaibu, line 1448: Still not a big fan, but I can live with this :)

amitaibu:
Merged, thanks.

Issue #171, opened 2014-09-28
Missing feature setter for sub_property 


ceoaliongroo:
We try to set any sub_property for the property field_address  but look like missing:

``` php
public function publicFieldsInfo()
    {
        $public_fields = parent::publicFieldsInfo();

        // Fields created by the scrapper server, in a post petition.
        $public_fields['place_description'] = array(
            'property' => 'field_address',
            'sub_property' => 'premise',
        );

        return $public_fields;
    }
```

![screen shot 2014-09-28 at 6 11 02 pm](https://cloud.githubusercontent.com/assets/2622152/4434704/dc9e037e-4721-11e4-9f9b-3a7408795470.png)

PR #172, opened 2014-09-28
fix: Add case setter with sub property.

ceoaliongroo:
#171

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -853,6 +853,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
       }
 
       $property_name = $info['property'];
+      $sub_property_name = $info['sub_property'];
       if (!isset($request[$public_field_name])) {
         // No property to set in the request.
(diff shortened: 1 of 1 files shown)


e0ipso:
Awesome @ceoaliongroo, thanks for contributing.

Apparently tests came back red though.

amitaibu:
I'm not sure that's the right approach. As for the address field use case, you can 

``` php
$public_fields['address'] = array(
  'property' => 'field_address',
);
```

And then in the request pass it the whole array.

``` php
$request['address'] = array(
  'region' => '...',
  'country' => '...',
);
```

btw, make sure to set your IDE to 2 spaces instead of tabs.

e0ipso:
@ceoaliongroo thanks for taking the time to contribute! As @amitaibu suggests partial updates are only supported at field level. That means that you can send a `PATCH` request and only send the fields that you want to update.

What you **cannot** do is have partial updates within a field (if that was supported then it would be impossible to fit all the custom field implementations that exist).

e0ipso:
@ceoaliongroo I am going to close this for now, please feel free to reopen if you think that RESTful should support this.

ceoaliongroo:
Great!

Issue #176, opened 2014-09-29
Wrong meta-data delivered on entity-create


amitaibu:
POST `api/v1/articles`:

``` javascript
{
    "data": [
        {
            "id": "76",
            "label": "foo",
            "self": "http://local/d7_dev/node/76"
        }
    ],
    "count": 75,
    "_links": {
        "next": "http://local/d7_dev/api/v1/articles?label=foo&page=2"
    }
}
```

`count` and `_links` are wrong.

PR #177, opened 2014-09-30
Remove count for non get requests

e0ipso:
Issue #176


amitaibu:
Tests seem to fail.

amitaibu:
also, if possible - to prevent regression,  lets add an assert to POST to make sure only `data` appears

e0ipso:
> lets add an assert to POST to make sure only data appears

That is already implicit in the current tests when setting the expected results. This is why I had to fix tests in 39f25a7.

e0ipso:
@amitaibu tests are green now.

plugins/formatter/RestfulFormatterHalJson.php
```diff
@@ -27,8 +27,13 @@ public function prepare(array $data) {
     $output = array('data' => $data);
 
     if (!empty($this->handler)) {
-      // Get the total number of items for the current request without pagination.
-      $output['count'] = $this->handler->getTotalCount();
+      if (method_exists($this->handler, 'isListRequest') && !$this->handler->isListRequest()) {
```
amitaibu, line 30: do we really need to check `method_exists()` -- since this is HalJson, doesn't it imply it's for an entity -- thus extended RestfulEntityBase?

e0ipso, line 30: Not really, we can have HalJson for any kind of result that we need. If we remove this we will need to add it back when we have Search API integrated results or generic query results.

amitaibu:
Merged, thanks.

PR #141, opened 2014-10-06
Allow specifying more options per endpoint (e.g. access callback)

e0ipso:
For example:

``` php
protected $controllers = array(
  '' => array(
    // Shorthand.
    \RestfulInterface::GET => 'getList',
  ),
  '\d+' => array(
    // Verbose
    \RestfulInterface::GET => array(
      'method'=> 'viewEntity',
      'access callback' => array($this, 'accessViewEntity'),
    ),
  ),
);
```

So the access could be moved to it's own method per endpoint.


e0ipso:
@amitaibu please review.

e0ipso:
Tests, including new ones are green.

amitaibu:
Do we want to add the access callback per public-field also in this one, or follow up?

e0ipso:
> Do we want to add the access callback per public-field also in this one, or follow up?

Do you mean additionally to the entity property access? I'm thinking that maybe that is a bit too much and can be handled within the `callback` for the property or even `process callbacks` if it's for removing stuff instead of throwing an exception.

amitaibu:
> if it's for removing stuff instead of throwing an exception.

access callback would be about allowing to access the property - not an exception.

Use case: In a `users` resource I'd like to have the `password` property, so client can register a new user with their password via POST. However, on GET (e.g. `api/v1/users/1`), I'd like to make sure password's access == FALSE

e0ipso:
Will it need to be a callback? Does it suffice to have an optional array of fields to hide? Like:

``` php
array(
  \RestfulInterface::get => array(
    'callback' => 'viewEntities',
    'access callback' => 'callback_for_the_GET_method',
    'hide fields' => array('label', 'password'),
  ),
);
```

amitaibu:
> Will it need to be a callback?

Yes, because maybe you want to deny access only to certain users, by arbitrary conditions.

Also, I think this feature makes a lot of sense, in terms of expected granularity: We have access on the resource level, this PR is for the method level, and next is for the public field level.

e0ipso:
@amitaibu if it has to be a callback then it should be specified in the public field configuration array.

I opened a follow up for this. https://github.com/Gizra/restful/issues/180

e0ipso:
This can be merged.

tests/modules/restful_test/plugins/restful/node/test_articles/1.3/RestfulTestArticlesResource__1_3.class.php
```diff
@@ -0,0 +1,46 @@
...
+   *
+   * @return boolean
+   *   TRUE for access granted, FALSE otherwise.
+   */
+  protected function accessViewEntityFalse() {
+    return FALSE;
+  }
+
+  /**
+   * Custom access callback for the GET method.
```
amitaibu: GET => HEAD

tests/RestfulEntityAndPropertyAccessTestCase.test
```diff
@@ -169,4 +169,23 @@ class RestfulEntityAndPropertyAccessTestCase extends DrupalWebTestCase {
       $this->pass('Non-privileged user cannot view entity.');
     }
   }
+
+  public function testEnpointAccessCallback() {
```
amitaibu, line 176: Can you add PHPdocs please

tests/RestfulEntityAndPropertyAccessTestCase.test
```diff
@@ -169,4 +169,23 @@ class RestfulEntityAndPropertyAccessTestCase extends DrupalWebTestCase {
...
+    $settings = array(
+      'type' => 'article',
+    );
+    $node = $this->drupalCreateNode($settings);
+
+    $handler = restful_get_restful_handler('test_articles', 1, 3);
+    try {
+      $handler->get($node->nid);
+      $this->fail('Custom access callback per endpoint not executed correctly.');
```
amitaibu: Maybe `"Custom access callback per resource's method not executed correctly."` or does endpoint imply a method?

amitaibu:
After the minor comments, this can be merged. Thanks.

e0ipso:
Comment changes done.

amitaibu:
Merged, thanks.

PR #153, opened 2014-10-08
Accept the OPTIONS method for discovery purposes.

e0ipso:
When hitting an endpoint with the OPTIONS method, we should return a list of the allowed verbs on that endpoint in the `Accept` header.


e0ipso:
Ready for review with newly added tests.

This is working towards discoverability and auto-documentation. I think @mirzu will like this one :stuck_out_tongue:.

e0ipso:
Tests are green.

plugins/restful/RestfulBase.php
```diff
@@ -513,6 +513,55 @@ public function delete($path = '', array $request = array()) {
...
+   *
+   * @return NULL
+   *   Only HTTP headers are populated.
+   */
+  public function options($path = '', array $request = array()) {
+    // Make your methods discoverable.
+    $allowed_methods = array();
+    foreach ($this->getControllers() as $pattern => $controllers) {
+      if ($pattern != $path && !($pattern && preg_match('/' . $pattern . '/', $path))) {
+        continue;
```
amitaibu, line 533: worth adding a comment why we continue - the above line is not trivial to understand.

plugins/restful/RestfulBase.php
```diff
@@ -513,6 +513,55 @@ public function delete($path = '', array $request = array()) {
...
+   * @param string $path
+   *   (optional) The path.
+   * @param array $request
+   *   (optional) The request.
+   *
+   * @return NULL
+   *   Only HTTP headers are populated.
+   */
+  public function options($path = '', array $request = array()) {
+    // Make your methods discoverable.
```
amitaibu: Maybe better `// A list of discoverable methods.`

plugins/restful/RestfulBase.php
```diff
@@ -513,6 +513,55 @@ public function delete($path = '', array $request = array()) {
...
+    $allowed_methods = array();
+    foreach ($this->getControllers() as $pattern => $controllers) {
+      if ($pattern != $path && !($pattern && preg_match('/' . $pattern . '/', $path))) {
+        continue;
+      }
+      $allowed_methods = array_keys($controllers);
+      // We have found the controllers for this path. Break.
```
amitaibu: You can probably remove the "Break." form the comment.

plugins/restful/RestfulBase.php
```diff
@@ -513,6 +513,55 @@ public function delete($path = '', array $request = array()) {
...
+      break;
+    }
+    if (!empty($allowed_methods)) {
+      $this->setHttpHeaders('Access-Control-Allow-Methods', implode(',', $allowed_methods));
+    }
+
+    // TODO: Allow configuring the allowed origins and return them here.
+    $this->setHttpHeaders('Access-Control-Allow-Origin', '*');
```
amitaibu, line 544: Maybe this can be a variable?

plugins/restful/RestfulBase.php
```diff
@@ -513,6 +513,55 @@ public function delete($path = '', array $request = array()) {
...
+    $accepted_formats = array();
+    foreach ($formatter_names as $formatter_name) {
+      $formatter = restful_get_formatter_handler($formatter_name, $this);
+      $accepted_formats[] = $formatter->getContentTypeHeader();
+    }
+    if (!empty($accepted_formats)) {
+      $this->setHttpHeaders('Accept', implode(',', $accepted_formats));
+    }
+
+    return;
```
amitaibu: This isn't needed.

plugins/restful/RestfulBase.php
```diff
@@ -513,6 +513,55 @@ public function delete($path = '', array $request = array()) {
...
+   * This is an special method since it does not return anything in the body, it
+   * only provides information about the selected endpoint. The information is
+   * provided via HTTP headers.
+   *
+   * @param string $path
+   *   (optional) The path.
+   * @param array $request
+   *   (optional) The request.
+   *
+   * @return NULL
```
amitaibu: I don't think we need this return, as we don't return anything.

plugins/restful/RestfulBase.php
```diff
@@ -825,4 +874,25 @@ public function cacheInvalidate($cid) {
...
+      // If there is formatter info in the plugin definition, return that.
+      return array($plugin_info['formatter']);
+    }
+    // If there is no formatter info in the plugin definition, return a list
+    // of all the formatters available.
+    $formatter_names = array();
+    foreach (restful_get_formatter_plugins() as $formatter_info) {
+      $formatter_names[] = $formatter_info['name'];
+    }
+    return empty($formatter_names) ? array(variable_get('restful_default_output_formatter', 'hal_json')) : $formatter_names;
```
amitaibu: Can't we just `return $formatter_names;`? Seems it will already include the hal_json. Also, I don't think there can be a (real case) situation where `$formatter_names` is empty.

tests/RestfulDiscoveryTestCase.test
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulDiscoverabilityTestCase
```
amitaibu: => RestfulDiscoveryTestCase

amitaibu:
Added some comments, also needs a re-roll

plugins/restful/RestfulBase.php
e0ipso, line 544: Nah, let's address this properly in https://github.com/Gizra/restful/issues/181.

e0ipso:
@amitaibu tests are green.

amitaibu:
Merged, thanks!

PR #183, opened 2014-10-09
Access origin

e0ipso:
This will also refactor the plugin related tasks to its own class.
Also, it fixes a bug when parsing the minor version header.


modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
```diff
@@ -40,6 +40,9 @@ public function publicFieldsInfo() {
    */
   public function getQueryForList() {
     $query = parent::getQueryForList();
+    if ($this->isEmpty('options')) {
```
amitaibu: Maybe `if (!$options = $this->getPluginInfo('options')) {`?

plugins/rate_limit/RestfulRateLimitManager.php
```diff
@@ -75,7 +59,8 @@ public function getPluginInfo() {
    * @param \stdClass $account
    *   The identified user account for the request.
    */
-  public function __construct($resource, $plugin_info, $account = NULL) {
+  public function __construct($resource, array $plugin_info, $account = NULL) {
+    parent::__construct($plugin_info);
```
amitaibu: seems we sometimes call it `$plugin`, and sometimes `$plugin_info` -- we should probably rename all to `$plugin`

plugins/restful/RestfulBase.php
```diff
@@ -903,4 +877,24 @@ public function formatterNames() {
     return $formatter_names;
   }
 
+  /**
+   * Checks access based on the referer header and the allow_origin setting.
+   *
+   * @return bool
+   *   TRUE if the access is granted. FALSE otherwise.
+   */
+  protected function allowOriginAccess() {
```
amitaibu: Maybe  better name is `accessByAllowOrigin()`?

plugins/restful/RestfulBase.php
```diff
@@ -903,4 +877,24 @@ public function formatterNames() {
...
+    // Check the referrer header and return false if it does not match the
+    // Access-Control-Allow-Origin
+    $referer = $_SERVER['HTTP_REFERER'];
```
amitaibu: Can we move `$_SERVER['HTTP_REFERER']` to the `$request['__application']` -- so the plugins won't need to call PHP globals themself?

plugins/restful/RestfulBase.php
```diff
@@ -903,4 +877,24 @@ public function formatterNames() {
...
+    // If there is no allow_origin assume that it is allowed. Also, if there is
+    // no referer then grant access since the request probably was not
+    // originated from a browser.
+    $origin = $this->getPluginInfo('allow_origin');
+    if ($this->isNull('allow_origin') || $origin == '*' || !$referer) {
```
amitaibu: i'm not sure we really need the `isNull()` and `isEmpty()` -- can't we just `$this->getPluginInfo('allow_origin')`?

plugins/restful/RestfulFilesUpload.php
```diff
@@ -27,10 +27,11 @@ public static function controllersInfo() {
    *   file size.
    * - "scheme": By default the default scheme (e.g. public, private) is used.
    */
-  public function __construct($plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
     parent::__construct($plugin, $auth_manager, $cache_controller);
 
-    if (!$options = $this->getPluginInfo('options')) {
+    $options = $this->getPluginInfo('options');
```
amitaibu: I think $options will be an empty array by default, as we already have the process functions that add default values.

amitaibu:
Also, tests seem to be red

modules/restful_example/plugins/restful/node/per_role_content/1.0/RestfulExampleRoleResource.class.php
e0ipso: ~~`getPluginInfo` will return all the plugin info if it cannot find the key 'options'. This is why we need this.~~

Actually that is just not true.

plugins/rate_limit/RestfulRateLimitManager.php
e0ipso: :+1:

plugins/restful/RestfulBase.php
e0ipso: Good point.

plugins/restful/RestfulBase.php

plugins/restful/RestfulBase.php
e0ipso: Those are utility functions. We expose them of possible use cases and delegate functionality to them in `getPluginInfo`. Useful but not important.

e0ipso:
@amitaibu tests are green and feedback applied.

plugins/restful/RestfulBase.php
amitaibu: I think we can avoid the isEmpty and isNull + be more consistent if we'll have to separate functions:

`function getPlugin()` -- simply return $this->plugin;
`function getPluginKey($key)` -- where key is required.

amitaibu: ^^ @mateu-aguilo-bosch what do you think about this?

e0ipso: cool

e0ipso:
PluginInterface has been redefined and applied throughout the code.

This is ready to be merged.

tests/RestfulDiscoveryTestCase.test
```diff
@@ -45,10 +45,6 @@ class RestfulDiscoveryTestCase extends DrupalWebTestCase {
     }
     $headers = $handler->getHttpHeaders();
     $this->assertEqual($headers['Accept'], implode(',', $accepted_formats), 'Accept header is populated correctly for non configured formatters.');
-
-    // 3. Assert Access-Control-Allow-Origin.
```
amitaibu, line 49: Don't we want to assert this?

amitaibu:
Needs a re-roll.

e0ipso:
@amitaibu re-rolled and added some more tests to the mix.

amitaibu:
Merged, thanks.

e0ipso:
w00t! I'm excited to see the plugin helpers in their own place. That's going to be handy for #187 and company.

Issue #180, opened 2014-10-07
Allow specifying more options per field (e.g. access callback) 


e0ipso:
See https://github.com/Gizra/restful/pull/141#issuecomment-57993393

amitaibu:
closing in favor of #184

PR #184, opened 2014-10-09
Add access-callback per public field

amitaibu:
#180

Next is adding tests.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -29,6 +29,11 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
    * The public fields that are exposed to the API.
    *
    * Array with the optional values:
+   * - "access_callback": A callable to determine if user has access to the
+   *   property. Note that this callback is on top of the access provided by
+   *   entity API, and is used for convenience, where for example write
+   *   operation on a property should be denied only on cretain request
+   *   conditions.
    * - "property": The entity property (e.g. "title", "nid").
    * - "sub_property": A sub property name of a property to take from it the
    *   content. This can be used for example on a text field with filtered text
@@ -589,7 +594,7 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
         $sub_wrapper = $info['wrapper_method_on_entity'] ? $wrapper : $wrapper->{$property};
 
         // Check user has access to the property.
-        if ($property && !$this->checkPropertyAccess($sub_wrapper, 'view', $wrapper)) {
+        if ($property && !$this->checkPropertyAccess($sub_wrapper, 'view', $wrapper, $info)) {
           continue;
         }
 
@@ -855,14 +860,14 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
       $property_name = $info['property'];
       if (!isset($request[$public_field_name])) {
         // No property to set in the request.
-        if ($null_missing_fields && $this->checkPropertyAccess($wrapper->{$property_name}, 'edit', $wrapper)) {
+        if ($null_missing_fields && $this->checkPropertyAccess($wrapper->{$property_name}, 'edit', $wrapper, $info)) {
           // We need to set the value to NULL.
           $wrapper->{$property_name}->set(NULL);
         }
         continue;
       }
 
-      if (!$this->checkPropertyAccess($wrapper->{$property_name}, 'edit', $wrapper)) {
+      if (!$this->checkPropertyAccess($wrapper->{$property_name}, 'edit', $wrapper, $info)) {
         throw new RestfulBadRequestException(format_string('Property @name cannot be set.', array('@name' => $public_field_name)));
       }
 
@@ -1212,14 +1217,20 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
    *   Defaults to "edit".
    * @param EntityMetadataWrapper $wrapper
    *   The wrapped entity.
+   * @param array $info
+   *   The info array of the property, as retrieved from
+   *   \RestfulEntityBase::getPublicFields().
    *
    * @return bool
    *   TRUE if the current user has access to set the property, FALSE otherwise.
    */
-  protected function checkPropertyAccess(EntityMetadataWrapper $property, $op = 'edit', EntityMetadataWrapper $wrapper) {
+  protected function checkPropertyAccess(EntityMetadataWrapper $property, $op = 'edit', EntityMetadataWrapper $wrapper, $info) {
+    if ($info['access_callback'] && !static::executeCallback($info['access_callback'], array($property, $op, $wrapper))) {
+      return FALSE;
+    }
+
     $account = $this->getAccount();
-    // @todo Hack to check format access for text fields. Should be removed once
-    // this is handled properly on the Entity API level.
+    // Check format access for text fields.
     if ($property->type() == 'text_formatted' && $property->value() && $property->format->value()) {
       $format = (object) array('format' => $property->format->value());
       if (!filter_access($format, $account)) {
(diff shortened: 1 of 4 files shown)


amitaibu:

plugins/restful/RestfulEntityBase.php
```diff
@@ -1212,14 +1217,20 @@ public function entityValidate(\EntityMetadataWrapper $wrapper) {
    *   Defaults to "edit".
    * @param EntityMetadataWrapper $wrapper
    *   The wrapped entity.
+   * @param array $public_field_info
```
e0ipso: Why are we passing this along instead of calling `\RestfulEntityBase::getPublicFields()` in the access method?

plugins/restful/RestfulEntityBase.php
```diff
@@ -29,6 +29,11 @@
    * The public fields that are exposed to the API.
    *
    * Array with the optional values:
+   * - "access_callback": A callable to determine if user has access to the
+   *   property. Note that this callback is on top of the access provided by
+   *   entity API, and is used for convenience, where for example write
+   *   operation on a property should be denied only on cretain request
+   *   conditions.
```
e0ipso: Maybe it's worth adding a comment on what are the parameters that the callback gets (additionally to any parameter manually passed in).

plugins/restful/RestfulEntityBase.php
```diff
@@ -29,6 +29,11 @@
    * The public fields that are exposed to the API.
    *
    * Array with the optional values:
+   * - "access_callback": A callable to determine if user has access to the
```
e0ipso: > A callable to determine …

Actually it can be a callable or:

``` php
array($callable, array($arg1, $arg2, …));
```

This is useful when you need to do a generic callback that you then use with different arguments depending on the field.

e0ipso:
Should we be turning this into an array of callbacks (like we did for `process_callback`) to allow stacking common operations?

If I am not mistaken only documentation is needed for that, since `executeCallback` will take care of it. We have to warn the user that subsequent access callbacks get the result of the previous callback. It is their responsibility to combine the results together.

Also, we should probably encourage implementors to use `\RestfulInterface::ACCESS_GRANT`, `\RestfulInterface::ACCESS_DENY`, `\RestfulInterface::ACCESS_UNK` instead of just `TRUE` or `FALSE` as their return values. If a stack of calls returns `\RestfulInterface::ACCESS_UNK` assume `\RestfulInterface::ACCESS_DENY`. `\RestfulInterface::ACCESS_UNK` will be useful for subsequent stacked calls.

e0ipso:
Real use case: _I want to hide a field for a certain group of roles (Group A). But there is another group of roles (Group B) that will be allowed if the user has an email address from the organization. Group C is always allowed.

Implementation:

``` php
  …
  'access_callbacks' => array(
    array($this, 'fieldRoleAccess'),
    array(array($this, 'checkEmailDomain'), array('drupal.org')),
  ),
  …
```

``` php
public function fieldRoleAccess(\EntityMetadataWrapper $property_wrapper, $op, \EntityMetadataWrapper $wrapper) {
  $user = $this->getUser();
  if (is_allowed_role($user)) {
    return \RestfulInterface::ACCESS_GRANT;
  }
  if (is_semi_trusted_role($user)) {
    return \RestfulInterface::ACCESS_UNK;
  }
  return \RestfulInterface::ACCESS_DENY;
}

public function fieldDomainAccess(\EntityMetadataWrapper $property_wrapper, $op, \EntityMetadataWrapper $wrapper, $access, $domain) {
  if ($access == \RestfulInterface::ACCESS_UNK) {
    return check_domain($this->getUser(), $domain) ? \RestfulInterface::ACCESS_GRANT : \RestfulInterface::ACCESS_DENY;
  }
  return $access;
}
```

Group A: DENY, DENY
Group B: UNK, GRANT | DENY
Group C: GRANT, GRANT

**Note**: We could probably use this example in the documentation.

amitaibu:
> We have to warn the user that subsequent access callbacks get the result of the previous callback. It is their responsibility to combine the results together.

That doesn't sound right. I would expect that if there's a single DENY we can fail the request. We could control by variable if `\RestfulInterface::IGNORED` (I think better then UNK) - is enough to get access, or a single GRANT is needed.

amitaibu:
Also, the example makes sense -- Just a note that as a best practice (to avoid overriding the global $user) `$user = $this->getUser();` should be `$account = $this->getUser()`

e0ipso:
> Just a note that as a best practice (to avoid overriding the global $user) $user = $this->getUser(); should be $account = $this->getUser()

Yep, sorry. I tried to throw something together real quick.

e0ipso:
> I would expect that if there's a single DENY we can fail the request.

That works too. There are several competing patterns on this. Maybe we should stick to what node_access does. :+1:

amitaibu:

plugins/restful/RestfulEntityBase.php
```diff
@@ -1203,38 +1213,76 @@ public function entityValidate(\EntityMetadataWrapper $wrapper) {
...
+   * @return bool
+   *   TRUE if the current user has access to set the property, FALSE otherwise.
+   *   The default implementation assumes that if no callback has explicitly
+   *   denied access, we grant the user permission.
+   */
+  protected function checkPropertyAccessByAccessCallbacks($op, $public_field_name, EntityMetadataWrapper $property_wrapper, EntityMetadataWrapper $wrapper) {
+    $public_fields = $this->getPublicFields();
+
+    foreach ($public_fields[$public_field_name]['access_callbacks'] as $callback) {
+      $result = static::executeCallback($callback, array($op, $public_field_name, $property_wrapper, $wrapper));
```
e0ipso, line 1278: I feel that subsequent callbacks should be able to get the result of the access chain as a parameter.

``` php
$result = static::executeCallback($callback, array($op, $public_field_name, $property_wrapper, $wrapper, $result));
```

This way a callback can act only if the previous callback returned IGNORED.

e0ipso:
This looks good. Just one minor change (and the change to the docs that reflect that).

I would be cool if one of the example resources had an example on how to chain property access callbacks. Maybe adapting my earlier example?

plugins/restful/RestfulEntityBase.php
amitaibu, line 1278: I'm not sure about this.

Lets say:
callback1 - returned ALLOW
callback2 - returned DENY

So callback3 knows only about the DENY.
If we wanted we could return the array of results, but then it's still important the order of the callbacks.

That's why I prefer a simpler approach - each callback is responsible for itself. We only care out of the box about an explicit DENY.

amitaibu, line 1278: @mateu-aguilo-bosch ^^

e0ipso, line 1278: Actually we only care about ALLOW and IGNORED.

Passing the previous response is probably enough for 95% of cases. Every callback is responsible to act only on the predecessor's response. There are some edge cases this will not cover, but I think it's ok.

The alternative would be to stack the responses in an array and pass it along (I'm not a big fan of it, but I can live with it).

amitaibu, line 1278: > Actually we only care about ALLOW and IGNORED.

Well, right now the in the implementation allow == ignored.

Anyway, can this be merged? :)

e0ipso, line 1278: > allow == ignored

That is for the final grant. Implementing access callbacks are the ones that care about the difference.

See https://github.com/Gizra/restful/issues/192

> Anyway, can this be merged? :)

Sure, let's not hold this any longer. I'll open a follow up and we can continue the discussion there.

e0ipso:
Merged.

Thanks!

jeff-h:
The example from above is now in the docs at https://github.com/RESTful-Drupal/restful/wiki/Custom-field-access-control

PR #165, opened 2014-10-11
$plugin['cache'] is a reserved property

e0ipso:
`ctools_plugin_get_plugin_type_info()` looks for $plugin['cache'] -- which we currently abuse for Restful's own cache.

We should set `$plugin['cache'] = TRUE`, and change the key to something not already used by CTools


amitaibu:
Thanks!

PR #178, opened 2014-10-11
Add documentation for filtering via API/URL.

e0ipso:
Forgive me if I am missing it, but I don't see any documentation for how to filter entities when making an API request via URL. It would be greatly beneficial to add this to the README.

https://github.com/Gizra/restful#api-via-url

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,6 +254,40 @@ curl https://example.com/api/v1/articles/1,2 \
   -H "X-Restful-Minor-Version: 1"
 ```
 
+### Filtering fields
+Using the ``?fields`` query string, you can decalre which fields should be
+returned.
+
+```shell
+# Handler v1.1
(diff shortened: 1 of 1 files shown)


e0ipso:
Thanks for contributing. You can see the committed changes in https://github.com/Gizra/restful#filtering-fields-1

markdorison:
@mateu-aguilo-bosch This is super helpful as well but I was actually referring to filtering a list similar to: https://github.com/Gizra/restful#filter

Is this possible via URL?

e0ipso:
That is probably a useful bit of information. It is something like:

``` shell
curl https://example.com/api/v1/articles/1?filter[label]=abc
```

Would you mind writing a pull request?

e0ipso:
Thanks @markdorison!

PR #188, opened 2014-10-14
Abstract data provider out

e0ipso:
Introducing the concept of `RestfulDataProviderInterface`. That will allow us to have a data provider for SearchAPI, another for QueryDBTNG, …


e0ipso:
@amitaibu green tests. In the next PR I will add an example of a fully working resource for the variable table.

amitaibu:
Wow, big PR :)

e0ipso:
Nah, it's mostly moving stuff around

amitaibu:
> In the next PR I will add an example of a fully working resource for the variable table.

I'd actually be happy to see that PR before this goes in, just so it would be easier for me to wrap my head around this. Would that be ok?

amitaibu:
(I understand what you do here, I just want to see what it takes for an implementing module to create a new data provider)

plugins/restful/RestfulBase.php
```diff
@@ -141,41 +170,30 @@ public static function cleanRequest(&$request) {
    *   route pattern.
    */
   public static function controllersInfo() {
+    // Provide sensible defaults for the HTTP methods. These methods (index,
```
amitaibu, line 173: I'm not sure "index" really better than "list"?

e0ipso, line 173: I agree 100%. Actually I did use _list_, but it turns out that you cannot use reserved words like _list_ or _delete_ as your method name. I belive that _index_ and _remove_ are close seconds.

e0ipso:
Since there are no broken tests and no big comments in here, can we merge this and fix whatever in #189?

amitaibu:
> no big comments in here

Actually, that's because I've added all the comments in #189. I would prefer to close this one in favor of #188 - as the renames are an important part of the fixes.

plugins/restful/RestfulBase.php
amitaibu, line 173: Are you sure about `delete`, I think it is not reserved.

e0ipso:
Closed in favor of #189.

plugins/restful/RestfulBase.php
amitaibu, line 173: @mateu-aguilo-bosch ^^

PR #189, opened 2014-10-15
Create an example resource for the variables table.

e0ipso:
Based on the work on #188, create a resource that can:
- [x] Get a list of all the variables in the site.
- [x] Get the content of a variable (or set of them) by providing the variable name(s).
- [x] Create a new variable.
- [x] Update an existing variable.
- [x] Delete a variable.
- [x] Explore the enpoints using the `OPTIONS` method.
- [x] Test the endpoints using the `HEAD` method.

All the other features (caching, rate limits, authentication, formatting, …) should be granted automatically without any additional work.

@amitaibu does this sound as a good proof of concept for #188?


amitaibu:
Sounds lovely :)

e0ipso:
To satisfy your curiosity, what you have to do to create a new data provider is do what I did in `RestfulQuery.php` and `RestfulQueryInterface.php`.

Then you need to create a resource as usual, but this time using the new data provider. There is a working example in `RestfulQueryVariable.class.php` and `variables__1_0.inc`.

I know the diff looks gigantic, but most of it is extracting common stuff to reuse code and moving functions around to make them available to the data provider through inheritance (which is something that we should have before but since we only had one particular use case…).

There are no tests yet, but testing CRUD on a table is not complex.

I'm excited about the possibilities that this opens. SearchAPI is now one step away.

``` shell
curl -u admin:admin -H 'X-CSRF-Token: sOwN3YYasoiSDRbQofXq4zpjF8A6QLIztOuUchxqwr4' --data 'name=Foo&value=bar' http://restful.local/api/v1/variables
```

e0ipso:
Also, the diff in here is augmented by all the changes already reviewed in #188.

amitaibu:
Awesome. For the diff, I will look at [this](https://github.com/mateu-aguilo-bosch/restful/compare/mateu-aguilo-bosch:187-abstract-data-provider-out...mateu-aguilo-bosch:189-variables-resource)

plugins/restful/RestfulBase.php
```diff
@@ -141,41 +170,30 @@ public static function cleanRequest(&$request) {
    *   route pattern.
    */
   public static function controllersInfo() {
+    // Provide sensible defaults for the HTTP methods. These methods (index,
```
amitaibu, line 177: I'm not sure "index" is better than "list"?

plugins/restful/RestfulBase.php
```diff
@@ -141,41 +170,30 @@ public static function cleanRequest(&$request) {
...
+        // POST
+        \RestfulInterface::POST => 'create',
+      ),
+      // We don't know what the ID looks like, assume that everything is the ID.
+      '^.*$' => array(
+        \RestfulInterface::GET => 'view',
+        \RestfulInterface::HEAD => 'view',
+        \RestfulInterface::PUT => 'replace',
+        \RestfulInterface::PATCH => 'update',
+        \RestfulInterface::DELETE => 'remove',
```
amitaibu, line 195: `remove` => `delete`? (the D in CRUD)

plugins/restful/RestfulBase.php
```diff
@@ -602,6 +623,20 @@ public function getControllerFromPath() {
   }
 
   /**
+   * Helper method to know if the current request is for a list of entities.
```
amitaibu: `is for a list of entities.` => `is for a list.`

plugins/restful/RestfulBase.php
```diff
@@ -614,6 +649,55 @@ protected function addExtraInfoToQuery(\EntityFieldQuery $query) {
...
 
   /**
+   * Parses the request to get the sorting options.
+   *
+   * @return array
+   *   With the different sorting options.
+   *
+   * @throws \RestfulBadRequestException
+   */
+  protected function parseRequestForListSort() {
```
amitaibu, line 663: `getSortFromRequestForList()` ?

plugins/restful/RestfulBase.php
```diff
@@ -614,6 +649,55 @@ protected function addExtraInfoToQuery(\EntityFieldQuery $query) {
...
+    $request = $this->getRequest();
+    $public_fields = $this->getPublicFields();
+
+    $sorts = array();
+    if (!empty($request['sort'])) {
```
amitaibu:
``` php
if (empty($request['sort'])) {
  return array();
}
```

plugins/restful/RestfulBase.php
```diff
@@ -614,6 +649,55 @@ protected function addExtraInfoToQuery(\EntityFieldQuery $query) {
...
+
+  /**
+   * Parses the request object to get the pagination options.
+   *
+   * @return array
+   *   A numeric array with the offset and length options.
+   *
+   * @throws \RestfulBadRequestException
+   */
+  protected function parseRequestForListPagination() {
```
amitaibu, line 692: same name change as for sorts

plugins/restful/RestfulBase.php
```diff
@@ -929,4 +1013,63 @@ public static function getResourceLastVersion($resource_name, $major_version = N
...
+   * Helper method with the code to run for non implemented CRUD operations.
+   *
+   * @param string $operation
+   *   The crud operation.
+   *
+   * @throws \RestfulNotImplementedException
+   */
+  protected static function missingCrudOperation($operation) {
+    // The default behavior is to not support the crud action.
+    throw new \RestfulNotImplementedException(format_string('The "@method" method is not implemented.', array('@method' => $operation)));
```
amitaibu: maybe also add the provider class name, as this error is most likely to appear for developers.

plugins/restful/RestfulBase.php
```diff
@@ -5,7 +5,10 @@
  * Contains RestfulBase.
  */
 
-abstract class RestfulBase extends RestfulPluginBase implements RestfulInterface {
+abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterface {
+  // The \RestfulDataProviderInterface is not declared as implemented on purpose
```
amitaibu: maybe move to the PHP doc under `@file`

plugins/restful/RestfulBase.php
```diff
@@ -929,4 +1013,63 @@ public static function getResourceLastVersion($resource_name, $major_version = N
...
+
+  /**
+   * Helper method with the code to run for non implemented CRUD operations.
+   *
+   * @param string $operation
+   *   The crud operation.
+   *
+   * @throws \RestfulNotImplementedException
+   */
+  protected static function missingCrudOperation($operation) {
```
amitaibu: `notImplementedCrudOperation()`?

plugins/restful/RestfulEFQ.php
```diff
@@ -0,0 +1,355 @@
...
+   */
+  public function index() {
+    // Defer the actual implementation to \RestfulEntityBase.
+    return $this->getList();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewMultiple(array $ids) {
```
amitaibu: I'm on the fence here regarding this (and below) wrapper(s). Maybe we should just rename  `viewEntities()` to `viewMultiple()` -- since it's already in the context of an entity related class?

plugins/restful/RestfulEFQ.php
```diff
@@ -0,0 +1,355 @@
...
+
+  /**
+   * Get a list of entities.
+   *
+   * @return array
+   *   Array of entities, as passed to RestfulEntityBase::viewEntity().
+   *
+   * @throws RestfulBadRequestException
+   */
+  abstract public function getList();
```
amitaibu: looks like over abstracting it. Can you think of a real use case for this?

plugins/restful/RestfulEFQInterface.php
```diff
@@ -0,0 +1,36 @@
...
+   *
+   * @return EntityFieldQuery
+   *   The EntityFieldQuery object.
+   *
+   * @throws RestfulBadRequestException
+   */
+  public function getQueryCount();
+
+  /**
+   * Helper method to get the total count of entities that match certain
```
amitaibu: Should be one liner:

``` php
// Get the total count of entities that match certain request.
```

plugins/restful/RestfulEFQ.php
```diff
@@ -0,0 +1,355 @@
+<?php
+
+/**
+ * @file
+ * Contains \RestfulEFQ
```
amitaibu: Maybe `RestfulDataProvdierEFQ`?

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
+<?php
+
+/**
+ * @file
+ * Contains \RestfulQuery
+ */
+
+abstract class RestfulQuery extends \RestfulBase implements \RestfulQueryInterface, \RestfulDataProviderInterface {
```
amitaibu: `RestfulQuery` => `RestfulDataProviderDbQuery`?

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+    return $this->idColumn;
+  }
+
+  /**
+   * Set the name of the column in the table to be used as the unique key.
+   *
+   * @param string $idColumn
+   *   The name of the column in the table to be used as the unique key.
+   */
+  public function setIdColumn($idColumn) {
```
amitaibu: idColumn => id_column

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+    return $this->tableName;
+  }
+
+  /**
+   * Set the name of the table to query.
+   *
+   * @param string $tableName
+   *   The name of the table to query.
+   */
+  public function setTableName($tableName) {
```
amitaibu: tableName => table_name

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+   * @var string
+   */
+  protected $tableName;
+
+  /**
+   * The name of the column in the table to be used as the unique key.
+   *
+   * @var string
+   */
+  protected $idColumn;
```
amitaibu: `idColumn` => `columnId` sounds better I think.

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+   * Filter the query for list.
+   *
+   * @param \SelectQuery $query
+   *   The query object.
+   *
+   * @throws \RestfulBadRequestException
+   *
+   * @see \RestfulEntityBase::getQueryForList
+   */
+  protected function queryForListFilter(\SelectQuery $query) {
```
amitaibu: looks like a lot of duplication, would be nicer to avoid it.

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function update($id, $full_replace = FALSE) {
+    $query = db_update($this->getTableName());
+    $query->condition($this->getTableName() . '.' . $this->getIdColumn(), $id);
+
+    // Build the update array.
+    $request = $this->getRequest();
+    static::cleanRequest($request);
```
amitaibu: we should move this somewhere before, so data providers don't have to deal with this.

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function remove($id) {
+    db_delete($this->getTableName())
+      ->condition($this->getIdColumn(), $id)
+      ->execute();
+
+    // Set the HTTP headers.
+    $this->setHttpHeaders('Status', 204);
```
amitaibu: Lets move this to parent, so data providers don't have to know these details.

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+  /**
+   * {@inheritdoc}
+   */
+  public function map($row) {
+    $output = &drupal_static(__CLASS__ . '::' . __FUNCTION__ . '::' . $this->getUniqueId($row));
+    if (isset($output)) {
+      return $output;
+    }
+    // Loop over all the defined public fields.
+    foreach ($this->getPublicFields() as $public_property_name => $info) {
```
amitaibu: public_property_name => public_field_name (although this is probably worth its own issue, about deciding if we should call our _thingy_ a `property` or a `public-field` ;)

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+      // If there is a callback defined execute it instead of a direct mapping.
+      if ($info['callback']) {
+        $value = static::executeCallback($info['callback'], array($row));
+      }
+      // Map row names to public properties.
+      elseif ($info['property']) {
+        $value = $row->{$info['property']};
+      }
+      else {
+        continue;
```
amitaibu: what case is this?

plugins/restful/RestfulQuery.php
```diff
@@ -0,0 +1,436 @@
...
+  /**
+   * Helper method to unserialize an object.
+   *
+   * @param string $data
+   *   The serialized data.
+   *
+   * @return mixed
+   *   The unserialized data.
+   */
+  public static function unserializeData($data) {
```
amitaibu: when will this be needed?

plugins/restful/RestfulQueryInterface.php
```diff
@@ -0,0 +1,47 @@
...
+  /**
+   * Prepares the output array from the database row object.
+   *
+   * @param object $row
+   *   The database row object.
+   *
+   * @return array
+   *   The structured array ready to be formatted.
+   */
+  public function map($row);
```
amitaibu: I think we should be more explicit here - `mapDbRowToPublicFields()`?

plugins/restful/csrf/RestfulCsrfTokenResource.class.php
```diff
@@ -37,4 +37,27 @@ public function access() {
     return (bool) $account->uid && parent::access();
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function index() {
```
amitaibu, line 43: Looks very close to what we do [here](https://github.com/Gizra/restful/pull/189/files#diff-4eea59d0b5a9ec63a5392678cd226c9fR377) -- can't we re-use?

restful.module
```diff
@@ -607,6 +607,10 @@ function restful_menu_process_callback($major_version, $resource_name) {
   $major_version = intval(str_replace('v', '', $major_version));
   $minor_version = !empty($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) && is_numeric($_SERVER['HTTP_X_RESTFUL_MINOR_VERSION']) ? $_SERVER['HTTP_X_RESTFUL_MINOR_VERSION'] : 0;
   $handler = restful_get_restful_handler($resource_name, $major_version, $minor_version);
+  if (!$handler instanceof \RestfulDataProviderInterface) {
+    throw new \RestfulServiceUnavailable('The selected plugin does not implement \RestfulDataProviderInterface.');
```
amitaibu: be nice and add the plugin name :)

amitaibu:
@mateu-aguilo-bosch 
So, after the review I actually wonder if we need this middle-step class of a data provider:
RestfulEntityBase is in fact the EFQ data provider. 
RestfulSolr would be in fact the Solr data provider.
- Is there a likely use case that a non entity will use the EFQ provider, or a non-solr to use the Solr data provider?
- Also, is there a likely case the implementing classes will not implement all CRUD?

I think this PR is awesome, it does a lot of cleanup which we'll definitely want. But I think that we can take a step back in abstraction, as it feels like adding complexity without much gain.

Que pensas?

plugins/restful/RestfulQuery.php
amitaibu: Given the [general](https://github.com/Gizra/restful/pull/189#issuecomment-59258440) comment, maybe `RestfulDbQueryBase`

plugins/restful/RestfulBase.php
e0ipso, line 177: See https://github.com/Gizra/restful/pull/188/files#r18939743

plugins/restful/RestfulBase.php
e0ipso, line 195: See https://github.com/Gizra/restful/pull/188/files#r18939743

plugins/restful/RestfulBase.php

plugins/restful/RestfulBase.php
e0ipso, line 663: Actually I want to convey that this method is just parsing the request object to get the user-provided options. The actual `getSortFromRequestForList` will call this one.

plugins/restful/RestfulBase.php

plugins/restful/RestfulBase.php
e0ipso, line 692: See https://github.com/Gizra/restful/pull/189/files#r18939790

plugins/restful/RestfulBase.php

plugins/restful/RestfulBase.php

plugins/restful/RestfulEFQ.php
e0ipso: :+1:

plugins/restful/RestfulEFQ.php
e0ipso: Agreed! But let's do it in a follow up. This is diff is big enough as is.

We will also be able to remove the abstract methods.

plugins/restful/RestfulEFQ.php
e0ipso: This will be removed in the follow-up from https://github.com/Gizra/restful/pull/189/files#r18939842

This was for convenience reasons: make sure all classes extending `\RestfulEFQ` implement `getList`, `viewEntity`, … since we have implemented the basic `index`, `view`, … for them.

As I said, this will not be an issue after the follow-up.

plugins/restful/RestfulEFQInterface.php
e0ipso: :+1:

plugins/restful/RestfulQuery.php
e0ipso: For me `columnId` sounds like the ID for that column (which is incorrect and does not make sense). I wanted the say the name of the ID column, as in the column that contains the IDs for the records.

plugins/restful/RestfulQuery.php
e0ipso: Damn you, PHPStorm code generator! :-D

:+1:

plugins/restful/RestfulQuery.php
e0ipso: :+1:

e0ipso:
I feel pretty strongly about this. Because classes should be in charge of one simple task, and `RestfulEntityBase` is super fat right now. So, I like the idea of extracting the CRUD work to a different class that handles that. Being able to have several implementing classes (although I agree with you that I don't think there will be any) is just the nice side effect, not the main motivation.

> Also, is there a likely case the implementing classes will not implement all CRUD?

Well, yeah. Imagine an API that aggregates local records along with data from an external service, something like 'Return the local movie metadata and records from IMDB', probably you will not care about writing methods. The reverse example would be a logging API where the endpoint writes to the system log the passed in message, you will use your usual tools to read the logs so you don't care about the read methods or the update methods.

> as it feels like adding complexity without much gain.

I actually feel that it is simplifying the code because we are adding separation of concerns. CRUD related methods are in the Data Provider (which is one command-click away from RestfulEntityBase).

So in summary:

> RestfulEntityBase is in fact the EFQ data provider

In summary, RestfulEntityBase (right now) is _not only_ EFQ, but _also_ autocomplete features, entity api access checkers and random helper methods.

plugins/restful/RestfulQuery.php
e0ipso: :+1:

plugins/restful/RestfulQuery.php
e0ipso: :+1:

plugins/restful/RestfulQuery.php
e0ipso: +1 on consistency. We can change it later to whatever we decide.

plugins/restful/RestfulQuery.php
e0ipso: I don't recall, it can be deleted.

plugins/restful/RestfulQuery.php
e0ipso: I don't think we need it. It was something I was testing. Can be deleted.

plugins/restful/RestfulQueryInterface.php
e0ipso: :+1:

plugins/restful/csrf/RestfulCsrfTokenResource.class.php
e0ipso, line 43: Good catch.

plugins/restful/variable/1.0/RestfulQueryVariable.class.php
```diff
@@ -0,0 +1,27 @@
...
+   * {@inheritdoc}
+   */
+  public function publicFieldsInfo() {
+    return array(
+      'name' => array(
+        'property' => 'name',
+      ),
+      'value' => array(
+        'property' => 'value',
+//        'process_callbacks' => array(
```
e0ipso: Remove commented out code.

restful.module
e0ipso: Damn, I cannot be lazy without you noticing.

plugins/restful/RestfulQuery.php
e0ipso: Not sure I get what the duplicated code is nor where the original is.

e0ipso:
@amitaibu thanks for the thorough review, I know that reviewing code is not the funniest thing in the world so I appreciate it.

amitaibu:
> So, I like the idea of extracting the CRUD work to a different class that handles that

Yes I can see the value in that. What I feel right now is that the distinction between the CRUD class and the data provider (=query) isn't totally clear. I feel that by making this distinction clearer (naming stuff better would probably help a lot), we can continue in this path.

Maybe DataProvider is too generic? Maybe it should be `RestfulCrudEntityBase` which implements `RestfulQueryProviderEFQBase`

plugins/restful/csrf/RestfulCsrfTokenResource.class.php
e0ipso, line 43: On second thought those are very different things. Even if their code is similar it is purely coincidental.

e0ipso:
@amitaibu take a look at the renamed stuff. I expect the tests to come back red still.

e0ipso:
Added cool new tests in https://github.com/Gizra/restful/commit/5a95641b37e4f16878bbcbc87772ea1fbf5ecbb7 :8ball:

plugins/restful/RestfulBase.php
```diff
@@ -549,17 +570,31 @@ public function process($path = '', array $request = array(), $method = \Restful
       $this->getRateLimitManager()->checkRateLimit($request);
     }
 
+    // If it is an update method, then clean the request.
+    if ($method == \RestfulInterface::PUT || $method == \RestfulInterface::PATCH) {
```
amitaibu: it should probably be on POST as-well.

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -0,0 +1,446 @@
...
+   * @param RestfulAuthenticationManager $auth_manager
+   *   (optional) Injected authentication manager.
+   * @param DrupalCacheInterface $cache_controller
+   *   (optional) Injected cache backend.
+   */
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
+    parent::__construct($plugin, $auth_manager, $cache_controller);
+    $this->tableName = $plugin['table_name'];
+    $this->idColumn = $plugin['id_column'];
+    if (!empty($plugin['primary'])) {
```
amitaibu: `$this->primary = !empty($plugin['primary']) ? $plugin['primary'] : FALSE;`

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -0,0 +1,446 @@
...
+  public function remove($id) {
+    db_delete($this->getTableName())
+      ->condition($this->getIdColumn(), $id)
+      ->execute();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function mapDbRowToPublicFields($row, $reset = FALSE) {
```
amitaibu: Can we avoid the `$reset` argument? Can't we use the cached value if `$this->getMethod() == 'get'`?

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -0,0 +1,355 @@
...
+   * Filter the query for list.
+   *
+   * @param \EntityFieldQuery $query
+   *   The query object.
+   *
+   * @throws \RestfulBadRequestException
+   *
+   * @see \RestfulEntityBase::getQueryForList
+   */
+  protected function queryForListFilter(\EntityFieldQuery $query) {
```
amitaibu, line 128: almost all this method and `queryForListSort` exists in `RestfulDataProviderDbQuery` - maybe we can avoid the duplication by moving this up?

tests/RestfulDbQueryTestCase.test
```diff
@@ -0,0 +1,107 @@
...
+      ->fields($mock_data)
+      ->execute();
+    $id = intval($id);
+
+    $this->assertTrue(!empty($id), 'The manual record could be inserted');
+
+    // Get the handler.
+    $handler = restful_get_restful_handler('db_query_test');
+
+    // Testing cRud
```
amitaibu: => `// Testing read.`

tests/RestfulDbQueryTestCase.test
```diff
@@ -0,0 +1,107 @@
...
+    $result = $handler->get($id);
+    $result = $result[0];
+    $this->assertEqual($result['string'], $mock_data['str_field'], 'The record was retrieved successfully.');
+    $this->assertEqual($result['integer'], $mock_data['int_field'], 'The record was retrieved successfully.');
+    $this->assertEqual($result['serialized'], $mock_data['serialized_field'], 'The record was retrieved successfully.');
+
+    // Testing crUd
```
amitaibu: `// Testing update.`

tests/RestfulDbQueryTestCase.test
```diff
@@ -0,0 +1,107 @@
...
+    $handler->post('', $mock_data4);
+    $count = db_select($this->tableName)
+      ->countQuery()
+      ->execute()
+      ->fetchField();
+    $this->assertEqual($count, 1, 'The record was created.');
+
+    // Testing list.
+    $result = $handler->get();
+    $this->assertEqual($result, array($mock_data4), 'All the content listed successfully.');
```
amitaibu, line 105: Well done on the test! :)

tests/modules/restful_test/restful_test.module
```diff
@@ -341,3 +341,47 @@ function restful_test_create_vocabulary_and_terms($machine_name = 'test_vocab',
 
   return $vid;
 }
+
+/**
+ * Instead of implementing hook_schema use this helper function.
```
amitaibu, line 346: I think we don't need to do this, and just use hook_schema() -- I think the time benefit here is probably marginal in comparison to the fact out test suite uses the standard profile instead of the minimal one.

plugins/restful/RestfulDataProviderDbQuery.php
e0ipso: Good call.

plugins/restful/RestfulBase.php
e0ipso: Actually I think POST does not need it b/c of how we are looping the request, but it's probably semantically better.

plugins/restful/RestfulDataProviderEFQ.php
e0ipso, line 128: ~~I can create a `\RestfulDataProviderBase`~~ I will add it to `\RestfulBase` to have this generalized.

tests/modules/restful_test/restful_test.module
e0ipso, line 346: When you are looping through tests that install at test setup the test module and uninstall it at tear down, then you are creating and dropping this table useless **a lot** of times. I'd rather leave it as is.

e0ipso:
@amitaibu tests will come back green now. I had to remove the changes made to put the cleanRequest in a general context since it will overwrite the CSRF token.

This is finally ready to merge! w00t!

e0ipso:
I think this is where I would draw the line where querying capabilities goes.

I can see other contrib being `restful_search_api`, `restful_views`, … do you agree with this policy?

amitaibu:
> I can see other contrib being restful_search_api, restful_views, … do you agree with this policy?

Absolutely.

plugins/restful/variable/1.0/RestfulQueryVariable.class.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+/**
+ * @file
+ * Contains \RestfulQueryVariable
+ */
+
+class RestfulQueryVariable extends \RestfulDataProviderDbQuery implements \RestfulDataProviderDbQueryInterface, \RestfulDataProviderInterface {
```
amitaibu: Lets move this to restful-example module. There could be security implications (we don't check access on set) -- so I want to make it clear it's not part of the core functionality. Maybe also worth also implementing access() method, and allow access only if `user_access('administer site configuration, $account);`

amitaibu:
A wonderful PR!

I've added 2 minor comments (rename => delete; and slightly better handling of the variable access). After that it can be merged :)

e0ipso:
> rename => delete

That's not possible. The `delete` method is taken as a helper (as in the DELETE HTTP method).

Variable is now an example.

amitaibu:
Merged, thanks! Great PR!!

Issue #191, opened 2014-10-14
Implement ng-admin as a (real use case) example


amitaibu:
Apart of a good example, ng-admin sounds like it's doing a lot of heavy lifting. If we'll put a login in front of it, this is a good step to the holy grail of decoupled admins.

I've already had some [problems](https://github.com/marmelab/ng-admin/issues/75) on installation, but admittedly I didn't invest too much time there yet.

amitaibu:
@davidhwang you don't get to sit this one out :)

amitaibu:
Now that there's a [demo](https://github.com/marmelab/ng-admin-demo) repo for ng-admin, I was able to get the first step

![angular_admin-4](https://cloud.githubusercontent.com/assets/125707/4669286/c7923cf8-556c-11e4-86bd-8db3261e5af0.jpg)

e0ipso:
I think that to make a game-changing & kick-ass AngularJS component we need more automatic discovery tools. I can see how you can make an OPTIONS call to the server for a particular endpoint and you get in return the information about the fields that you can get/set for every method.

This way the client side application can build forms **without knowing anything about the API in advance!**. This component could be distributed as a generic starting point for AngularJS form generators.

I'm going to tag these backend issues as _Discovery effort_.

Strategic thoughts?

amitaibu:
> Strategic thoughts?
1. Take it in steps. ng-admin is already doing a lot of heavy lifting, so I wouldn't want to start from scratch
2. Get Drupal people excited about it
3. Get ng-admin people excited about it
4. Standardize and profit :)

e0ipso:
> ng-admin is already doing a lot of heavy lifting

That's awesome, I need to read more about that.

> Get Drupal people excited about it

Working on it…

> Get ng-admin people excited about it

Not sure how to do that. Maybe you have ideas…

> Standardize and profit :)

I think this is a key (unspoken) point. I tried to find standards to follow after every idea I had so we can get things for free (even if we don't know what are those things). Good profit to come behind this project, let's try to deliver and keep the hype high.

amitaibu:
Profit in point 4 == Getting more people on board. Monetary stuff are a bonus ;)

e0ipso:
> Monetary stuff are a bonus ;)

Yeah, I realized a long time ago that monetary profit is something that doesn't drive motivation. I also meant profit as in rewarding experiences in general.

davidhwang:
[...]

PR #193, opened 2014-10-16
Try to decode JSON the request

amitaibu:
lessons learned from [ng-admin](https://github.com/Gizra/restful/issues/191)

Also, ng-admin sends `id: null` on POST. @mateu-aguilo-bosch apart of unset of the `$request['id']` can you think of a nicer way?

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -699,9 +699,14 @@ function restful_parse_request() {
   if (!$request && $query_string = file_get_contents('php://input')) {
     // When trying to POST using curl on simpleTest it doesn't reach
     // $_POST, so we try to re-grab it here.
-    parse_str($query_string, $request);
+    // Also, sometimes the client might send the input still encoded.
+    if ($decoded_json = drupal_json_decode($query_string)) {
+      $request = $decoded_json;
+    }
+    else {
+      parse_str($query_string, $request);
+    }
   }
-
 
   // This flag is used to identify if the request is done "via Drupal" or "via
   // CURL";


e0ipso:
Are we supporting JSON encoded input now?

Handling a particular use case for ng-admin makes me uneasy. Also what if someone expects to have the `'id'` field populated?. My take is that for this cases the implementor module is in charge of implementing [`hook_restful_parse_request_alter`](https://github.com/Gizra/restful/blob/7.x-1.x/restful.api.php#L14-L24); to remove whatever they need.

amitaibu:
> Are we supporting JSON encoded input now?

That's actually non ng-admin specific, but more AngularJs specific. Seems that `$http` doesn't send the params as you would expect (so usually I see [this](http://stackoverflow.com/a/22807867/750039) workaround being used). So the `drupal_json_decode()` is just to be a little nicer to Angular, without a big performance penalty. 

> My take is that for this cases the implementor module is in charge of implementing hook_restful_parse_request_alter; to remove whatever they need.

I agree, that's indeed a better idea.

e0ipso:
> but more AngularJs specific

Well, we are supporting the PHP specific urlencoded crap in that gets turned into the $_POST superglobal… Anyways, you can do that also in `hook_restful_parse_request_alter` but I'm OK on providing this as a default since I cannot see a real use case for payloads in XML or HAL or proprietary formats (excel? xD).

restful.module
```diff
@@ -697,11 +697,16 @@ function restful_parse_request() {
   }
 
   if (!$request && $query_string = file_get_contents('php://input')) {
-    // When trying to POST using curl on simpleTest it doesn't reach
-    // $_POST, so we try to re-grab it here.
-    parse_str($query_string, $request);
-  }
 
+    if ($decoded_json = drupal_json_decode($query_string)) {
```
e0ipso, line 703: We can keep this part.

plugins/restful/RestfulBase.php
```diff
@@ -131,6 +131,9 @@ public function setRequest(array $request = array()) {
    */
   public static function cleanRequest(&$request) {
     unset($request['__application']);
+
+    // @todo: How to get this in a clean way?
+    unset($request['id']);
```
e0ipso: Let's remove this.

amitaibu:
Ready for review.

amitaibu:
@mateu-aguilo-bosch ^^

PR #194, opened 2014-10-18
Field definition discovery.

e0ipso:
Add some way to describe the field definitions to the API clients.


amitaibu:
I've already got some code, I'll attach later

‫ב-18 באוק 2014, בשעה 18:09, ‏Mateu Aguiló Bosch notifications@github.com כתב/ה:‬

e0ipso:
@amitaibu argh! I've just seen your comment above…

New tests added.

amitaibu:
Your code is actually nicer, (I've implemented it in a very entity centric manner inside a `formatter::prepare()` for a very specific use case).

plugins/restful/RestfulBase.php
```diff
@@ -530,6 +534,36 @@ public function options($path = '', array $request = array()) {
       $this->setHttpHeaders('Accept', implode(',', $accepted_formats));
     }
 
+    $output = array();
```
amitaibu, line 559: Lets move to own method, for easier override in RestfulEntityBase (`getFieldSchemaForOptionsMethod()`?)

plugins/restful/RestfulBase.php
```diff
@@ -530,6 +534,36 @@ public function options($path = '', array $request = array()) {
...
+        'description' => '',
+      ),
+      // Describe the data.
+      'data' => array(
+        'type' => NULL,
+        'read_only' => FALSE,
+        'cardinality' => 1,
+      ),
+      // Information about the form element.
+      'form' => array(
```
amitaibu: form => form_element

plugins/restful/RestfulBase.php
```diff
@@ -530,6 +534,36 @@ public function options($path = '', array $request = array()) {
...
+        'type' => NULL,
+        'default_value' => '',
+        'placeholder' => '',
+        'size' => NULL,
```
amitaibu, line 578: Lets also add `allowed_values` for select lists.

plugins/restful/RestfulBase.php
```diff
@@ -530,6 +534,36 @@ public function options($path = '', array $request = array()) {
...
+      ),
+    );
+    foreach ($this->getPublicFields() as $public_field => $field_info) {
+      if (empty($field_info['discovery'])) {
```
amitaibu, line 584: we'll need to take care in entity related stuff to check if user has access to the public_field, so we don't expose fields that user can't actually change (+This might be complicated, and I'm not sure how to tackle, as for example a user may be able to create field `foo` but not edit it).

plugins/restful/RestfulEntityBase.php
```diff
@@ -1370,10 +1370,37 @@ public function publicFieldsInfo() {
...
+        ),
       ),
       'label' => array(
         'wrapper_method' => 'label',
         'wrapper_method_on_entity' => TRUE,
+        'discovery' => array(
+          // Information about the field for human consumption.
+          'info' => array(
+            'name' => t('Label'),
+            'description' => t('The entity label.'),
```
amitaibu: Lets avoid the word entity in the output. Maybe "The label of the resource"?

plugins/restful/RestfulEntityBase.php
```diff
@@ -1370,10 +1370,37 @@ public function publicFieldsInfo() {
...
+          ),
+          // Describe the data.
+          'data' => array(
+            'type' => 'string',
```
amitaibu, line 1075: `required => TRUE`

plugins/restful/RestfulEntityBase.php
```diff
@@ -1370,10 +1370,37 @@ public function publicFieldsInfo() {
...
+          ),
+          // Information about the form element.
+          'form' => array(
+            'type' => 'texfield',
```
amitaibu, line 1079: `size => 255` (or should this be in the data)?

also, if in the data `required == true`, do we need to repeat it in the form_element?

tests/RestfulDiscoveryTestCase.test
```diff
@@ -22,34 +22,114 @@ class RestfulDiscoveryTestCase extends RestfulCurlBaseTestCase {
...
-    $handler = restful_get_restful_handler('test_articles', 1, 2);
-    $handler->options('');
-    $accepted_formats = array();
-    foreach ($handler->formatterNames() as $formatter_name) {
-      $formatter = restful_get_formatter_handler($formatter_name, $handler);
-      $accepted_formats[] = $formatter->getContentTypeHeader();
-    }
-    $headers = $handler->getHttpHeaders();
-    $this->assertEqual($headers['Accept'], implode(',', $accepted_formats), 'Accept header is populated correctly for non configured formatters.');
+//  function testOptionsMethod() {
```
amitaibu: We probably still need this test :P

tests/RestfulDiscoveryTestCase.test
```diff
@@ -22,34 +22,114 @@ class RestfulDiscoveryTestCase extends RestfulCurlBaseTestCase {
...
+        ),
+        'info' => array(
+          'description' => t('This field holds different text inputs.'),
+          'name' => t('Text multiple'),
+        ),
+      ),
+    );
+
+    foreach ($expected as $public_field => $discovery_info) {
+      $r_data = $result[$public_field]['data'];
```
amitaibu: => `$result_data`
=> `$expected_data`

plugins/restful/RestfulBase.php
amitaibu, line 559: ignore this ^^, I mis-read it.

plugins/restful/RestfulEntityBase.php
e0ipso, line 1079: `size => 255` can be in both. I am not sure if exposing so many details about the data backend is secure or not.

Since these arrays are exposed as declared to the client, implementors can add as many fields and/or sections as they want. I only think that we ought to provide some sensible defaults for the basics.

In summary, I don't really care if we put it in both places. I think it should be (at least) in the form element.

plugins/restful/RestfulEntityBase.php
e0ipso, line 1075: In this case I think that required is something for the form element. The **entity validator** will take care of the required properties for entity creation.

plugins/restful/RestfulEntityBase.php
e0ipso: :+1:

tests/RestfulDiscoveryTestCase.test
e0ipso: Nah! You sure? :stuck_out_tongue:

plugins/restful/RestfulBase.php
e0ipso, line 584: I agree, but this sound like something that we can do in a follow up. Let's try to keep PRs as simple as possible.

plugins/restful/RestfulBase.php
e0ipso, line 578: :+1:

plugins/restful/RestfulBase.php
e0ipso, line 584: See: https://github.com/Gizra/restful/issues/197

plugins/restful/RestfulBase.php

e0ipso:
@amitaibu, @nader77 please review.

amitaibu:
@mateu-aguilo-bosch Maybe  this PR can include also the field discovery @nader77 did in https://github.com/mateu-aguilo-bosch/restful/pull/2 ?

amitaibu:
@nader77 can you please take care of the failing tests (due to the array keys rename).

e0ipso:
This is the documentation PR: https://github.com/Gizra/restful/pull/199

The actual docs: https://github.com/mateu-aguilo-bosch/restful/blob/document-discovery-options/README.md#documenting-your-api

amitaibu:
I'll work on the tests.

amitaibu:
@mateu-aguilo-bosch I believe we can get this in, right?

e0ipso:
Merged.

Thanks!

PR #195, opened 2014-10-18
WIP: Self url should point to the RESTful URL.

amitaibu:
@mirzu pointed out that currently self isn't correct.

The only problem remaining is that the minor version (passed via headers) isn't taken into account, so the URL only represents the major version.

local_d7_dev_api_v1_articles_page_2-2

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1375,7 +1375,9 @@ abstract class RestfulEntityBase extends RestfulBase implements RestfulEntityInt
         'wrapper_method' => 'label',
         'wrapper_method_on_entity' => TRUE,
       ),
-      'self' => array('property' => 'url'),
+      'self' => array(
+        'callback' => array($this, 'getEntitySelf'),
+      ),
     );
 
     if (!empty($entity_info['entity keys']['label'])) {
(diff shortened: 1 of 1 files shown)


e0ipso:
@amitaibu that exposes a deeper problem about the versioning. It feels weird to have to provide different bits of the version at different places.

There are 4 alternatives that I can see:
1. Leave it as is.
2. Add the minor version to the URL: `https://example.org/api/v1.2/articles/70`.
3. Add the major version to a header. Maybe also combine both headers into one parseable string.
4. A combination of 2 and 3. If the version string is in the URL (detected by regexp) then use that. If not present, then inspect the headers and use those, if both are absent then use the latest version.

I like 4 the best since it allows you to:
- Be more RESTful (and benefit from 3rd parties that assume a given URL structure). Not to win a medal :wink:
- Be able to deep link to an endpoint at a certain version.
- Be able to always use the latest version without even knowing what that one is. Not sure if that is actually useful or not, but I imagine it can be in certain scenarios where the client deploys are complex.

With all this, if we go with option 4, then this issue is unblocked. It would mean a moderate rewrite of the codebase, but I think I can tackle that one.

@amitaibu, @mirzu thoughts?

amitaibu:
> It feels weird to have to provide different bits of the version at different places.

It's taken from [here](http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#versioning)

A combination of 2 + 3 indeed sounds pretty robust. We could also easily make it configurable. Still how do you show a `self` link if the URL options would be turned off?

> Be able to always use the latest version without even knowing what that one is

Github is doing that, and it sounds right to me.

e0ipso:
This should be pretty trivial when https://github.com/Gizra/restful/pull/202 gets merged in.

e0ipso:
Closing in favor of #205. That PR uses the new capabilities of #202.

Issue #197, opened 2014-10-19
Discovery fields filtered by HTTP method.


e0ipso:
Take care of which fields are available for a certain user (also, be explicit about the fact that those are the fields **for that user**).

This comes off of https://github.com/Gizra/restful/pull/194/files#r19055018

PR #198, opened 2014-10-19
Discovery: Add special endpoint to get all the resources.

e0ipso:
Broken down to version and sub version. To dig for more information one would need to make an OPTIONS call to the particular resource (with the version options).

This and https://github.com/Gizra/restful/issues/197 are the _basics_ for discovery features that I can see. With this in place I can see a `restful_apiary` contrib module that generates automatic up-to date documentation.

### DIFF
--- a/plugins/restful/discovery/RestfulDiscoveryResource.class.php
+++ b/plugins/restful/discovery/RestfulDiscoveryResource.class.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulDiscoveryResource.
+ */
+
+class RestfulDiscoveryResource extends \RestfulBase implements \RestfulDataProviderInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function publicFieldsInfo() {
+    return array(
+      'resources' => array(
+        'callback' => 'static::getResourceInfo',
+      ),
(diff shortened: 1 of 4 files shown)


e0ipso:
This was easier than expected :-)

amitaibu:
Nice. I would take a different approach, and treat the discovery resource like any other resource:
1. Add a RestfulDataProviderCToolsPlugin (which is read-only). btw, we can move all data provider specific options to `data_provider_options` in the `$plugin`
2. Add a resource, with explicit public fields (name, description, and URL IMO are enough), instead of one single callback where we unset stuff
3. Let RESTful declare the menu item, so it's api/v1/discovery  -- and the result is still under the `data` key with `count` and `_links` present

e0ipso:
I like this approach much better!

Please disregard this PR when implementing this.

e0ipso:
I used your approach in the docs for this: https://github.com/Gizra/restful/pull/199

The actual docs: https://github.com/mateu-aguilo-bosch/restful/blob/document-discovery-options/README.md#documenting-your-api

amitaibu:
> Please disregard this PR when implementing this.

No problem, I'll give it a stab.

amitaibu:
Closing in favor of #200

Issue #201, opened 2014-10-23
Refactor version negotiation.


e0ipso:
Comes from:

https://github.com/Gizra/restful/pull/195#issuecomment-60211736

e0ipso:
This is done with a lot of sweat.

PR #200, opened 2014-10-20
Discovery resource

amitaibu:
#198

Apart of adding test, I still need to figure out how to do sorting on multiple values. But the filtering works, so you can already get all resource with `minor_version > 4`

``` php
  $handler = restful_get_restful_handler('discovery');
  $request = array(
    'sort' => '-name',
    'filter' => array(
      'minor_version' => array(
        'value' => '4',
        'operator' => '>=',
      ),
    ),
  );
  $result = $handler->get('', $request);
```

PR #202, opened 2014-10-23
Refactor version negotiation.

e0ipso:
#201

### DIFF
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -1103,6 +1103,104 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
   }
 
   /**
+   * Gets the major and minor version for the current request.
+   */
+  public static function getVersionFromRequest() {
+    $version = &drupal_static(__CLASS__ . '::' . __FUNCTION__);
+    if (isset($version)) {
+      return $version;
+    }
+    $router_item = static::getMenuItem();
+    list($resource_name, $version) = $router_item['page_arguments'];
+    if ($version[0] == 'v') {
+      $version = static::parseVersionString($version, $resource_name);
+      return $version;
+    }
+    // If there is no version in the URL check the header.
+    if (!empty($_SERVER['HTTP_X_RESTFUL_VERSION'])) {
+      $version =  static::parseVersionString($_SERVER['HTTP_X_RESTFUL_VERSION'], $resource_name);
+      return $version;
+    }
+    // If there is no version negotiation information return the latest version.
+    $version =  static::getResourceLastVersion($resource_name);
+    return $version;
+  }
+
+  /**
+   * Get the non translated menu item.
+   *
+   * @param null $path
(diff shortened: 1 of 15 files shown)


plugins/restful/RestfulBase.php
```diff
@@ -1103,6 +1103,104 @@ public static function getResourceLastVersion($resource_name, $major_version = N
...
+   * Gets the major and minor version for the current request.
+   */
+  public static function getVersionFromRequest() {
+    $version = &drupal_static(__CLASS__ . '::' . __FUNCTION__);
+    if (isset($version)) {
+      return $version;
+    }
+    $router_item = static::getMenuItem();
+    list($resource_name, $version) = $router_item['page_arguments'];
+    if ($version[0] == 'v') {
```
amitaibu: Do we want to support all options? We could have a variable to choose the versioning method, so we could avoid guess work.

plugins/restful/RestfulBase.php
```diff
@@ -1103,6 +1103,104 @@ public static function getResourceLastVersion($resource_name, $major_version = N
...
+   * Get the non translated menu item.
+   *
+   * @param null $path
+   *   The path to match the router item. Leave it empty to use the current one.
+   * @return array
+   *   The page arguments.
+   *
+   * @see menu_get_item().
+   */
+  public static function getMenuItem($path = NULL) {
```
amitaibu, line 1148: why not call menu_get_item() directly - to avoid the possible language prefix?

plugins/restful/RestfulBase.php
amitaibu: If we would go with such a variable, I think the default one should be: no versioning in the URL, get major + minor via headers, and assume latest if nothing in the headers.

amitaibu: @mateu-aguilo-bosch @mirzu ^^

e0ipso: I really like the idea of benefiting from both versioning options at the same time. I'd rather keep support for both.

e0ipso: I'll change `$version[0] == 'v'` by `preg_match('/^v\d+(\.\d+)?$/', $version)` to be more precise.

plugins/restful/RestfulBase.php
e0ipso, line 1148: It was causing a recursion loop because of the access callback calling this. Do you think we need to take into account the language prefix to extract the _page arguments_ and the _page callback_?

e0ipso:
@amitaibu tests are green now.

amitaibu:
@mateu-aguilo-bosch can you add specific test for the new negotiation? -- it would make also the review easier :)

e0ipso:
@amitaibu tests added.

e0ipso:
I'm merging this one to unblock #200 and #195.

e0ipso:
Merged.

amitaibu:
Awesome PR!

PR #205, opened 2014-10-25
Self url should point to the RESTful URL.

e0ipso:


e0ipso:
articles 2014-10-25 17-44-14

Issue #207, opened 2014-10-25
api/error-resource is no longer treated correctly


amitaibu:
`Fatal error: Call to a member function getMethod() on a non-object in ../restful/restful.module on line757`

amitaibu:
I'll also improve the `testNotFoundDelivery` test case, as obviously it didn't catch the regression.

PR #209, opened 2014-10-25
Fix handling of 404 errors

amitaibu:
#207


amitaibu:
I think [this](https://github.com/Gizra/restful/pull/209/files#diff-2dd0d416dd78166fa5d35d3ac72a6878R771) code is never tested, we should probably add a test as a follow up.

amitaibu:
> I think this code is never tested, we should probably add a test as a follow up.

Sorry, we actually do check this in `\RestfulViewEntityTestCase::testHeaders`.

amitaibu:

e0ipso:
Cool! This is almost the same I had in https://github.com/Gizra/restful/pull/200/files#diff-2dd0d416dd78166fa5d35d3ac72a6878L754 but better.

e0ipso:
Merged!

Thanks.

PR #208, opened 2014-10-26
_links URLs are wrong

e0ipso:
`api/articles`

local_d7_dev_api_articles_


amitaibu:
@mateu-aguilo-bosch ^^ this seems to be related to the `$plugin['menu_item']` change -- would you mind giving it a look?

e0ipso:
We should also check the test coverage for that. I would expect tests to fail for the pagination generation.

e0ipso:
@amitaibu this should be fixed now.

amitaibu:
Merged, thanks.

Issue #214, opened 2014-10-26
Only display the latest resource on the discovery endpoint.


e0ipso:
See: https://github.com/Gizra/restful/pull/200#issuecomment-60497108

mirzu:
I think this can be closed. . . right?

PR #216, opened 2014-10-26
Show only latest resource by default in discovery

amitaibu:
#214

Still need to add a test.

### DIFF
--- a/tests/RestfulDataProviderCToolsPluginsTestCase.test
+++ b/tests/RestfulDataProviderCToolsPluginsTestCase.test
@@ -69,10 +69,10 @@ class RestfulDataProviderCToolsPluginsTestCase extends DrupalWebTestCase {
     );
     $result = $handler->get('', $request);
 
-    $this->assertEqual(count($result), 1, 'Latest resource shown by default.');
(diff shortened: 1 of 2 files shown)


amitaibu:
@mateu-aguilo-bosch @mirzu Here's indeed a place where a human-readable hint about the `?all=true` would be nice.

Question is where to add this? I would expect in the formatter's prepare function, but the Restful handler doesn't (yet?) have control over it.

amitaibu:
@mateu-aguilo-bosch tests are green.

tests/RestfulDataProviderCToolsPluginsTestCase.test
```diff
@@ -27,6 +27,20 @@ class RestfulDataProviderCToolsPluginsTestCase extends DrupalWebTestCase {
 
     // Assert sorting and filtering works as expected.
     $request = array(
+      // Get all resources.
+      'filter' => array(
+        'resource' => array(
+          'value' => 'articles',
+          'operator' => '=',
```
e0ipso, line 34: Is this necessary? Anyways, I like being declarative.

e0ipso:
Fantástico!

Merging.

PR #218, opened 2014-10-27
Return resource version in the header

e0ipso:
When one does `api/articles` they should know which resource's version served their response.

So we can return the version for all cases (e.g. also in `api/v1.0/articles`)

### DIFF
--- a/tests/RestfulViewEntityTestCase.test
+++ b/tests/RestfulViewEntityTestCase.test
@@ -274,10 +274,16 @@ class RestfulViewEntityTestCase extends RestfulCurlBaseTestCase {
 
     // When there is no header version passed in there is no need of Vary.
     $response = $this->httpRequest('api/v1.1/articles');
-    $this->assertFalse(preg_match('/X-Restful-Version/', $response['headers']), 'The Vary header was not added.');
+    $this->assertFalse(preg_match('/X-API-Version/', $response['headers']), 'The Vary header was not added.');
 
-    $response = $this->httpRequest('api/articles', \RestfulInterface::GET, NULL, array('X-Restful-Version' => 'v1.1'));
-    $this->assertTrue(preg_match('/X-Restful-Version/', $response['headers']), 'The Vary header was added.');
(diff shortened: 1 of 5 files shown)


e0ipso:
I also took the opportunity to change our custom version header to use something more widely spread:

`X-Restful-Version` by `X-API-Version`.

https://www.npmjs.org/package/restful-versioning

tests/RestfulViewEntityTestCase.test
```diff
@@ -274,10 +274,16 @@ class RestfulViewEntityTestCase extends RestfulCurlBaseTestCase {
...
-    $response = $this->httpRequest('api/articles', \RestfulInterface::GET, NULL, array('X-Restful-Version' => 'v1.1'));
-    $this->assertTrue(preg_match('/X-Restful-Version/', $response['headers']), 'The Vary header was added.');
+    // Make sure that the version header is present in the response.
+    $this->assertTrue(preg_match('/v1\.1/', $response['headers']), 'The Vary header was added.');
+
+    $response = $this->httpRequest('api/articles', \RestfulInterface::GET, NULL, array('X-API-Version' => 'v1.1'));
+    $this->assertTrue(preg_match('/X-API-Version/', $response['headers']), 'The Vary header was added.');
+
+    // Make sure that the version header is present in the response.
+    $this->assertTrue(preg_match('/v1\.1/', $response['headers']), 'The Vary header was added.');
```
amitaibu, line 286: `The Vary header was added.` => `The version header is present in the response.`

amitaibu:
Looks good, with just minor comment. Once green, can be merged.

tests/RestfulViewEntityTestCase.test
e0ipso, line 286: Actually this is intentional.

The `$response['headers']` only contains the contents of the header, so in this case if we want to detect `Vary: X-API-Version` we need to find `'X-API-Version'`. On the other hand if we want to detect `X-API-Version: v1.1` we'll look for `'v1.1'`.

PR #220, opened 2014-10-27
Avoid notices for missing plugins upon clear cache

amitaibu:
The bug can be seen upon clear cache on the watchdog. The class was loaded by Drupal's registry instead of CTools' one.


e0ipso:
Won't this fail to load the XML formatter? Because of:

``` php
class RestfulFormatterHalXml extends \RestfulFormatterHalJson implements \RestfulFormatterInterface {
```

e0ipso:
Nah! I just tested it and it works just fine.

Merging.

Thanks!

amitaibu:
I've checked and it seems there are other notices, related to CTools registry concerning `articles__1_2` and `articles__1_4` after clear cache. However they seem to work fine.

e0ipso:
Ugh! Sorry, I may have merged too quick then…

amitaibu:
No, the fix is valid. I think that it's anyway a CTools issue. Why would `articles__1_2` not find the file, as it's declared by CTools itself..

`Plugin articles__1_2 of plugin type restful:restful points to nonexistent file sites/all/modules/restful/modules/restful_example/plugins/restful/node/articles/1.2/RestfulExampleArticlesResource.class.php`

e0ipso:
@amitaibu I've seen this issue before with other projects. The CTools autoloader does not work well for `extend` clauses. Under those situations you have to do a `files[] = filename`.

PR #223, opened 2014-10-29
Corrected the values given by angularFileUpload to restful

BurningDog:
I couldn't get file uploads working in the restful_angular_example.module at `/restful-example/form`. The file was being sent to drupal, uploaded fine, and showed up in the file_managed table, but this wasn't being reflected in angular.

The issue is that the name of the array within `data` returned by angularFileUpload is slightly different from what restful-app.js was using. I've attached a screenshot of the data structure (I did a `console.log(data)` in the same place as the code changes in this pull request).

Basically, it should be `data.data.data` not `data.data.list`

This change then allowed angular to show that a file had been uploaded, and when I created an article node the image was there too (previously it wasn't).

screen shot 2014-10-29 at 6 17 53 pm

### DIFF
--- a/modules/restful_angular_example/components/restful-app/dist/restful-app.js
+++ b/modules/restful_angular_example/components/restful-app/dist/restful-app.js
@@ -155,8 +155,8 @@ angular.module('restfulApp')
       for (var i = 0; i < $files.length; i++) {
         var file = $files[i];
         FileUpload.upload(file).then(function(data) {
-          $scope.data.image = data.data.list[0].id;
-          $scope.serverSide.image = data.data.list[0];
+          $scope.data.image = data.data.data[0].id;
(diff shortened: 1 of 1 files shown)


amitaibu:
Thanks!

e0ipso:
@BurningDog thanks for contributing! It's exciting to see new faces in the issue queue :smiley:

BurningDog:
Yay! I watched @amitaibu's BoF session on headless drupal, and have started two headless drupal projects, one simple (an angular form posting to drupal, but the rest of the site is drupal) and one more complex (backend in drupal, used by editors, and frontend in angular, used by students and lecturers). Services module was going to be...painful. I like how you're approaching things with Restful, and the returned JSON is so clean! 

If I find any other issues and am able to help, I'll submit pull requests.

PR #226, opened 2014-10-30
Add allowed values for fields in OPTIONS

amitaibu:
This is a naive implementation that looks at the field's widget, and if it's not an autocomplete, it exposes the allowed values using `_options_get_options`. This will serves for both entity-reference module, and taxonomy term reference.

term reference:
postman-3

ER:
postman-2

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1184,8 +1184,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
 
     $discovery_info['form_element']['default_value'] = isset($instance_info['default_value']) ? $instance_info['default_value'] : NULL;
 
-    $this->getFormSchemaAllowedValues($field_info, $discovery_info);
-
+    $discovery_info['form_element']['allowed_values'] = $this->getFormSchemaAllowedValues($field_info);
 
     return array('discovery' => $discovery_info);
   }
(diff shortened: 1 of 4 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -1182,18 +1182,106 @@ protected function getFieldInfoAndFormSchema($field_info) {
...
+
+    if (!module_exists('options')) {
+      return;
+    }
+
+    $entity_type = $this->getEntityType();
+    $bundle = $this->getBundle();
+
+    $instance_info = field_info_instance($entity_type, $field_info['field_name'], $bundle);
+    $widget_types = array(
```
e0ipso: ~~Isn't there a way to get this list but a hard-coded list of values?~~ I guess not…

e0ipso:
Good use of the internals of the Field API! Although you're using some private API function I don't expect them to change.

Do we want tests for this?

amitaibu:
They are private functions but obviously they shouldn't ;) I also think they won't change.

Yes, tests are needed

‫ב-30 באוק 2014, בשעה 22:23, ‏Mateu Aguiló Bosch notifications@github.com כתב/ה:‬

plugins/restful/RestfulEntityBase.php
```diff
@@ -1182,18 +1182,73 @@ protected function getFieldInfoAndFormSchema($field_info) {
...
+      'taxonomy_term_reference',
+    );
+
+    $widget_types = array(
+      'taxonomy_autocomplete',
+      'entityreference_autocomplete',
+      'entityreference_autocomplete_tags',
+    );
+
+    if (in_array($field_info['type'], $field_types) && in_array($instance_info['widget']['type'], $widget_types)) {
```
amitaibu: @mateu-aguilo-bosch I was able to remove some of the hardcoding, by always using field APi to get the values -- unless if using a reference field with autocomplete.

amitaibu: still need to finish the tests.

amitaibu:

tests/RestfulDiscoveryTestCase.test
```diff
@@ -137,5 +137,46 @@ class RestfulDiscoveryTestCase extends RestfulCurlBaseTestCase {
     }
   }
 
+  /**
+   * Field discovery allowed values.
+   */
+  public function testFieldDiscoveryAllowedValues() {
```
e0ipso, line 143: We will also want to test:
- Taxonomy term fields. With hierarchical information.
- Option fields (one type is probably enough).

amitaibu, line 143: yeah, I was too lazy - and assumed that since options module took care of all cases it was enough with ER. but ok, I'll add it ;)

e0ipso, line 143: Fair enough. I don't want this to go stale :smile:

e0ipso:
This got meged in https://github.com/Gizra/restful/commit/250c1a38b9a1a421f186c0fd5eb57e798c17baf6.

Thanks!!

amitaibu:
Thanks :)

Issue #227, opened 2014-11-02
Allow defining a resource a property reference


amitaibu:
e.g. we should be able to define the `uid` as the user resource.

PR #228, opened 2014-11-02
Allow defining a resource for a property reference

amitaibu:
#227

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -361,26 +361,38 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
    *
    * @param $property
    *   The field name.
+   * @param \EntityMetadataWrapper $property_wrapper
+   *   The wrapped property.
+   *
    * @return string
    *   The target type of the referenced entity.
    *
    * @throws Exception
    *   Errors is the passed field name is invalid.
    */
-  protected function getTargetTypeFromEntityReference($property) {
-    if (!$field = field_info_field($property)) {
-      throw new Exception('Property is not a field.');
-    }
+  protected function getTargetTypeFromEntityReference($property, \EntityMetadataWrapper $property_wrapper) {
+    $params = array('@property' => $property);
 
-    if ($field['type'] == 'entityreference') {
-      return $field['settings']['target_type'];
-    }
-    elseif ($field['type'] == 'taxonomy_term_reference') {
-      return 'taxonomy_term';
-    }
+    if ($field = field_info_field($property)) {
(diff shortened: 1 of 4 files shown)


amitaibu:

plugins/restful/RestfulEntityBase.php
```diff
@@ -361,26 +361,38 @@ public function viewEntity($entity_id) {
...
+        return 'taxonomy_term';
+      }
 
-    throw new Exception('Property is not an entity reference field.');
+      throw new Exception(format_string('Field @property is not an entity reference or taxonomy reference field.', $params));
+    }
+    else {
+      // This is a property referencing another entity (e.g. the "uid" on the
+      // node object).
+      $info = $property_wrapper->info();
```
e0ipso, line 389: Nice!

plugins/restful/RestfulEntityBase.php
```diff
@@ -1135,6 +1146,12 @@ public function getPublicFields() {
           $info += $this->getFieldInfoAndFormSchema($field);
         }
       }
+      elseif ($info['resource']) {
+        // This is a property with a reference (e.g. the UID on a node entity).
+        $info += array('wrapper_method_on_entity' => TRUE);
+      }
+
+      $info += array('wrapper_method_on_entity' => FALSE);
```
e0ipso, line 1154: We can simplify this by doing:

``` php
<?php
$info += array('wrapper_method_on_entity' => !empty($info['resource']));
```

tests/RestfulReferenceTestCase.test
```diff
@@ -0,0 +1,48 @@
...
+      'uid' => $user1->uid,
+    );
+    $node1 = $this->drupalCreateNode($settings);
+    $node2 = $this->drupalCreateNode($settings);
+
+    $handler = restful_get_restful_handler('test_articles', 1, 2);
+    variable_set('restful_property_reference_simple', TRUE);
+
+    $result = $handler->get($node1->nid);
+    $this->assertEqual($result[0]['user']->uid, $user1->uid, 'Property is not defined as resource, thus the referenced entity appears as whole.');
```
e0ipso: `'Property is not defined as resource, thus the referenced entity appears as an ID.'`

e0ipso:
Great PR, just some minor comments before we hit the green button.

plugins/restful/RestfulEntityBase.php
amitaibu, line 1154: > We can simplify this by doing:

No, because we need `wrapper_method_on_entity == TRUE` only when the resource is defined as property. If it's a field, it should be `FALSE`

e0ipso, line 1154: :+1:

Issue #222, opened 2014-10-28
Provide an expireable token auth


amitaibu:
Currently token auth creates an entity that doesn't expire. I believe we should provide also an `expire` bundle + the cron job to expire it.

e0ipso:
Closed in favor of #230

Issue #221, opened 2014-10-28
Enforce X-Csrf-Token only for cookies and base auth


amitaibu:
Token auth for example doesn't need it.

amitaibu:
Looking at the code it seems we currently enforce it only for cookies. I'll check further.

amitaibu:
Works as expected.

PR #230, opened 2014-11-02
Add expire time for token auth

amitaibu:
#222 & #221

### DIFF
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -45,6 +45,12 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
 
     if (!empty($auth_token->expire) && $auth_token->expire < REQUEST_TIME) {
       // Token is expired.
+
+      if (variable_get('restful_token_auth_delete_expired_tokens', TRUE)) {
+        // Token has expired, so we can delete this token.
+        $auth_token->delete();
+      }
+
       return;
     }
 
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -43,7 +43,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
     $id = key($result['restful_token_auth']);
     $auth_token = entity_load_single('restful_token_auth', $id);
 
-    if (!empty($auth_token->expire) && $auth_token->expire > REQUEST_TIME) {
+    if (!empty($auth_token->expire) && $auth_token->expire < REQUEST_TIME) {
       // Token is expired.
       return;
     }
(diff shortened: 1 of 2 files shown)


modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
```diff
@@ -43,8 +43,14 @@ public function authenticate(array $request = array(), $method = \RestfulInterfa
     $id = key($result['restful_token_auth']);
     $auth_token = entity_load_single('restful_token_auth', $id);
 
-    if (!empty($auth_token->expire) && $auth_token->expire > REQUEST_TIME) {
+    if (!empty($auth_token->expire) && $auth_token->expire < REQUEST_TIME) {
       // Token is expired.
+
+      if (variable_get('restful_token_auth_delete_expired_tokens', TRUE)) {
```
e0ipso, line 49: Do we want to leave this for cron? I am cool with either way, I don't think there will be a performance impact.

e0ipso:
The principles behind this are very similar to the ones behind OAuth2 Bearer Token. Maybe we want to adapt to that.

http://tools.ietf.org/html/rfc6750

modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
amitaibu, line 49: Indeed, Cron is already implemented, I thought it can be a nice house-keeping with a small impact.

amitaibu:
> http://tools.ietf.org/html/rfc6750

I think that with 3 small kids at home I have an exempt from reading OAuth 2.0 Authorization Framework docs ;)

e0ipso:
@amitaibu you're exempt just for being a human being.

I was thinking on adaption our output to match the only thing that I read, [the example](http://tools.ietf.org/html/rfc6750#page-10).

``` javascript
     {
       "access_token":"mF_9.B5f-4.1JqM",
       "token_type":"Bearer",
       "expires_in":3600,
       "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA"
     }
```

amitaibu:
@mateu-aguilo-bosch care to take this PR from here? :)

e0ipso:
@amitaibu consider it done.

Also, I'm having a travel ordeal on Saturday and I'd like to have _restful_ stuff to do. Can we prioritize?

I feel we're _this close_ to the release candidate.

e0ipso:
**NOTES:**
There are 3 different ways we will want to support for clients to send the token:
1. `Authentication: Bearer <token>`. MANDATORY support to this method.
2. Form-Encoded Body Parameter: `Content-Type: application/x-www-form-urlencoded` header + `access_token=mF_9.B5f-4.1JqM` in the body of a POST request. This is **not suitable** for GET requests. Supporting this method is OPTIONAL.
3. URI Query Parameter: `GET /resource?access_token=mF_9.B5f-4.1JqM HTTP/1.1`. Client and server **must** populate the `Cache-Control` header. The use of this is discouraged, unless it's impossible to use 1. Supporting this method is OPTIONAL.

amitaibu:
@mateu-aguilo-bosch maybe we should get this in, and deal with improvements in follow up?

e0ipso:
I'm working on this right now.

e0ipso:
Closed in favor of #264.

Issue #233, opened 2014-11-04
New entity reference breaks taxonomy term handling, I think. 


mirzu:
500 Error: Unknown data property field_blog_categories_term_tree. 

For a public fields declaration: 

```
    $public_fields['categories'] = array(
      'property' => 'field_blog_categories_term_tree',
      'resource' => array(
        'categories' => 'categories',
      ),
    ); 
```

I think it's because we are sending $item_wrapper which is a wrapper for each taxonomy term to getValueFromResource() but treating like it's the outer entity wrapper when calling getTargetTypeFromEntityReference  $wrapper->{$property}

amitaibu:
closed in favor of #234

PR #234, opened 2014-11-04
Fix 'resource' for reference fields to other entities

amitaibu:
#233

Next is adding tests to prevent regression.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -357,20 +357,19 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
   }
 
   /**
-   * Get the "target_type" property from an entity reference field.
+   * Get the "target_type" property from an field or property reference.
    *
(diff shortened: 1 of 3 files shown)


amitaibu:
It seems we are [already](https://github.com/Gizra/restful/blob/7.x-1.x/tests/RestfulViewEntityTestCase.test#L104-L119) testing the reference fields, and somehow it passes - so I'll need to fix also the wrong(?) test.

amitaibu:
Ok, the error seems to happen when the reference is to another resource. Currently the test has a reference field that references same resource as the parent.

amitaibu:
@mateu-aguilo-bosch @mirzu ready for review.

mirzu:
I did a functional test and it works. I don't have anything to add in regards to the code. :+1:

e0ipso:
Merged in https://github.com/Gizra/restful/commit/627cf6f834478f5fc564096662e72c457cdc2f23!

Thanks!

Issue #186, opened 2014-10-13
Make our HAL+JSON more HAL compliant


amitaibu:
http://stateless.co/hal_specification.html

Seriously @mirzu, this issue has your name all over it! :)

amitaibu:
@mateu-aguilo-bosch @mirzu do we need to start by removing the "data" wrapper from the HAL-JSON formatter plugin?

![local_d7_dev_api_v1_articles_](https://cloud.githubusercontent.com/assets/125707/4669346/6d32119c-556d-11e4-94a7-5ad3e5b84bf7.jpg)

(See the ng-admin [issue](https://github.com/marmelab/ng-admin/issues/86) that made me think of it)

amitaibu:
Ignore above comment, seems wrapping with `data` is a common practice.

mirzu:
I'm curious to see other examples where data is used to list embedded resources. 

From the specification it looks like what we are wrapping in data belongs in the reserved property "_embedded" The reason for using _embedded is that API client libraries will treat _embedded as a list of resources that can potentially be followed, rendered etc,  while they will treat data a property.

 The reserved "_embedded" property is OPTIONAL

   It is an object whose property names are link relation types (as
   defined by [RFC5988]) and values are either a Resource Object or an
   array of Resource Objects.

   Embedded Resources MAY be a full, partial, or inconsistent version of
   the representation served from the target URI.

mirzu:
More thoughts:
The way we are doing it isn't invalid, but using embedded for lists of resources would be better.

Wrapping the properties in a data object isn't bad, but I'm curious to know what we gain by doing so. 

Wrapping _embdedded in data along with properties would be bad.

amitaibu:
> I'm curious to see other examples where data is used to list embedded resources.

@mirzu , the examples I've seen were not HAL specific. But in my case, the fact that we wrap our data with `data` was very handy, as I just had to point ng-admin to the correct key, rather then ignoring other root level keys.

mirzu:
One reason to make the API follow HAL conventions more closely is this little gem:

It's a HAL specific ng-admin type thing that is 100% drop in 0 config if you implement enough of the features. 
Example that is on it's way: http://mirzu.fourword.webchefs.org/hal-browser/browser.html#/api/v1/blogposts/142
Source: https://github.com/mikekelly/hal-browser
Full Example: http://haltalk.herokuapp.com/explorer/browser.html#/

amitaibu:
[...]

PR #238, opened 2014-11-05
Make HAL formatter strict.

amitaibu:
#186

Just playing around with some code...

### DIFF
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -218,13 +218,18 @@ class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulF
    *   Output array to be modified.
    * @param array $row
    *   The row being processed.
-   * @param $public_field
+   * @param array $public_field
+   *   The public field configuration array.
    * @param $public_field_name
    *   The name of the public field.
    */
   protected function moveReferencesToEmbeds(array &$output, array &$row, $public_field, $public_field_name) {
     $value_metadata = $this->handler->getValueMetadata($row['id'], $public_field_name);
     foreach ($row[$public_field_name] as $index => $resource_row) {
+      if (empty($value_metadata[$index])) {
+        // No metadata.
+        continue;
+      }
       $metadata = $value_metadata[$index];
 
       // If there is no resource name in the metadata for this particular value,
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -219,7 +219,7 @@ class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulF
    * @param array $row
    *   The row being processed.
    * @param array $public_field
-   *   The public field ocnfiguration array.
+   *   The public field configuration array.
    * @param $public_field_name
    *   The name of the public field.
    */
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -218,13 +218,18 @@ class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulF
    *   Output array to be modified.
    * @param array $row
    *   The row being processed.
-   * @param $public_field
+   * @param array $public_field
+   *   The public field ocnfiguration array.
    * @param $public_field_name
    *   The name of the public field.
    */
   protected function moveReferencesToEmbeds(array &$output, array &$row, $public_field, $public_field_name) {
     $value_metadata = $this->handler->getValueMetadata($row['id'], $public_field_name);
     foreach ($row[$public_field_name] as $index => $resource_row) {
+      if (empty($value_metadata[$index])) {
+        // No metadata.
+        continue;
+      }
       $metadata = $value_metadata[$index];
 
       // If there is no resource name in the metadata for this particular value,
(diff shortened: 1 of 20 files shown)


amitaibu:
local_d7_dev_api_v2 0_articles_filter_id__68

amitaibu:
And now with `_embedded`:

local_d7_dev_api_v1 5_articles_filter_id__68

amitaibu:
Seems that even the HAL browser already likes it :)

the_hal_browser

the_hal_browser

amitaibu:
Not ready for review yet, but @mateu-aguilo-bosch @mirzu -  are you ok with the direction?

plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
```diff
@@ -81,10 +86,113 @@ protected function addHateoas(array &$data) {
...
+
+        foreach ($public_field['resource'] as $resource_bundle => $resource) {
+          if ($resource_bundle == $bundle) {
+            $resource_name = $resource['name'];
+            continue 2;
+          }
+        }
+      }
+
+      $curies_resource = variable_get('restful_hal_curies_name', 'hal') . ':' . $resource_name;
```
mirzu: This makes curies required, and they should always be optional. so need to check if it's set maybe?

amitaibu: > so need to check if it's set maybe?

Question is _where_ do we set this? Current implementation moves all the responsibility to the `formatter` handler, so the `restful` handler has no knowledge about it.

mirzu: no just check that the variable is set. 

$curies_name = variable_get('restful_hal_curies_name', 'hal');
$resource_name = ($curies_name) ? $curies_name . ':' . $resource_name : $resource_name; 

do the same check for the curies link. . .

e0ipso: This probably deserves its own helper function.

e0ipso:
I like the general direction of this. Maybe move the hal prefix to the Formatter Plug-in definition?

e0ipso:
I added some minor changes, let's see if you like the idea. Now curies are optional.

plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
```diff
@@ -69,22 +63,170 @@ protected function addHateoas(array &$data) {
...
+  /**
+   * Prefix a property name with the curie, if present.
+   *
+   * @param string $property_name
+   *   The input string.
+   *
+   * @return string
+   *   The property name prefixed with the curie.
+   */
+  protected function withCurie($property_name) {
```
amitaibu, line 197: maybe better `getResourceNameWithCurie()`?

Also why is the signature `$property_name` shouldn't it be `$resource_name`?

plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
```diff
@@ -69,22 +63,170 @@ protected function addHateoas(array &$data) {
...
+    if ($curie = $this->getCurie()) {
+      return $property_name ? $curie['name'] . ':' . $property_name : $curie['name'];
```
amitaibu, line 199: When will `$property_name` be empty?

plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
```diff
@@ -69,22 +63,170 @@ protected function addHateoas(array &$data) {
...
+      );
     }
+
+    if (!$curie = $this->getCurie()) {
+      return;
+    }
+
+    $data['_links']['curies'] = array(
+      'name' => $curie['name'],
+      'href' => $curie['href'] ? $curie['href'] : url('docs/rels', array('absolute' => TRUE)) . '/{rel}',
```
amitaibu, line 116: I think the default one shouldn't be hardcoded -- it should use `variable_get('restful_hal_json_curie_url', FALSE)`

amitaibu:
I've added a a default `json` formatter - which is what we called `hal_json` until now.

e0ipso:
> I've added a a default json formatter - which is what we called hal_json until now.

yikes! I did the same thing in https://github.com/Gizra/restful/pull/244 :-S

amitaibu:
@mateu-aguilo-bosch your is with testing, so we can probably use yours.

amitaibu:
@mateu-aguilo-bosch you want to merge your changes instead of mine?

plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
e0ipso, line 116: Then it makes more sense to me that we only add the href when there is a value for it, instead of providing a variable with the defaults.

plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
e0ipso, line 199: This is as a helper way to get the curie name.

``` php
print withCurie();
```

e0ipso:
@amitaibu we can merge mine, but we'll want to wait until #186 lands first… or cherry pick the relevant commits.

plugins/restful/RestfulEntityBase.php
```diff
@@ -403,11 +403,16 @@ protected function getTargetTypeFromEntityReference(\EntityMetadataWrapper $wrap
    *   The property name (i.e. the field name).
    * @param array $resource
    *   Array with resource names, keyed by the bundle name.
+   * @param string $public_field_name
+   *   Field name in the output. This is used to store additional metadata
+   *   useful for the formatter.
+   * @param int $host_id
```
amitaibu, line 409: maybe better `resource_id`?

e0ipso:
Closed in favor of #256.

PR #242, opened 2014-11-09
Access exception list

e0ipso:


plugins/restful/RestfulDataProviderCToolsPlugins.php
```diff
@@ -202,7 +202,14 @@ public function index() {
     $return = array();
 
     foreach (array_keys($this->getPluginsSortedAndFiltered()) as $plugin_name) {
-      $return[] = $this->view($plugin_name);
+      try {
+        $return[] = $this->view($plugin_name);
+      }
+      catch (\RestfulForbiddenException $e) {
```
amitaibu, line 208: what if the request is `api/articles?id=1,2,3` -- i would expect an exception to be thrown, no?

plugins/restful/RestfulBase.php
```diff
@@ -68,6 +68,22 @@
   protected $range = 50;
 
   /**
+   * Supressed rows. Used to correct the count items.
+   *
+   * @var int
+   */
+  protected $supressedRows = 0;
```
amitaibu, line 75: I'm not sure about this. Let's say you have 2 pages of results with a total of 100 items. The first 49 items are un-accessible by the user via `hook_node_access()`

Then on page 1 she will see only one items with a count of 51
On page 2 she will see 50 items with a count of 100.

I would day it's better to keep always the 100 count, eventhough it hints the user there are un-accessible items - but I think it's an edge case.

amitaibu:
@mateu-aguilo-bosch I'm taking a slightly different approach in #253

PR #244, opened 2014-11-09
Simple JSON formatter

e0ipso:
This requires #238.

This will introduces explicit formatter testing.


amitaibu:
Lets consolidate the work done here with #238?

amitaibu:
I've merged this with @238 -- so we'll have one single place to work on this.

Issue #245, opened 2014-11-09
Custom headers expect prefix HTTP_X_


ceoaliongroo:
- When a request is send it with a custom header 

![screen shot 2014-11-09 at 9 59 53 am](https://cloud.githubusercontent.com/assets/2622152/4966828/2632bc2e-67e7-11e4-9835-71e55d4e2335.png)
- The restful token auth module look for the header with a prefix HTTP_X_

https://github.com/Gizra/restful/blob/7.x-1.x/modules/restful_token_auth/restful_token_auth.module#L26
- But the server set the custom header without _X_ 

![screen shot 2014-11-09 at 9 53 24 am](https://cloud.githubusercontent.com/assets/2622152/4966833/b4f873cc-67e7-11e4-9b19-aac3ac530509.png)

PR #246, opened 2014-11-09
fix: change prefix as is defined by the server.

ceoaliongroo:
#245


amitaibu:
Seems we will do some changes to the token in #230, but for now we can get this one in.

PR #251, opened 2014-11-14
Added methods to define default sort

raphaeltraviss:
My first ever pull request!

OK, so with this pull, we can define default sorts and field limiters with defaultSortInfo() and defaultFieldInfo().  If the client passes a query string in the URL, they are completely overridden.

One thing I tried to do was to add the processing methods to check for invalid stuff, and allow you to supply the defaults in the same way as you specify them in the URL; e.g. instead of DESC/ASC you can simply supply 'fieldname' or '-fieldname'.

Field filters (as opposed to field limiters) seem to be more complicated, so I've leaving them for now.

So... how does it look?


README.md
```diff
@@ -202,6 +244,47 @@ array(
...
+defaultSortInfo() in your class definition.
+
+This method should return an indexed array, with each element being a property
+name that you want to use for sorting.  Each sort field you declare should match
+a field defined in publicFieldsInfo().
+
+Prefixing the sort field names with a minus (-) sign will apply a descending sort,
+instead of the default ascending.
+
+This default sort will be nullified if the request URL contains a sort query.
```
e0ipso: _This is only the default sort, any explicit sort declared in the URL will take precedence_

README.md
```diff
@@ -143,6 +143,48 @@ array(
...
 
+You can also define default field limiters in your plugin, by overriding
+defaultFieldInfo() in your class definition.
+
+This method should return an indexed array, with each element being a property
+name that you want to restrict the response to.  Each field limiter you declare
+should match a field defined in publicFieldsInfo().
+
+These field limiters will be nullified if the request URL contains a sort query.
+
```
e0ipso: I can only see a use case for this and I'm not sure it is worth adding another layer of complexity.

Use case I see is:
1. You declare a resource with lots of fields.
2. You limit the fields to only show a bunch of them.
3. Whenever you need any of the fields not being shown, you use the URL options to bring them.

Otherwise you can just remove the fields from `publicFieldsInfo`.

Not sure if this edge case is worth the trouble.

README.md
```diff
@@ -202,6 +244,47 @@ array(
...
+    );
+
+    $public_fields['tid'] = array(
+      'property' => 'tid'
+    );
+
+    return $public_fields;
+  }
+
+  public function defaultSortInfo() {
```
e0ipso, line 218: I like this idea of allowing an optional default sort.

2 minor things:
- [ ] Make sure that the docs say that this is optional.
- [ ] Make sure to use a different example, since the default sort is based on entity id (even though you are using the reverse order).

README.md
e0ipso: @amitaibu thoughts? I'm leaning towards leaving this out, but I can be convinced otherwise. :arrow_up:

e0ipso:
w00t! Thanks a lot for contributing @rfx1007!!!

README.md
```diff
@@ -143,6 +143,48 @@ array(
...
+should match a field defined in publicFieldsInfo().
+
+These field limiters will be nullified if the request URL contains a sort query.
+
+```php
+class MyPlugin extends \RestfulEntityBaseTaxonomyTerm {
+
+  // ...
+
+  public function publicFieldsInfo() {
```
amitaibu: this can be removed as the `\RestfulEntityBaseTaxonomyTerm` already provides ID and label.

README.md
```diff
@@ -143,6 +143,48 @@ array(
...
+    );
+
+    $public_fields['tid'] = array(
+      'property' => 'tid'
+    );
+
+    return $public_fields;
+  }
+
+  public function defaultSortInfo() {
```
amitaibu: add PHPdocs even in the docs.

README.md
```diff
@@ -143,6 +143,48 @@ array(
...
+    return array('-tid');
```
amitaibu: this should be 

``` php
return array(
  'id' => 'DESC',
);
```

README.md
```diff
@@ -202,6 +244,47 @@ array(
...
+name that you want to use for sorting.  Each sort field you declare should match
+a field defined in publicFieldsInfo().
+
+Prefixing the sort field names with a minus (-) sign will apply a descending sort,
+instead of the default ascending.
+
+This default sort will be nullified if the request URL contains a sort query.
+
+```php
+class MyPlugin extends \RestfulEntityBaseTaxonomyTerm {
```
amitaibu, line 214: seems the docs are repeating -- it's enough to describe once `defaultSortInfo`

plugins/restful/RestfulEntityBase.php
```diff
@@ -281,6 +281,10 @@ public function viewEntity($entity_id) {
 
     $limit_fields = !empty($request['fields']) ? explode(',', $request['fields']) : array();
 
+    if (empty($limit_fields)) {
```
amitaibu: remove

plugins/restful/RestfulEntityBase.php
```diff
@@ -1175,6 +1179,94 @@ public function getPublicFields() {
...
+
+  /**
+   * Process default sort field definitions to get the sorting options.
+   *
+   * @return array
+   *   With the different sorting options.
+   *
+   * @throws Exception
+   */
+  protected function processDefaultSorts() {
```
amitaibu: I don't think we need it, we could pas the default sort to our regular sort -- and if there's an error there it will throw the exception.

raphaeltraviss: These methods handle invalid values that could be supplied in the defaultXInfo methods, which I've thought of a couple ways to deal with:
1. provide simplified copies of the parseRequest methods as processX methods.  Then, the data provider classes would need to call processX methods in addition to parseRequestForX methods.
2. change the name of parseRequestForX methods and modify them to merge values from both the request and the defaultXInfo methods.  The data provider classes would need to call the method by its new name.
3. Don't handle invalid values supplied via the plugin definition at all; let them hit the sorting/limiting logic and throw an exception there.  The data provider classes would call the defaultXInfo() methods directly.

I was leaning towards options 1 or 3, but I decided to try option 1, because I felt that having the processX methods was more consistent with the existing pattern of calling parseRequestForX in the data provider, and then having parseRequest sanitize the options.

On the other hand, I can see where it's not technically necessary to have this consistency, since the plugin-supplied options are being supplied by the plugin developer him/herself, and not by an API consumer.

So, should I go ahead with option 3?

plugins/restful/RestfulEntityBase.php
raphaeltraviss: The purpose of this empty() check is to add in the default field limiters if there are none supplied via the URL.  Are you saying that we should remove the empty() check, and have the default field limiters added all the time, even if some are supplied via the URL?  This would make it impossible to fully override the default field limiters... am I misinterpreting your 'remove' comment?

plugins/restful/RestfulEntityBase.php
amitaibu: > So, should I go ahead with option 3?

Yes, simply because it's the simplest solution.

plugins/restful/RestfulEntityBase.php
amitaibu: Yes, sorry - my intention was that that the entire field limiters functionality should be removed. If you wish to provide those limiters you may do it via `access callbacks` on each public field.

raphaeltraviss: Ok, so the intent is to remove all field limiter functionality altogether, including the ?fields capability from the URL, and therefore, we should remove all field limiter functionality from this pull request?

amitaibu: Indeed, the PR should be only about defaultSort (I'm sorry if I misguided you in my previous comments about those limiters),

raphaeltraviss:
Ok, I pared it down a bit.

Also, I moved defaultSortInfo() to RestfulBase because it should work for both EFQ and DbQuery plugins; it's working as expected with my custom query plugin.

README.md
```diff
@@ -202,6 +202,47 @@ array(
 );
 ```
 
+You can also define default sort fields in your plugin, by overriding
+defaultSortInfo() in your class definition.
```
amitaibu: wrap `defaultSortInfo()` and `publicFieldsInfo()`  in <code>``</code>

README.md
```diff
@@ -202,6 +202,47 @@ array(
...
+
+This method should return an associative array, with each element having a key
+that matches a field from publicFieldsInfo(), and a value of either 'ASC' or 'DESC'.
+
+This default sort will be nullified if the request URL contains a sort query.
```
amitaibu: will be nullified => will be ignored

README.md
```diff
@@ -202,6 +202,47 @@ array(
...
+
+```php
+class MyPlugin extends \RestfulEntityBaseTaxonomyTerm {
+
+  /**
+   * Overrides \RestfulEntityBase::publicFieldsInfo().
+   */
+  public function publicFieldsInfo() {
```
amitaibu: I don't think `publicFieldsInfo()` is needed here, enough to show defaultSortInfo()

README.md
```diff
@@ -202,6 +202,47 @@ array(
...
+
+    return $public_fields;
+  }
+
+  /**
+   * Overrides \RestfulEntityBase::defaultSortInfo().
+   */
+  public function defaultSortInfo() {
+    // Sort by 'field_myfield' in descending order.
+    return array('myfield' => 'DESC');
```
amitaibu: Maybe keep it basic?
`return array('id' => 'DESC');`

plugins/restful/RestfulBase.php
```diff
@@ -882,6 +882,17 @@ public function getPublicFields() {
   }
 
   /**
+   * Defines default sort fields if none are provided via the request URL.
+   *
+   * @return array of key/value pairs:
```
amitaibu:
``` php
* @return array
*    Array keyed by the public field name, and the order ('ASC' or 'DESC') as value.
```

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -121,8 +121,14 @@ public function getQueryForList() {
    */
   protected function queryForListSort(\SelectQuery $query) {
     $public_fields = $this->getPublicFields();
+
     // Get the sorting options from the request object.
     $sorts = $this->parseRequestForListSort();
+
+    if (empty($sorts)) {
+      $sorts = $this->defaultSortInfo();
```
amitaibu: `$sorts = $sorts ? $sorts : $this->defaultSortInfo();`

amitaibu:
PR is getting closer :)
Now we also need to make sure it's tested in `\RestfulListTestCase::testSort()`

You can add for example to `RestfulTestArticlesResource__1_1` a `defaultSortInfo()` and then test it:

``` php
$handler = restful_get_restful_handler('test_articles', 1, 1);
$result = $handler->get();

// Assert IDs are sorted descending.
```

plugins/restful/RestfulBase.php
```diff
@@ -882,6 +882,17 @@ public function getPublicFields() {
...
+   *   each key must match a field returned by publicFieldsInfo(), and each
+   *   value must be one of either ASC or DESC.
+   */
+  public function defaultSortInfo() {
+    return array();
```
amitaibu: I think this should actually be populated with something like:

``` php
  // Some endpoints like 'token_auth' don't have an id public field. In that
  // case, skip the default sorting.
  return !empty($public_fields['id']) ? array('id' => 'ASC') : array();
```

and for DB: `return array($this->getIdColumn() => 'ASC')`

So we can avoid that logic in `queryForListSort`

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -95,6 +95,11 @@ public function getQueryForList() {
   protected function queryForListSort(\EntityFieldQuery $query) {
     $public_fields = $this->getPublicFields();
     $sorts = $this->parseRequestForListSort();
+
+    if (empty($sorts)) {
+      $sorts = $this->defaultSortInfo();
```
e0ipso: `$sorts = $sorts ? $sorts : $this->defaultSortInfo();`

README.md
e0ipso: I actually like the idea of keeping `myfield` following the declared field above.

e0ipso:
@rfx1007 please, let me know if you need more clarification. I understand that you are not as familiar with the codebase as we are :-)

raphaeltraviss:
Moved the logic back into the data providers, since they have different defaults.  Also, added a test for defaultSortInfo() and a test for overriding it.

amitaibu:
@rfx1007 you are getting very close, just need to get those tests back to green! Don't be discouraged - everybody hates (fixing the) tests ;)

tests/RestfulRenderCacheTestCase.test
```diff
@@ -34,6 +34,14 @@ class RestfulRenderCacheTestCase extends DrupalWebTestCase {
       $nodes[$node->nid] = $node;
     }
 
+    // Fake defaultSortInfo() on resource.
+    // Ignore secondary id sort, since the chances of having duplicate
+    // randomly-generated node titles is astronomical.
+    function labelSort($a, $b) {
+      return strcasecmp($a->title, $b->title);
+    }
+    usort($nodes, 'labelSort');
```
e0ipso: I don't understand why we need this. Was this causing test failures? Is so, why?

raphaeltraviss: The failures were caused by the random title on the test nodes.  Since we are sorting by label, this was causing the nodes to be sorted in a different order than they were originally created in.

Consider:

``` php
    $node = reset($nodes);
    $cid = 'v1.1' . '::uu' . $account->uid . '::pa' . 'et:node::ei:' . $node->nid;
    $record = $cache->get($cid);
    $this->assertEqual($pre_cache_results[0], $record->data, 'Data in cache bins is correct.');
```

$pre_cache_results is being built from $handler->get(), which is now being sorted by label, whereas $record->data is being built from the nid of the first element of $nodes, which is ordered by id, so the test will fail 66% of the time, depending on what the random titles generate.

By applying the sort function to $nodes, reset($nodes) and the $pre_cache_results[0] now point to the same node, so the test is now valid.

amitaibu: I agree with @mateu-aguilo-bosch - lets avoid complex non-standard stuff in the tests. If the difficulty is the random titles - then go ahead and use hardcoded ones.

raphaeltraviss: hard-coded the titles with $index so they are sorted in the expected order.

tests/RestfulRenderCacheTestCase.test
```diff
@@ -29,7 +29,7 @@ class RestfulRenderCacheTestCase extends DrupalWebTestCase {
 
     $num_articles = 3;
     for ($index = 0; $index < $num_articles; $index++) {
-      $settings['title'] = $this->randomName();
+      $settings['title'] = 'Node title '.$index;
```
amitaibu: `'Node title '.$index` => `'Node title ' . $index`

tests/RestfulRenderCacheTestCase.test
```diff
@@ -79,7 +80,9 @@ class RestfulRenderCacheTestCase extends DrupalWebTestCase {
     $record = $cache->get($cid);
     $this->assertFalse($record, 'Cache bin is emptied after updating a user.');
 
-    // Test cache segmentation with request params.
+    $node = reset($nodes);
```
amitaibu: can you add a comment about this part?

raphaeltraviss: I just set this back to the way it was; now that the hard-coded titles sort in the expected order, we don't need to reset $node and $cid after appending to one.

amitaibu:
Well done, merged!

raphaeltraviss:
:trophy: finally legit!

amitaibu:
@rfx1007 I don't know what your twitter name is, but you should defiantly brag about this ;)

e0ipso:
YES!

Issue #239, opened 2014-11-07
If access check fails for one resource in a list the whole list is permission denied.


mirzu:
If one user on a site is blocked the whole list of users won't show up.

e0ipso:
My bet is that we should do like we do with properties with access denied, remove them.

amitaibu:
Yes, we should ignore them, unless specifically asked for that idea. 

So if `/api/articles` -- just skip the unaccessible entity (btw, this means that the `count` property would be wrong).

But if `/api/articles?id=1,2,3` un ID 3 is un-acecssible, we should throw an error.

amitaibu:
@mirzu care to create a PR :)

e0ipso:
Closed in favor of https://github.com/Gizra/restful/pull/242

PR #253, opened 2014-11-17
Don't throw exception in list of entities with un-accessible items

amitaibu:
#239

If no IDs were requested, we should not throw an exception in case an
entity is un-accessible by the user.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -130,7 +130,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
 
     // If no IDs were requested, we should not throw an exception in case an
     // entity is un-accessible by the user.
-    $error_on_no_access = !empty($request['id']) || !empty($request['filter']['id']);
+    $error_on_no_access = !$this->isListRequest() || !empty($request['filter']['id']);
 
     foreach ($ids as $id) {
       if ($row = $this->viewEntity($id, $error_on_no_access)) {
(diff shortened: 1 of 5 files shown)


amitaibu:
Next is actually checking the code I just wrote, and writing tests to it :)

amitaibu:
Update: Error seems related to entity-validator module. Checking

e0ipso:
@amitaibu cool! Let me know.

plugins/restful/RestfulEntityBase.php
```diff
@@ -128,8 +128,14 @@ public function getList() {
 
     $return = array();
 
+    // If no IDs were requested, we should not throw an exception in case an
+    // entity is un-accessible by the user.
+    $error_on_no_access = !empty($request['id']) || !empty($request['filter']['id']);
```
e0ipso: We should be using the `isListRequest` helper function.

e0ipso:
One minor comment and we can hit the green.

e0ipso:
Is this still WIP? The title says so.

amitaibu:
Yeah, still WIP -- I want to add tests.

e0ipso:
That's a good idea :-)

amitaibu:

amitaibu:
Actually I got the filter part wrong, we just need to check if no specific ID was requested. So no review yet :)

amitaibu:
Ready for review.

e0ipso:
Merged!

Thanks Amitai!

PR #257, opened 2014-11-18
Try to get HTTP headers directly from Apache for a token with underscore

amitaibu:
I hate computers!


amitaibu:

amitaibu:
btw, here's some info about this -- http://stackoverflow.com/questions/18185366/header-names-with-underscores-ignored-in-php-5-5-1-apache-2-4-6

amitaibu:
Actually, I think it would be better to have a helper function that will do this check for any variable we need from the request (e.g. the major/ minor version).

e0ipso:
~~@amitaibu you're absolutely right. We should address this here in this PR.~~~

See https://github.com/Gizra/restful/pull/257/files#r20633137

e0ipso:
From [Overview of new features in Apache HTTP Server 2.4](http://httpd.apache.org/docs/trunk/new_features_2_4.html):

_Translation of headers to environment variables is more strict than before to mitigate some possible cross-site-scripting attacks via header injection. Headers containing invalid characters (including underscores) are now silently dropped. Environment Variables in Apache has some pointers on how to work around broken legacy clients which require such headers. (This affects all modules which use these environment variables.)_

modules/restful_token_auth/restful_token_auth.module
```diff
@@ -23,10 +23,19 @@ function restful_token_auth_restful_parse_request_alter(&$request) {
...
 
-  $capital_name = strtoupper('HTTP_' . $param_name);
+  $capital_name = 'HTTP_' . strtoupper($param_name);
+
+  $value = !empty($_SERVER[$capital_name]) ? $_SERVER[$capital_name] : NULL;
+
+  if (!$value && function_exists('apache_request_headers')) {
+    // Apache may be strict with variables with underscore, so we get the
+    // headers directly from Apache.
+    $headers = apache_request_headers();
```
e0ipso: I think that we should just sanitize the token values on generation to avoid the header from being dropped. This would save us from this apache specific hack.

amitaibu:
> I think that we should just sanitize the token values on generation to avoid the header from being dropped

Can you explain what you mean?

e0ipso:
Basically Apache seems to be dropping headers with invalid characters in them, that means that if you set a header with curl to be `a_b` you won't have access to it in `$_SERVER`. I think that is fine, and even encourageable.

What we need to do is provide tokens that don't contain any invalid character (including underscores) and forget about the matter. If any client is sending header values with invalid characters, then Apache does not support them.

Does that make sense?

amitaibu:
I would prefer to "fix" it in RESTful, rather then developers not understanding why "their_token" doesn't work. Also I think we already have other variables (e.g. `api_version`) that need to be taken care of.

amitaibu:
> What we need to do is provide tokens that don't contain any invalid character (including underscores)

Or maybe I'm getting this wrong. Do you know why is a variable with underscores considered invalid? Is it Apache specific considering this invalid?

e0ipso:
See https://github.com/Gizra/restful/pull/261 to see my point and the code I think it can fix it.

e0ipso:
Should we also convert all these?

www 2014-11-23 09-30-46

includes/RestfulManager.php
```diff
@@ -291,4 +291,32 @@ public static function invalidateEntityCache($cid) {
...
+   * @param string $key
+   *   The key to use.
+   * @param string $default_value
+   *   The default value to return if no value exists. Defaults to NULL.
+   *
+   * @return string
+   *   The value in the HTTP header if exists, other the value of the given
+   *   "default value".
+   */
+  public static function getHttpHeader($key, $default_value = NULL) {
```
e0ipso: `getRequestHttpHeader`

amitaibu:
> Should we also convert all these?

I'll execute tests on a machine with php > 5.5.1

e0ipso:
I meant if we want to be consistent and always get request headers through the helper function instead of having two different ways of doing it.

amitaibu:
> I meant if we want to be consistent and always get request headers

Yes, we want to be consistent. We can actually also test php 5.5.x on travis, although it will probably double the testing time.

e0ipso:
Gotcha!

amitaibu:
Travis is awesome - php5.5 are running in parallel :)
https://travis-ci.org/Gizra/restful/builds/41935651

e0ipso:
That is so cool!

e0ipso:
If tests come back green this should be ready to merge.

amitaibu:
hmm, the failing tests are passing locally for me. @mateu-aguilo-bosch can you try to reproduce to failing tests please?

e0ipso:
Hmm, this is funny.

Executing the tests with the core's script yields no errors.

``` shell
php /var/www/docroot/scripts/run-tests.sh --color --verbose --url http://rest-performance.local --class RestfulCsrfTokenTestCase
```

Executing the tests with drush yields errors:

``` shell
drush test-run RestfulCsrfTokenTestCase
```

@amitaibu I hope this helps.

amitaibu:
Thanks, I think I was able to reproduce...

amitaibu:
@mateu-aguilo-bosch green...

plugins/restful/RestfulBase.php
```diff
@@ -1333,10 +1333,11 @@ public static function getVersionFromRequest($path = NULL) {
       return $version;
     }
     // If there is no version in the URL check the header.
-    if (!empty($_SERVER['HTTP_X_API_VERSION'])) {
-      $version =  static::parseVersionString($_SERVER['HTTP_X_API_VERSION'], $resource_name);
+    if ($api_version = \RestfulManager::getRequestHttpHeader('X_API_VERSION')) {
```
e0ipso: This should be `'X-API-Version'`.

restful.api.php
```diff
@@ -19,7 +19,7 @@
  */
 function hook_restful_parse_request_alter(&$request) {
   $request['__application'] += array(
-    'some_header' => !empty($_SERVER['X_HTTP_SOME_HEADER']) ? $_SERVER['X_HTTP_SOME_HEADER'] : NULL,
+    'some_header' => \RestfulManager::getRequestHttpHeader('X_HTTP_SOME_HEADER'),
```
e0ipso: This was probably intended to be `'X-Some-Header'`.

amitaibu:
@mateu-aguilo-bosch comments are fixed.

e0ipso:
Thanks!

Merged.

PR #258, opened 2014-11-18
Empty value should not create a new entity from a reference field.

HelenaEksler:
Sometimes a client may send a NULL value, but it is somehow translated to an empty string.
Following check makes sure with are returning a NULL value in that case.


e0ipso:
What if you send the payload without the reference at all (no need to make it NULL). Would that work?

HelenaEksler:
Yes, you are right. If I don't send this reference at all - it's working.

PR #215, opened 2014-11-25
Make the render cache work with all the data providers

e0ipso:
There are several `TODO` in the code base regarding this.


e0ipso:

e0ipso:
Now that I think of it, the cache invalidation part on update is still pending

e0ipso:
@amitaibu tests are green, including new ones.

A semi-important bug was fixed on the render cache layer with this.

plugins/restful/RestfulBase.php
```diff
@@ -1085,9 +1105,13 @@ protected function generateCacheId(array $context = array()) {
     $version = $this->getVersion();
     $cid = 'v' . $version['major'] . '.' . $version['minor'] . '::uu' . $this->getAccount()->uid . '::pa';
     $cid_params = array();
-    $request = $this->getRequest();
-    static::cleanRequest($request);
-    $options = $context + $request;
+    $options = $context;
+    if ($this->isReadMethod($this->getMethod())) {
+      // We don't want to split the cache with the body data on write requests.
```
amitaibu, line 1110: Can you explain what you mean by "split the cache"?

amitaibu:
> A semi-important bug was fixed on the render cache layer with this.

Also can you clarify what was the bug :)

plugins/restful/RestfulBase.php
e0ipso, line 1110: It means having different versions of the cache due to generating different cache ids.

This is the bug I was referring.

amitaibu:
Merged, thanks.

PR #275, opened 2014-11-26
getAccount performance

e0ipso:
Improve the performance of getAccount.

### DIFF
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -921,6 +921,10 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
    *   The user object.
    */
   public function getAccount($cache = TRUE) {
+    $account = &drupal_static(__CLASS__ . '::' . __FUNCTION__);
+    if (isset($account)) {
+      return $account;
+    }
     // The request.
     $request = $this->getRequest();
     // The HTTP method. Defaults to "get".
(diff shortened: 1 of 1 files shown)


e0ipso:
After all, this turns out not to be a very good idea.

amitaibu:
I'm also seeing getAccount getting a big hit, on a clean Drupal with 100 articles, and an auth user doing `api/v1.5/articles`

xhgui__hierarchical_profiler_report_and_message_subscribe_email test_-_vlad_-___applications_mamp_htdocs_vlad_

> After all, this turns out not to be a very good idea.

Why not?

amitaibu:
#276

amitaibu:
so, looking closely, I think you were in the right direction. possible also adding `drupal_static_reset()` to `\RestfulBase::process()`. I'll open a PR.

e0ipso:
Caching a user is scary. specially since we cannot rely on the user being the same throughout all the request (because of the Drupal use of the API).

I'm up for this, but this is not the approach.

amitaibu:
I'm thinking something like this:

``` php
public function getAccount($cache = TRUE) {
    $cached_uid = &drupal_static(__CLASS__ . '::getAccount', array());

    // The request.
    $request = $this->getRequest();
    // The HTTP method. Defaults to "get".
    $method = $this->getMethod();

    $unique_identifier = md5(serialize($request) . $method);

    if (isset($cached_uid[$unique_identifier])) {
      return $cached_uid ? user_load($cached_uid[$unique_identifier]) : drupal_anonymous_user();
    }

    $account = $this->getAuthenticationManager()->getAccount($request, $method, $cache);
    $cached_account[$unique_identifier] = $account->uid;
   ... 
```

I'll check, profile, and open a PR :)

PR #278, opened 2014-11-28
Add global option for caching resources

amitaibu:
Like this switching resources that didn't explicitly set the caching would be easier.


amitaibu:
@mateu-aguilo-bosch what do you think about this?

e0ipso:
Fantastic!

PR #279, opened 2014-11-29
Missing dependency.

e0ipso:
It now depends on `entityreference`.


e0ipso:
I'm going to merge this, since this is trivial.

PR #280, opened 2014-11-29
Cache id performance

e0ipso:


e0ipso:
Not sure if all these static caches will make using the Drupal API useless. Many of them assume one call per request and have to be cleared manually.

Maybe we should detect if the call is made using the internal API by checking `$request['__application']['rest']` and clear the statics so we don't have to do it manually.

amitaibu:
How much do we gain from those improvements?

e0ipso:
I haven't done any measures but this showed up in my profiling. We're short circuiting `getVersion`, `getAccount` and several string manipulation calls.

amitaibu:
> Maybe we should detect if the call is made

You can invalidate_cache in `process()` - which is the entry point for all calls.

e0ipso:
Anyways, that'll be a new pr. This one can be merged

amitaibu:
Merged, thanks.

PR #224, opened 2014-11-30
Verify external caches are working properly

e0ipso:
https://www.drupal.org/node/2364011

(worth also checking https://www.drupal.org/node/2364011#comment-9291895)


e0ipso:
Also, we are now safe with the new use of the `'Vary'` header.

includes/RestfulManager.php
```diff
@@ -306,4 +306,31 @@ public static function echoMessage($value, $message) {
...
 
+  /**
+   * Performs end-of-request tasks.
+   *
+   * This function sets the page cache if appropriate, and allows modules to
+   * react to the closing of the page by calling hook_exit().
+   *
+   * @see drupal_page_footer().
+   */
+  public static function pageFooter() {
```
amitaibu, line 320: why not use `drupal_page_footer()`?:
1. Is there a use case for needing our own `exit` hook instead of the regular one?
2. Is there a use case for setting the `cache` variable, but not setting restful's page cache?

amitaibu:
> I can confirm it is not working.

btw, how did you check this? By checking if varnish returns a hit?

includes/RestfulManager.php
e0ipso, line 320: > Is there a use case for needing our own exit hook instead of the regular one?

It's the only way of running a hook whenever a request dispatched by RESTful is made. It will run for cached and uncached responses. It is interesting to do things like custom statistics of the API usage.

> Is there a use case for setting the cache variable, but not setting restful's page cache?

I think so. When Drupal is serving content and RESTful is exposing some other content you may want to cache one and not the other, depending on the nature of the content you're serving on each sub-system.

> why not use drupal_page_footer()?:

To be able to use our own `module_invoke_all('restful_exit')` and `variable_get('restful_page_cache', FALSE)`.

amitaibu, line 320: how about keeping `pageFooter()`, but only as a wrapper around `drupal_page_footer`, and maybe move the current implementation to the docs under the "super-advanced-mateu-gone-wild" section? :P

e0ipso, line 320: Alright, point taken. Time to simplify :-P

e0ipso:
@amitaibu this is ready again.

restful.admin.inc
```diff
@@ -55,10 +55,24 @@ function restful_admin_settings($form_state) {
...
 
-  $form['restful_render_cache'] = array(
+  $form['restful_cache'] = array(
+    '#type' => 'fieldset',
+    '#title' => t('Cache'),
+    '#description' => t('Cache options for the different layers of RESTful.'),
+    '#collapsible' => FALSE,
+  );
+
+  $form['restful_cache']['restful_page_cache'] = array(
```
amitaibu, line 65: is this still needed?

restful.api.php
```diff
@@ -24,5 +24,29 @@ function hook_restful_parse_request_alter(&$request) {
...
+ * Only use this hook if your code must run even for cached page views.
+ * If you have code which must run once on all non-cached pages, use
+ * hook_init() instead. That is the usual case. If you implement this hook
+ * and see an error like 'Call to undefined function', it is likely that
+ * you are depending on the presence of a module which has not been loaded yet.
+ * It is not loaded because Drupal is still in bootstrap mode.
+ *
+ * @see hook_exit().
+ */
+function hook_restful_exit() {
```
amitaibu: can be removed

e0ipso: Done.

restful.admin.inc
e0ipso, line 65: This is just markup. It's merely informative. I'd like to keep it.

restful rest performance 2014-12-02 08-27-05

e0ipso:
Merge conflicts solved.

amitaibu:
Looks good - Once green it can be merged.

amitaibu:
Merged, thanks!

PR #274, opened 2014-11-30
Allow configuring the range (max items per page in listings)

e0ipso:
Right now this is what you are forced to do to override the limit of 50:

``` php
<?php

/**
 * @file
 * Contains RestfulExampleArticlesResource.
 */

class RestfulPerformanceArticlesResource__1_0 extends RestfulEntityBaseNode {

  protected $range = 300;

  /**
   * Overrides RestfulExampleArticlesResource::publicFieldsInfo().
   */

    $public_fields['body'] = array(
      'property' => 'body',
      'sub_property' => 'value',
    );

    return $public_fields;
  }
}
```

### DIFF
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -661,10 +661,17 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
     $this->setMethod($method);
     $this->setPath($path);
     $this->setRequest($request);
-    if (!empty($request['range']) && is_int($request['range']) && $request['range'] < $this->getRange()) {
-      // If there is a valid range property in the request override the range.
-      $this->setRange($request['range']);
+    // Override the range with the value int he URL.
+    if (!empty($request['range'])) {
+      if (!ctype_digit((string) $request['range']) || $request['range'] < 1) {
+        throw new \RestfulBadRequestException('"Range" property should be numeric and higher than 0.');
+      }
+      if ($request['range'] < $this->getRange()) {
+        // If there is a valid range property in the request override the range.
+        $this->setRange($request['range']);
+      }
     }
+
     $version = $this->getVersion();
     $this->setHttpHeaders('X-API-Version', 'v' . $version['major']  . '.' . $version['minor']);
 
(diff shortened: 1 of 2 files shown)


amitaibu:
actually I think we should only deal here with passing in the URL. I think it's fine to keep the `$range` variable in the class

plugins/restful/RestfulBase.php
```diff
@@ -661,6 +661,10 @@ public function process($path = '', array $request = array(), $method = \Restful
     $this->setMethod($method);
     $this->setPath($path);
     $this->setRequest($request);
+    if (!empty($request['range'])) {
```
amitaibu, line 665: is this `range` being taken into account in the cache?
I would also verify `is_int` of the value.

e0ipso, line 665: > is this range being taken into account in the cache?

Everything in the query string is taken into account in the cache ID generation.

> I would also verify is_int of the value.

Good point.

amitaibu, line 665: 1) If range is wrong we should throw an exception
2) we shouldn't allow a range higher than $range as it may be abused by malicious users

‫ב-1 בדצמ 2014, בשעה 08:39, ‏Mateu Aguiló Bosch notifications@github.com כתב/ה:‬

e0ipso, line 665: > If range is wrong we should throw an exception

Why not ignore altogether?

> we shouldn't allow a range higher than $range as it may be abused by malicious users

:+1:

amitaibu, line 665: > Why not ignore altogether?

I think it's a nicer DX to be informed you sent something wrong -- we are already doing it for `page` -- https://github.com/Gizra/restful/blob/ef2310b21fbc0317af9b62ec0c09ec9b96504cd5/plugins/restful/RestfulBase.php#L800-L802

plugins/restful/RestfulBase.php
```diff
@@ -661,6 +661,17 @@ public function process($path = '', array $request = array(), $method = \Restful
     $this->setMethod($method);
     $this->setPath($path);
     $this->setRequest($request);
+    // Override the range with the value int he URL.
```
amitaibu: int he => in the

After that it can be merged

e0ipso:
Merged.

PR #288, opened 2014-12-02
ng-admin integration

e0ipso:
This will provide basic ng-admin integration.

ng-admin is not fully functional since I know @amitaibu has been playing with it and can replace the necessary code in `src/restful-app.js`. My ultimate goal is to configure ng-admin by making `OPTIONS` calls to the RESTful API, but one step at a time.

**Important change:** This PR will turn the RESTful Angular example module into an AngularJS app repository, and it will serve them without the Drupal theme to convey that they are separated front-end apps. See the GIF.

angular-app-repo


amitaibu:
@mateu-aguilo-bosch you are so awesome!

modules/restful_angular_example/components/restful-app/dist/restful-app.js
```diff
@@ -59,6 +55,77 @@ angular.module('restfulApp', [
     }];
   });
 
+if (restfulExampleModules.indexOf('ng-admin') !== -1) {
```
amitaibu, line 59: why not have a different app just for ng-admin?

modules/restful_angular_example/components/restful-app/dist/restful-app.js
```diff
@@ -59,6 +55,77 @@ angular.module('restfulApp', [
...
+  // This configuration is specific for ng-admin. Do not add it for other
+  // examples.
+  app.config(function (NgAdminConfigurationProvider, Application, Entity, Field, Reference, ReferencedList, ReferenceMany) {
+    // set the main API endpoint for this admin
+    var app = new Application('RESTful Admin')
+      .baseApiUrl('http://localhost/restful');
```
amitaibu: url should be passed from drupal (`url('api', array('absolute' => TRUE)`)

e0ipso: :+1:

modules/restful_angular_example/templates/restful-angular-html.tpl.php
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML+RDFa 1.0//EN"
```
amitaibu, line 1: Why not show this inside the normal Drupal page?

and1truong:
Why don't we create base api, then reuse in example module?

amitaibu:
> Why don't we create base api

What do you mean by that?

modules/restful_angular_example/templates/restful-angular-html.tpl.php
e0ipso, line 1: I want it to be outside of any Drupal template for 2 reasons.
1. We're trying to simulate a front end app that lives outside Drupal. We want to avoid the page having any customizations the current theme has for `html.tlp.php`.
2. We have full control on the HTML, just as we do in these AngularJS apps.

modules/restful_angular_example/components/restful-app/dist/restful-app.js
e0ipso, line 59: The problem of separating apps is that if we go down that path se should probably want separate `package.json`, `bower.json`, folder structure… That adds extra complication that can be avoided with this `if`.

amitaibu:
ready for review?

e0ipso:
> ready for review?

Not yet. I'm taking a break from this one…

e0ipso:
This is as ready as it's going to be for now. Let's merge it.

amitaibu:
Merged. Seriously, this should have a blog post!

e0ipso:
I think I may write it ☺

e0ipso:
The thing is that I don't see this as finished, until the schema for the fields is read from OPTIONS calls.

amitaibu:
> The thing is that I don't see this as finished, until the schema for the fields is read from OPTIONS calls.

Maybe, but we're still providing out of the box something far more advanced then anything else I've seen implemented in Angular + Drupal. Don't be a purist ;)

PR #281, opened 2014-12-07
Allow PER_USER/ PER_ROLE cache garnularity

e0ipso:
Currently our granularity is `PER_USER`, which might be over paranoiac when dealing with content that all users may consume in the same way. In that case It would significantly improve performance for users as only the first user in the role will get the cache miss.

@mateu-aguilo-bosch thoughts?


e0ipso:
This is ready for review.

It depends on #289.

amitaibu:
You are Incredible! :)

plugins/restful/RestfulEntityBase.php
```diff
@@ -1384,4 +1384,19 @@ public function getImageUris(array $file_array, $image_styles) {
...
+  /**
+   * Clear all caches corresponding to the current resource for a given entity.
+   *
+   * @param int $id
+   *   The entity ID.
+   */
+  public function clearResourceRenderedCacheEntity($id) {
+    // Build the cache ID.
+    $version = $this->getVersion();
+    $cid = 'v' . $version['major'] . '.' . $version['minor'] . '::' . $this->getResourceName() . '::uu' . $this->getAccount()->uid . '::paet:';
```
amitaibu, line 1396: At one stage we should think how to improve these strings. Maybe passing an array to a helper function that will sort by alphabet, and know how to do the abbreviations, as currently it's very error prone.

amitaibu:
Needs a re-roll. After that can be merged.

e0ipso:
Merged.

PR #302, opened 2014-12-09
Allow rendering entities providing a view mode.

e0ipso:
The idea is to leverage Drupal's configuration capabilities to specify which fields to render and how to render them using the field formatters.

I think it's _great_ to have the fine grain control that we have now on how to build the output, but we're core missing a lot from Drupal core by not using the entity view modes.
## Goals:
- Minimal configuration. Probably without even needing a custom class that extends from `RestfulNodeBase`, `RestfulEntityBase`, …
- Generating the **exact** desired output **without code**.
- The only rendering limitations should be those inherited from rendering a view mode.
## Suggested approach:
- Provide a `'view_mode'` configuration key in the plugin definition.
- Provide a mapping config array between the keys in the rendered view mode and the desired output.
  - This mapping would be a replace for the default field info method, from there on everything else should be **the same** from the implementation standpoint.
  - This mapping should be seen as plugin configuration, not as code.
  - This mapping should allow path nesting. This of a view mode that has an ER field and renders the referenced entity with only its body. We would do `'field_referenced_node' => 'referenced::body::value'`. That could yield:

``` javascript
{
  "referenced": {
    "body": {
      "value": "Lorem ipsum dolor…"
    }
  }
}
```


e0ipso:
I think this could be a deal breaker in DX.

cc @justafish

e0ipso:
With this PR we can now expose a resource by just doing [this](https://github.com/mateu-aguilo-bosch/restful/blob/302-entity-view-modes/modules/restful_example/plugins/restful/node/articles/1.7/articles__1_7.inc):

``` php
<?php

$plugin = array(
  'label' => t('Articles'),
  'resource' => 'articles',
  'name' => 'articles__1_7',
  'entity_type' => 'node',
  'bundle' => 'article',
  'class' => 'RestfulEntityBaseNode',
  'authentication_types' => TRUE,
  'authentication_optional' => TRUE,
  'minor_version' => 7,
  // Add the view mode information.
  'view_mode' => array(
    'name' => 'default',
    'mappings' => array(
      'body' => 'body',
      'field_tags' => 'tags',
      'field_image' => 'image',
    ),
  ),
);
```

e0ipso:
w00t! Tests came back green @amitaibu. This is ready for review.

e0ipso:
BTW, this was **surprisingly easy** thanks to the flexibility that we have built in the module so far. Take a moment to congratulate yourself.

This could have even been a contrib module.

modules/restful_example/plugins/restful/node/articles/1.7/articles__1_7.inc
```diff
@@ -0,0 +1,23 @@
...
+  'bundle' => 'article',
+  'description' => t('Export the article content type using view modes.'),
+  'class' => 'RestfulEntityBaseNode',
+  'authentication_types' => TRUE,
+  'authentication_optional' => TRUE,
+  'minor_version' => 7,
+  // Add the view mode information.
+  'view_mode' => array(
+    'name' => 'default',
+    'mappings' => array(
```
amitaibu: maybe `mappings` => `field_map`?

Anyway, this could be optional. If a field that exists in the display isn't present here in the array we can assume the public field name to be `str_replace('field_', '', $field_name)`

modules/restful_example/plugins/restful/node/articles/1.7/articles__1_7.inc
```diff
@@ -0,0 +1,23 @@
+<?php
+
+$plugin = array(
+  'label' => t('Articles'),
+  'resource' => 'articles',
+  'name' => 'articles__1_7',
+  'entity_type' => 'node',
+  'bundle' => 'article',
+  'description' => t('Export the article content type using view modes.'),
+  'class' => 'RestfulEntityBaseNode',
```
amitaibu, line 10: Just an idea (for follow-up), but we could have been smarter about choosing the base class, based on the entity we use the `view_mode` for.

includes/RestfulEntityViewMode.php
```diff
@@ -0,0 +1,151 @@
...
+
+      // Add it to the public fields array with a special callback function.
+      $public_fields[$mappings[$field_name]] = array(
+        'callback' => array(
+          array($this, 'renderField'),
+          array($field_name, $view_mode),
+        ),
+      );
+    }
+    if (empty($public_fields)) {
```
amitaibu, line 79: worth also throwing an exception for invalid view_mode

includes/RestfulEntityViewMode.php
```diff
@@ -0,0 +1,151 @@
...
+   *   An array of field names that are displayed in this view mode.
+   */
+  protected function displayedFieldsList($view_mode) {
+    $entity_field_instances = field_info_instances($this->entityType);
+    $bundle_field_instances = reset($entity_field_instances);
+    if ($bundle = $this->bundle) {
+      $bundle_field_instances = $entity_field_instances[$bundle];
+    }
+    // Traverse the field instance to get all non hidden fields.
+    $displayed_fields = array_map(function ($field_instance) use ($view_mode) {
```
amitaibu, line 102: that's a first in RESTful. are we doing that way? ;)

includes/RestfulEntityViewMode.php
```diff
@@ -0,0 +1,151 @@
...
+   *   The entity wrapper.
+   * @param string $field_name
+   *   The field name to render.
+   * @param string $view_mode
+   *   The view mode to use.
+   *
+   * @return mixed
+   *   The output value to be rendered.
+   */
+  public function renderField(\EntityDrupalWrapper $wrapper, $field_name, $view_mode) {
```
amitaibu, line 126: maybe this is the capability that should be generalised (in a follow up?) -- allowing to pass a public field through a formatter.

includes/RestfulEntityViewMode.php
amitaibu, line 79: Oh, we actually do it [here](https://github.com/Gizra/restful/pull/302/files#diff-3c241a6c3b4e93a88a7da3e91c3e9358R1134)

plugins/restful/RestfulEntityBase.php
```diff
@@ -1130,6 +1130,16 @@ public function publicFieldsInfo() {
       ),
     );
 
+    if ($view_mode_info = $this->getPluginKey('view_mode')) {
+      if (empty($view_mode_info['name'])) {
+        throw new \RestfulServerConfigurationException('View mode not found.');
+      }
+      $helper = new \RestfulEntityViewMode($this->getEntityType(), $this->getBundle());
```
amitaibu: $helper sounds weird. Maybe `$view_mode_handler`?

amitaibu:
> BTW, this was surprisingly easy thanks to the flexibility that we have built in the module so far.

Indeed, my first reaction when I read the issue was "move to contrib". But given how very little is needed and how useful it can be, I'm ok with that. Thanks for working on this!

includes/RestfulEntityViewMode.php
e0ipso, line 102: I'm doing node.js these days, can you tell?

Actually I think that it's a pretty good pattern supported by php5.3

modules/restful_example/plugins/restful/node/articles/1.7/articles__1_7.inc
e0ipso: > Anyway, this could be optional. If a field that exists in the display isn't present here in the array we can assume the public field name to be str_replace('field_', '', $field_name)

I was debating about this as well. I don't like the idea of doing these automated things, I like enforcing being explicit with what you expose. That being said, I think that it would be pretty cool to have a zero-conf way to expose your resources with RESTful.

My take, if we want to be explicit about it leave it in here. If we want to break the pattern and go for the zero-conf approach, then make it contrib.

modules/restful_example/plugins/restful/node/articles/1.7/articles__1_7.inc
e0ipso, line 10: Same as in https://github.com/Gizra/restful/pull/302#discussion_r21586467

includes/RestfulEntityViewMode.php
e0ipso, line 126: Care to create the issue? I'm not sure I understand the implications of what you say.

amitaibu, line 126: better, I'll create the PR ;)

amitaibu:
Go ahead - you can merge this :)

e0ipso:
Merged.

PR #315, opened 2014-12-14
Additional hateoas

e0ipso:
This will allow us to add generic HATEOAS information from the handler, like facets, without having to override the formatter.


amitaibu:
Looks good. maybe worth implementing in one of the example plugins.

Issue #317, opened 2014-12-15
Check output with Entity Translation enabled.


e0ipso:
See https://github.com/RESTful-Drupal/restful/issues/314#issuecomment-66924934 by @attiks.

e0ipso:
Closed. Thanks @Jelle-S!!

PR #318, opened 2014-12-15
Fixes Issue #317: Check output with Entity Translation enabled.

Jelle-S:


e0ipso:
This looks good.

plugins/restful/RestfulEntityBase.php
```diff
@@ -268,6 +268,7 @@ protected function getQueryResultForAutocomplete() {
    * {@inheritdoc}
    */
   public function viewEntity($entity_id) {
+    global $language;
```
amitaibu, line 271: I think language should be injected rather then using the global one

amitaibu, line 271: So you can also use the API to get the right language without mocking with the `global $language` variable.

e0ipso, line 271: Maybe just a new property in `\RestfulBase` that is (optionally) initialized in the constructor. If no value is passed to the constructor, then use `global $language`.

This can be a good follow up.

PR #321, opened 2014-12-15
Set language in constructor of \RestfulBase 

Jelle-S:
(follow-up of #318)

### DIFF
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -76,6 +76,13 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
   protected $valueMetadata = array();
 
   /**
+   * Determines the language of the items that should be returned.
+   *
+   * @var string
+   */
+  protected $langcode;
+
+  /**
    * Static cache controller.
    *
    * @var \RestfulStaticCacheController
@@ -212,6 +219,25 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
   }
 
   /**
+   * Get the language code.
+   *
+   * @return string
+   */
+  public function getLangCode() {
+    return $this->langcode;
+  }
+
+  /**
+   * Sets the language code.
+   *
+   * @param string $langcode
+   *   The language code.
+   */
+  public function setLangCode($langcode) {
+    $this->langcode = $langcode;
+  }
+
+  /**
    * Set the request array.
    *
    * @param array $request
@@ -345,7 +371,7 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
    * @param DrupalCacheInterface $cache_controller
    *   (optional) Injected cache backend.
    */
-  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL, $langcode = NULL) {
     parent::__construct($plugin);
     $this->authenticationManager = $auth_manager ? $auth_manager : new \RestfulAuthenticationManager();
     $this->cacheController = $cache_controller ? $cache_controller : $this->newCacheObject();
@@ -353,6 +379,13 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
       $this->setRateLimitManager(new \RestfulRateLimitManager($this->getPluginKey('resource'), $rate_limit));
     }
     $this->staticCache = new \RestfulStaticCacheController();
+    if (is_null($langcode)) {
+      global $language;
+      $this->langcode = $language->language;
+    }
+    else {
+      $this->langcode = $langcode;
+    }
   }
 
   /**
--- a/plugins/restful/RestfulEntityBaseMultipleBundles.php
+++ b/plugins/restful/RestfulEntityBaseMultipleBundles.php
@@ -27,8 +27,8 @@ class RestfulEntityBaseMultipleBundles extends RestfulEntityBase {
     );
   }
 
-  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
-    parent::__construct($plugin);
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL, $language = NULL) {
+    parent::__construct($plugin, $auth_manager, $cache_controller, $language);
 
     if (!empty($plugin['bundles'])) {
       $this->bundles = $plugin['bundles'];
--- a/tests/RestfulViewEntityMultiLingualTestCase.test
+++ b/tests/RestfulViewEntityMultiLingualTestCase.test
@@ -1,0 +1,177 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulViewEntityMultiLingualTestCase
+ */
+
+class RestfulViewEntityMultiLingualTestCase extends RestfulCurlBaseTestCase {
+
+  /**
+   * The user to run the tests as.
+   */
+  protected $testUser;
+
+  public static function getInfo() {
+    return array(
+      'name' => 'View multilingual entity',
+      'description' => 'Test the viewing of a multilingual entity.',
+      'group' => 'Restful',
+    );
+  }
+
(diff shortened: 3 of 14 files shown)


e0ipso:
Awesome PR!

plugins/restful/RestfulBase.php
```diff
@@ -345,7 +371,7 @@ public function getRateLimitManager() {
    * @param DrupalCacheInterface $cache_controller
    *   (optional) Injected cache backend.
    */
-  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL, $langcode = NULL) {
```
e0ipso, line 374: We will probably need to update the inheriting constructors too.

plugins/restful/RestfulBase.php
```diff
@@ -76,6 +76,13 @@
   protected $valueMetadata = array();
 
   /**
+   * Determines the language of the items that should be returned.
+   *
+   * @var string
+   */
+  protected $langcode;
```
e0ipso, line 83: I like having it in `\RestfulBase` even that this property is probably only interesting to entities.

Jelle-S:
Inherited constructors are updated ;-)

amitaibu:
Thanks @Jelle-S ! I'm going to push our luck here -- any chance for a simpleTest? :)

plugins/restful/RestfulEntityBaseMultipleBundles.php
```diff
@@ -27,8 +27,8 @@ public static function controllersInfo() {
     );
   }
 
-  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL) {
-    parent::__construct($plugin);
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL, $language = NULL) {
+    parent::__construct($plugin, $auth_manager, $cache_controller, $language);
```
e0ipso, line 31: Good catch!

e0ipso:
This probably means that we need to adapt our constructor for the Search API integration too. :-)

Awesome work!

Jelle-S:
I'm guessing Travis isn't enabling/downloading Entity Translation? I have no problems on my local machine

amitaibu:
I think we don't need Entity Translation, as we can just enable `translatable` on field. So only need to enable locale.

amitaibu:
> I think we don't need Entity Translation

btw, In case I'm wrong - we can download the module via `.travis.yml`

Jelle-S:
The tests failed withouth Entity Translation (just setting `translatable` wasn't enough)

Jelle-S:
Ah, I didn't apply the two patches for entity on my local setup, updated tests coming soon

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,180 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulViewEntityMultiLingualTestCase
+ */
+
+class RestfulViewEntityMultiLingualTestCase extends RestfulCurlBaseTestCase {
```
amitaibu, line 8: Thanks for the test!

I think we should trim it down a bit, though. I would expect a smaller test case, where we simply assert a single/ multi value fields for 2 different languages - so the test would be really focused on the functionality, easier to maintain, and easier to follow.

Also, I do believe entity translation shouldn't be needed. For example, the Message module uses only local to provide translatable fields.

Jelle-S:
New (simplified) tests without entity translation

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+        'text_processing' => 0,
+      ),
+    );
+    field_create_instance($instance);
+  }
+
+  /**
+   * Test viewing an entity (GET method).
+   *
+   * v1.0 - Simple entity view (id, label, self).
```
amitaibu: I believe this can be removed.

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+   * v1.1 - Text and entity reference fields.
+   * v1.2 - "callback" and "process callback".
+   * v1.3 - Non-existing "callback" property.
+   * v1.4 - Non-existing "process callback" property.
+   * v1.6 - XML output format.
+   */
+  function testViewEntity() {
+    $user = $this->drupalCreateUser();
+    $entity = entity_create(
```
amitaibu: matter of taste but I would prefer:

``` php
$values = array(
  'name' => 'restful_test_translatable_entity',
);
$entity = entity_create( 'restful_test_translatable_entity', $values);
```

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+      'label' => t('Text multiple'),
+      'settings' => array(
+        'text_processing' => 0,
+      ),
+    );
+    field_create_instance($instance);
+  }
+
+  /**
+   * Test viewing an entity (GET method).
```
amitaibu: Maybe better `Test viewing an entity with translatable fields.`

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+   *
+   * v1.0 - Simple entity view (id, label, self).
+   * v1.1 - Text and entity reference fields.
+   * v1.2 - "callback" and "process callback".
+   * v1.3 - Non-existing "callback" property.
+   * v1.4 - Non-existing "process callback" property.
+   * v1.6 - XML output format.
+   */
+  function testViewEntity() {
```
amitaibu: `testViewEntity` => `testViewMultiLangualEntity`

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+      array(
+        'name' => 'restful_test_translatable_entity',
+        'uid' => $user->uid,
+        'label' => 'Test translation',
+      )
+    );
+    $wrapper = entity_metadata_wrapper('restful_test_translatable_entity', $entity);
+
+    $text1 = array(
+      'en' => $this->randomName(),
```
amitaibu, line 110: I think out assert should be "smarter" and assert we indeed got the expected results. So we should probably have for example `$text1_en = $this->randomName()`, which we can later compare the results to

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   * Helper to test viewing an entity (GET method) in a certain language.
+   *
+   * @param string $langcode
+   *   The language to view the entity in.
+   * @param int $id
+   *   The ID of the entity to test.
+   */
+  private function _testViewEntityLang($langcode, $id, $text1, $text2) {
```
amitaibu: Maybe better: `_testViewEntityLang` => `assertExpectedResult`

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+    $handler = restful_get_restful_handler('restful_test_translatable_entity');
+    $handler->setLangCode($langcode);
```
amitaibu, line 146: `// Explicitly set the langcode.`

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -0,0 +1,171 @@
...
+    $response = $handler->get($id);
+    $result = $response[0];
+
+    $expected_result = array(
```
amitaibu: We can probably just assert specifically the `text_single`/ `text_multiple` results.

amitaibu:
Thanks @Jelle-S , I've added a few comments. We are getting very close! :)

Jelle-S:
about the `$text1_en`:
That is basically what we do now? we pass `$text1_en[$langcode]` to `_testViewEntityLang()`

amitaibu:
> That is basically what we do now?

Sorry, you are right.

Jelle-S:
The tests have been updated according to your comments.

amitaibu:
Merged, thanks!

Issue #327, opened 2014-12-17
\RestfulAuthenticationBasic::authenticate shouldn't always trigger flood_register_event


amitaibu:
This was triggered when we called multiple times resources via API.

Basically, the solution should probably be that if called via API, we shouldn't try to authenticate using the plugins, and just use the global $user (as we already do, but probably to late in the process)

amitaibu:
or a better approach would be to check if indeed some info was sent to base auth, and flood_register_event() only then.

I'll be working on a PR.

e0ipso:
Closed in an issue purge. Please, feel free to reopen if necessary.

PR #328, opened 2014-12-17
WIP: Don't register flood event for base auth, if no attempt has been made.

amitaibu:
#327


amitaibu:
Next is adding a test

amitaibu:
Seems the PR is wrong as we already check this in `\RestfulAuthenticationBasic::applies`

I'll need to try and reproduce on the existing installation why it happened. Maybe the server somehow populated `$_SERVER['PHP_AUTH_USER']` and `$_SERVER['PHP_AUTH_PW']`

amitaibu:
Turns out the "problem" was `.htaccess` password. Once that is set, it indeed populates `$_SERVER['PHP_AUTH_USER']` and `$_SERVER['PHP_AUTH_PW']`

@mateu-aguilo-bosch maybe worth adding a word about it in the docs?

e0ipso:
@amitaibu yeah, I agree. This is going to be a typical thing.

Issue #282, opened 2014-11-29
Notice on \RestfulFormatterHalJson::moveReferencesToEmbeds


amitaibu:
I've seen a notice when I have my reference as an ID only:

``` php
  if (field_info_instance($this->getEntityType(), OG_AUDIENCE_FIELD, $this->getBundle())) {
      $public_fields['company'] = array(
        'property' => OG_AUDIENCE_FIELD,
        'resource' => array(
          'company' => array(
            'name' => 'companies',
            'full_view' => FALSE,
          )
        ),
      );
    }
```

amitaibu:
@mateu-aguilo-bosch @mirzu - If the referenced resource is shown only as ID, should we indeed move it to the `_embedded` - or should we keep it in place?

e0ipso:
I would say that an id is just a regular field, that is not the result of calling another resource. Therefore I would leave it in place, out of `_embedded`.

amitaibu:
Yeah, that was my same line of thought. Thanks.

‫ב-1 בדצמ 2014, בשעה 08:37, ‏Mateu Aguiló Bosch notifications@github.com כתב/ה:‬

e0ipso:
Closed in an issue purge. Please, feel free to reopen if necessary.

PR #329, opened 2014-12-18
Use a strategy pattern to get the field value

e0ipso:


e0ipso:
@amitaibu do you like the direaction this is taking? The hard one, of course, is being the EFQ data provider.

amitaibu:
What would be the advantage here - is it mainly to reduce the code duplication?

e0ipso:
@amitaibu well, applying an _Strategy Pattern_ has several advantages like: decoupling class responsibilities, coping better with change, more flexibility and runtime behaviour swapping.

The main objective is keeping the codebase in good shape, so we can keep having good surprises of how easy to implement are things. Like the ones we had with the view modes and the _restful_search_api_ module.

e0ipso:
I discovered that HAL JSON was not moving the embeds for fields/properties with cardinality = 1. That is fixed now.

amitaibu:
> I discovered that HAL JSON was not moving the embeds for fields/properties with cardinality = 1. That is fixed now.

I think it's #282?

> The main objective is keeping the codebase in good shape, so we can keep having good surprises of how 

I understand, but still would like to better understand how it can help us on a concrete real life use case. As the counter part of this PR is that is raises the complexity of the module. I have no problem with more complexity - as long as it has a value that can be easily recognized.

amitaibu:
not to say your PR isn't great - but just making sure we keep the "not theory - but real solution" approach we have so far :)

e0ipso:
> a concrete real life use case. 

@amitaibu I think that the _real use case scenario_ applies to features. This PR does not address any new feature. It just reorganizes the code and rewrites some parts for the afore mentioned benefits: _decoupling class responsibilities, coping better with change, more flexibility and runtime behaviour swapping_.

> As the counter part of this PR is that is raises the complexity of the module.

I was actually thinking about the opposite. [Cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) –which is usually the measurement for code complezity– is reduced significantly with this PR. I think that it's just that as Drupal developers we're not used to have this many classes.

> we keep the "not theory - but real solution" approach we have so far :)

Yes! I actually did this PR with the module's core values in mind (practical + best practises). In this case using this design patter is a best practise when writing maintainable code.

Let me know if you feel strongly about it. If not, I'll fix the tests.

amitaibu:
> Let me know if you feel strongly about it.

No, I feel ok with it - even though I don't see the immediate benefit I trust your judgment.

e0ipso:
RESTful love :heart_decoration: 

Great, I'll try to get those green.

includes/property/retriever/RestfulPropertyValueRetrieverEntity.php
```diff
@@ -0,0 +1,47 @@
...
+class RestfulPropertyValueRetrieverEntity implements \RestfulPropertyValueRetrieverInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function retrieve(array $info, \RestfulPropertySourceInterface $source) {
+    $value = NULL;
+    // If there is a callback defined execute it instead of a direct mapping.
+    if ($info['callback']) {
+      $value = \RestfulBase::executeCallback($info['callback'], array($source));
```
e0ipso, line 17: As much as I would like that all the data provider callback methods had the same signature (getting a \RestfulPropertySourceInterface as the first parameter) this is an API change to an RC :-( This is a central piece of code that almost every site using RESTful will be using.

I'll keep backwards compatibility with that.

amitaibu, line 17: not sure I follow. How does the code you would like to do (if it wasn't RC) look like?

e0ipso, line 17: ### Right now

Let's say I provide a `'callback' => array(array($this, 'myCallback'), array($arg1, $arg2))`.

This is the function signature for an EFQ provider:

``` php
  /**
   * My callback
   * 
   * @param \EntityMetadataWrapper $value
   * @param mixed $arg1
   * @param mixed $arg2
   * 
   * @return mixed
   */
  public function myCallback(\EntityMetadataWrapper $value, $arg1, $arg2) {
    ...
  }
```

This is the function signature for a DB provider:

``` php
  /**
   * My callback
   * 
   * @param object $value
   * @param mixed $arg1
   * @param mixed $arg2
   * 
   * @return mixed
   */
  public function myCallback($value, $arg1, $arg2) {
    ...
  }
```

This is the function signature for a CTools provider:

``` php
  /**
   * My callback
   * 
   * @param array $value
   * @param mixed $arg1
   * @param mixed $arg2
   * 
   * @return mixed
   */
  public function myCallback(array $value, $arg1, $arg2) {
    ...
  }
```

### Desired scenario

As I said, this will break a lot of sites using RESTful.

This is the function signature for **all** providers:

``` php
  /**
   * My callback
   * 
   * @param \RestfulPropertySourceInterface $source
   * @param mixed $arg1
   * @param mixed $arg2
   * 
   * @return mixed
   */
  public function myCallback(\RestfulPropertySourceInterface $source, $arg1, $arg2) {
    $value = $source->getSource();
    ...
  }
```

amitaibu, line 17: Two options:
1. Break compatibility. Even at RC I think it's fine when the value is great
2. Pass \RestfulPropertySourceInterface $source as the 4th argument.

I'm actually leaning to (1).

e0ipso, line 17: I'm actually thinking about postponing this PR indefinitely… The `RestfulEntityBase` code is too coupled.

I think that a possible version 2 of the module ca go through a rewrite with the requirements in mind. The problem here was that we grew slowly without realizing how much we could accomplish.

amitaibu, line 17: > The RestfulEntityBase code is too coupled.

Can you explain which parts you would like to change. I think that at this point we can still do breaking changes if their value is high.

Issue #332, opened 2014-12-21
Request should be piped to sub-requests


amitaibu:
I have [this](https://gist.github.com/amitaibu/8925b6d59a8bf715f46e) class where the resource company is "private".

A user tries to get the `me` resource with a token e.g. `https://exmaple.com/api/me?access_token=aH1grLnUp6T-BW5J_mtF9J0lkgJSDItYHgnY2X_ND2w`.

The problem is that the $this->getRequest() doesn't return the previous request, so the `access_token` is lost

PR #333, opened 2014-12-21
Pipe the parent request to the sub-request

amitaibu:
#332

### DIFF
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -228,6 +228,31 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
   }
 
   /**
+   * Gets a request array with the data that should be piped to sub requests.
+   */
+  protected function getRequestForSubRequest() {
+    $piped_request = array();
+
+    foreach ($this->getRequest() as $key => $value) {
+      if (in_array($key, array(
+          'filter',
+          'page',
+          'q',
+          'range',
+          'sort',
+        ))) {
+        continue;
+      }
+
+      $piped_request[$key] = $value;
+    }
+
+    return $piped_request;
+  }
+
+
+
+  /**
    * Sets the language code.
    *
    * @param string $langcode
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -240,6 +240,7 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
           'q',
           'range',
           'sort',
+          'fields',
         ))) {
         continue;
       }
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -220,6 +220,9 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
 
   /**
    * Gets a request array with the data that should be piped to sub requests.
+   *
+   * @return array
+   *   The request array to be piped.
    */
   protected function getRequestForSubRequest() {
     $piped_request = array();
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -219,15 +219,6 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
   }
 
   /**
-   * Get the language code.
-   *
-   * @return string
-   */
-  public function getLangCode() {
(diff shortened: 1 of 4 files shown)


e0ipso:
The challenge here will be to know which parts of the request array we should pass in. There are things that only apply to the “parent” request, like `page`, `range`, `filter`, `sort`, …

I think an opt-in strategy would be the way to go. That means constructing a sub_request array to pass along with the information needed:

``` php
   // Pipe the parent request to the sub-request.
   $request = $this->getRequest();
   $sub_request = array('token' => $request['token'], …);
   $bundle_handler->setRequest($sub_request);
```

e0ipso:
Passing the `sort` (for instance) to the sub-request will cause sorting on an unexisting field an rasing an exception. I'm surprised to see that there is no test covering invalid params for sub-requests. Maybe a follow up?

amitaibu:
Yeah, I guess an opt-in is indeed the way to go. it should probably be a hook/alter, so implementing modules may change

amitaibu:
or maybe we just need to pipe the `__application` (and make sure the access token is moved there)

plugins/restful/RestfulBase.php
```diff
@@ -219,6 +219,31 @@ public function getRequest() {
   }
 
   /**
+   * Gets a request array with the data that should be piped to sub requests.
+   */
+  protected function getRequestForSubRequest() {
+    $piped_request = array();
+
+    foreach ($this->getRequest() as $key => $value) {
+      if (in_array($key, array(
```
amitaibu, line 231: @mateu-aguilo-bosch something like this maybe?

e0ipso, line 231: This is an opt-out strategy, I think it's better to just select the ones we know need to be passed in.

e0ipso:
> or maybe we just need to pipe the __application (and make sure the access token is moved there)

I like this idea. I'm just a bit concerned about 2 things:
1. The `cleanRequest` method will remove the token.
2. Will this affect API calls from within Drupal? (I think it will not).

plugins/restful/RestfulBase.php
amitaibu, line 231: > This is an opt-out strategy

You are right. My reason is 2 cases I think of:
1. `access_token` is defined by restful_token_auth - so restful module shouldn't know about it. With the current approach the access_token is still piped.
1. We are often adding some params via query string (e.g. `https://example.com/api/articles?month=1&year=2015`). Instead of a poor developer having to figure out why their params are not piped to the sub-request we simply filter out the ones that are most probably to crash his system.

e0ipso, line 231: > we simply filter out the ones that are most probably to crash his system

The problem is that we may not know about additional ones that we are not filtering and will also be crashing their system.

e0ipso, line 231: Maybe having a hook that defaults to an empty array. Then restful_access_token can add the token param name.

e0ipso, line 231: I have to say that I dislike hooks. I would much rather having an observer that can update the resource state providing the param names. But the hook is the fast way to go. Maybe a follow up?

amitaibu, line 231: > The problem is that we may not know about additional ones that we are not filtering and will also be crashing their system.

Indeed, but we are applying sensible defaults. For the edge cases you will have to override this method, instead of always override for any custom param you need to pipe.

> Maybe having a hook that defaults to an empty array.

I thought about it at first, but I think it's cleaner that each resource may override it.

e0ipso, line 231: OK. I'm cool with that.

plugins/restful/RestfulBase.php
```diff
@@ -219,6 +219,31 @@ public function getRequest() {
   }
 
   /**
+   * Gets a request array with the data that should be piped to sub requests.
+   */
+  protected function getRequestForSubRequest() {
```
e0ipso, line 227: This is missing a docblock :trollface:

amitaibu:
@mateu-aguilo-bosch Add phpdocs. Ready for review.

e0ipso:
This looks good. The only thing left is that we may want to document this in the docs to surface this feature.

amitaibu:
Docs added.

e0ipso:
This is good to merge.

amitaibu:
Merged.

PR #334, opened 2014-12-22
Allow loading entities by any field.

e0ipso:
I've found need this several times already. I need to be able to load an entity by a unique field or property attached to that particular entity.

This is the URL schema I'm looking for:

```
https://www.example.org/[resource-name]/[non-standard-id-value]?loadById=[non-standard-id-name]
```

At the same time we need to communicate to the API consumer that the canonical resource for what RESTful is returning is different. We should do that with a `Link` header and a `rel="canonical"` in the response. That is one of the REST principles that is really useful to normalize caches –among other things–.
### Real scenario

There's a _Segment_ content type that has a unique field. We want to be able to load by that ID.

```
https://www.example.org/segments/asdf-asdf-asdf-asdf?loadById=unique
```

Returns:

```
HTTP/1.1 200 OK
Date: Mon, 22 Dec 2014 08:08:53 GMT
Expires: -1
Cache-Control: private, max-age=0
Content-Type: text/html; charset=ISO-8859-1
X-XSS-Protection: 1; mode=block
X-Frame-Options: SAMEORIGIN
Alternate-Protocol: 80:quic,p=0.02
Transfer-Encoding: chunked
Link: https://www.example.org/segments/12; rel="canonical" <----

[RESPONSE BODY]
```

### DIFF
--- a/docs/api_url.md
+++ b/docs/api_url.md
@@ -97,6 +97,35 @@ articles after a certain date:
 curl https://example.com/api/articles?filter[created][value]=1417591992&filter[created][operator]=">="
 ```
 
+## Loading by an alternate ID.
+Some times you need to load an entity by an alternate ID that is not the regular
+entity ID. All that you need to do is provide the alternate ID as the regular
+resource ID and inform that the passed in ID is not the regular entity ID but a
+different field. To do so use the `loadByFieldName` query parameter.
+
+```
+curl -H 'X-API-version: v1.5' https://www.example.org/articles/1234-abcd-5678-efg0?loadByFieldName=uuid
+```
+
+That will load the article node and output it as usual. Since every REST
+resource object has a canonical URL, but we are using a different one, a _Link_
+header will be added to the response with the canonical URL so the consumer can
+use it in future requests.
+
+```
+HTTP/1.1 200 OK
+Date: Mon, 22 Dec 2014 08:08:53 GMT
+Content-Type: application/hal+json; charset=utf-8
+...
+Link: https://www.example.org/articles/12; rel="canonical"
+
+{
+  ...
+}
+```
+
+The only requirement to use this feature is that the value for your
+`loadByFieldName` field needs to be one of your exposed fields.
 
 ## Working with authentication providers
 Restful comes with ``cookie``, ``base_auth`` (user name and password in the HTTP
--- a/docs/api_url.md
+++ b/docs/api_url.md
@@ -99,16 +99,17 @@ curl https://example.com/api/articles?filter[created][value]=1417591992&filter[c
 
 ## Loading by an alternate ID.
 Some times you need to load an entity by an alternate ID that is not the regular
-entity ID. All that you need to do is provide the alternate ID as the regular
-resource ID and inform that the passed in ID is not the regular entity ID but a
-different field. To do so use the `loadByFieldName` query parameter.
+entity ID, for example a unique ID title. All that you need to do is provide the
+alternate ID as the regular resource ID and inform that the passed in ID is not
+the regular entity ID but a different field. To do so use the `loadByFieldName`
+query parameter.
 
 ```
 curl -H 'X-API-version: v1.5' https://www.example.org/articles/1234-abcd-5678-efg0?loadByFieldName=uuid
 ```
 
 That will load the article node and output it as usual. Since every REST
-resource object has a canonical URL, but we are using a different one, a _Link_
+resource object has a canonical URL (and we are using a different one) a _Link_
 header will be added to the response with the canonical URL so the consumer can
 use it in future requests.
 
@@ -125,7 +126,9 @@ Link: https://www.example.org/articles/12; rel="canonical"
 ```
 
 The only requirement to use this feature is that the value for your
-`loadByFieldName` field needs to be one of your exposed fields.
+`loadByFieldName` field needs to be one of your exposed fields. It is also up to
+you to make sure that that field is unique. Note that in case that more tha one
+entity matches the provided ID the first record will be loaded.
 
 ## Working with authentication providers
 Restful comes with ``cookie``, ``base_auth`` (user name and password in the HTTP
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -106,6 +106,7 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
       if (in_array($param, array(
         '__application',
         'filter',
+        'loadByFieldName',
         'page',
         'q',
         'range',
@@ -314,6 +315,21 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
    */
   public function getHttpHeaders() {
     return $this->httpHeaders;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addHttpHeaders($key, $value) {
+    $headers = $this->getHttpHeaders();
+    // Add a value to the (potentially) existing header.
+    $values = array();
+    if (!empty($headers[$key])) {
+      $values[] = $headers[$key];
+    }
+    $values[] = $value;
+    $header = implode(', ', $values);
+    $this->setHttpHeaders($key, $header);
   }
 
   /**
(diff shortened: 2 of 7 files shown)


amitaibu:
:+1: 

I wonder if `loadById` is the correct term - maybe `loadByFieldName`?

amitaibu:
(as, we expose the term "field" regardless is the internal is a property or a configurable field)

e0ipso:
Only missing tests.

e0ipso:

docs/api_url.md
```diff
@@ -97,6 +97,35 @@ articles after a certain date:
 curl https://example.com/api/articles?filter[created][value]=1417591992&filter[created][operator]=">="
 ```
 
+## Loading by an alternate ID.
+Some times you need to load an entity by an alternate ID that is not the regular
+entity ID. All that you need to do is provide the alternate ID as the regular
```
amitaibu: entity ID. => entity ID, for example a unique ID title

docs/api_url.md
```diff
@@ -97,6 +97,35 @@ articles after a certain date:
...
+resource ID and inform that the passed in ID is not the regular entity ID but a
+different field. To do so use the `loadByFieldName` query parameter.
+
+```
+curl -H 'X-API-version: v1.5' https://www.example.org/articles/1234-abcd-5678-efg0?loadByFieldName=uuid
+```
+
+That will load the article node and output it as usual. Since every REST
+resource object has a canonical URL, but we are using a different one, a _Link_
```
amitaibu: but we are using a different one => `(and we are using a different one)`

docs/api_url.md
```diff
@@ -97,6 +97,35 @@ articles after a certain date:
...
+...
+Link: https://www.example.org/articles/12; rel="canonical"
+
+{
+  ...
+}
+```
+
+The only requirement to use this feature is that the value for your
+`loadByFieldName` field needs to be one of your exposed fields.
```
amitaibu: Maybe add `It is also up to you to make sure that that field is unique.`

plugins/restful/RestfulBase.php
```diff
@@ -1399,7 +1417,11 @@ public function versionedUrl($path = '', $options = array()) {
     }
 
     $base_path = variable_get('restful_hook_menu_base_path', 'api');
-    $url = $base_path . '/v' . $plugin['major_version'] . '.' . $plugin['minor_version'] . '/' . $plugin['resource'] . '/' . $path;
+    $url = $base_path;
+    if ($version_string) {
```
amitaibu, line 1421: when don't we want to add the version string? I think we always want to refer to it.

plugins/restful/RestfulEntityBase.php
```diff
@@ -86,7 +86,7 @@ public static function controllersInfo() {
         // POST
         \RestfulInterface::POST => 'createEntity',
       ),
-      '^(\d+,)*\d+$' => array(
+      '^.*$' => array(
```
amitaibu, line 89: seems we are loosing an important validation here, although I'm not sure how smart we want to be about it. Any ideas? (anyway this isn't a blocker)

plugins/restful/RestfulEntityBase.php
```diff
@@ -397,7 +398,7 @@ protected function getValueFromProperty(\EntityMetadataWrapper $wrapper, \Entity
   protected function getValueFromFieldFormatter(\EntityMetadataWrapper $wrapper, \EntityMetadataWrapper $sub_wrapper, array $info) {
     $property = $info['property'];
 
-    if (!field_info_field($property)) {
+    if (!static::propertyIsField($property)) {
```
amitaibu, line 401: over abstraction?

plugins/restful/RestfulEntityBase.php
```diff
@@ -1472,4 +1473,87 @@ public function clearResourceRenderedCacheEntity($id) {
     $this->cacheInvalidate($cid);
   }
 
+  /**
+   * Get the entity ID based on the ID provided in the request.
```
amitaibu, line 1477: maybe worth giving an example, since it might not be obvious without this PR as context. maybe:

`As any field may be used as the ID, we convert it to the numeric internal ID of the entity,`

plugins/restful/RestfulEntityBase.php
```diff
@@ -1472,4 +1473,87 @@ public function clearResourceRenderedCacheEntity($id) {
...
+   *
+   * @param mixed $id
+   *   The provided ID.
+   *
+   * @throws RestfulBadRequestException
+   *
+   * @return int
+   *   The entity ID.
+   */
+  protected function entityId($id) {
```
amitaibu: `getEntityIdByFieldId`?

plugins/restful/RestfulEntityBase.php
```diff
@@ -1472,4 +1473,87 @@ public function clearResourceRenderedCacheEntity($id) {
...
+    return $entity_id;
+  }
+
+  /**
+   * Initialize an EntityFieldQuery (or extending class).
+   *
+   * @return \EntityFieldQuery
+   *   The initialized query with the basics filled in.
+   */
+  protected function initEntityQuery() {
```
amitaibu: `getEntityFieldQuery`

plugins/restful/RestfulEntityBase.php
```diff
@@ -1472,4 +1473,87 @@ public function clearResourceRenderedCacheEntity($id) {
...
+    // Find out if the provided ID is a Drupal field or an entity property.
+    if (static::propertyIsField($public_field_info['property'])) {
+      $query->fieldCondition($public_field_info['property'], $public_field_info['column'], $id);
+    }
+    else {
+      $query->propertyCondition($public_field_info['property'], $id);
+    }
+
+    // Execute the query and gather the results.
+    $results = $query->execute();
```
amitaibu: `$results` => `$result`

plugins/restful/RestfulEntityBase.php
```diff
@@ -1472,4 +1473,87 @@ public function clearResourceRenderedCacheEntity($id) {
...
+    if (empty($results[$this->getEntityType()])) {
+      throw new RestfulUnprocessableEntityException(format_string('The entity ID @id for @resource does not exist loaded by @name.', array(
```
amitaibu: maybe, reorder the params a bit:

`The entity ID @id by @name for @resource cannot be loaded.`

plugins/restful/RestfulEntityBase.php
```diff
@@ -1472,4 +1473,87 @@ public function clearResourceRenderedCacheEntity($id) {
...
+        '@id' => $id,
+        '@resource' => $this->getPluginKey('label'),
+        '@name' => $public_property_name,
+      )));
+    }
+
+    // There is nothing that guarantees that there is only one result, since
+    // this is user input data. Return the first ID.
+    $entity_id = key($results[$this->getEntityType()]);
```
amitaibu: also we can `$query->range(0, 1)` so atleast the db query is bound to give us a single result.

tests/RestfulViewEntityTestCase.test
```diff
@@ -158,6 +158,21 @@ class RestfulViewEntityTestCase extends RestfulCurlBaseTestCase {
...
+    $text = $this->randomName();
+    $wrapper->text_single->set($text);
+    $wrapper->save();
+
+    $request = array('loadByFieldName' => 'text_single');
+    $result = $handler->get($text, $request);
+    $this->assertNotNull($result[0]);
+
+    // Make sure canonical header is added.
+    $headers = $handler->getHttpHeaders();
```
amitaibu, line 173: good stuff! :)

amitaibu:
Looks promising :) Some comments added.

plugins/restful/RestfulBase.php
e0ipso, line 1421: When having the URL in the `Link` header along with the `X-API-Version: v1.5` header. It's sloppy and error prone (what if we have `X-API-Version: v1.5` and `https://www.example.org/api/v1.4/…`?

plugins/restful/RestfulEntityBase.php
e0ipso, line 89: I don't think it's a big deal, since there's still validation when trying to load the entity.

plugins/restful/RestfulEntityBase.php
e0ipso, line 401: I feel that it makes semantic sense. A contributor or a module extending restful doesn't need to know about Field API to discern between an entity property and a field.

plugins/restful/RestfulBase.php
amitaibu, line 1421: ok, makes sense

plugins/restful/RestfulEntityBase.php
e0ipso: :+1:

e0ipso:
@amitaibu feedback implemented.

plugins/restful/RestfulEntityBase.php
amitaibu, line 401: not a big fan, but I can live with that ;)

amitaibu:
go ahead - merge it :)

e0ipso:
w00t! My first PR for a work related project :-)

amitaibu:
mazal tov dude :)

PR #335, opened 2014-12-22
Allow specifying EntityFieldQuery class.

e0ipso:
Allow providing a custom EFQ class. It will need to inherit from `EntityFieldQuery`.


e0ipso:
This PR depends on the work done in #334.

PR #336, opened 2014-12-23
Add docs about using .htaccess along with basic auth.

amitaibu:


amitaibu:
@mateu-aguilo-bosch actually a better/another solution would be providing a variable that will skip basic auth authentication. Like this we could keep having `'authentication_types' => TRUE` in the code, by have a variable that will allow enable/ disable between environments.

amitaibu:
@mateu-aguilo-bosch added variable..

e0ipso:
@amitaibu wouldn't this be achieved by not having the authentication provider enabled for the resource?

amitaibu:
This is for a scenario where DEV server is protected by .htaccess, but you want to provide basic auth on the LIVE server.

e0ipso:
Alright then. This looks good.

amitaibu:
Merged.

Issue #340, opened 2014-12-23
WSOD on filter of a property


amitaibu:
``` php
    $public_fields['user'] = array(
      'property' => 'author',
      'resource' => array(
        'user' => array(
          'name' => 'users',
        ),
      ),
    );
```

`https://example.com/api/foo?filter[user]=1`

``` javascript
{
type: "http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.5.1",
title: "SQLSTATE[42S22]: Column not found: 1054 Unknown column 'node.author' in 'where clause'",
status: 500
}
```

The reason is that in \RestfulDataProviderEFQ::queryForListFilter we should get the real property from the one defined in `$public_fields` which is in fact the $wrapper property.

In short: PR will follow :)

e0ipso:
Shouldn't this be

```
https://example.com/api/foo?filter[user]=1
```

?

amitaibu:
Sorry, yes it's `https://example.com/api/foo?filter[user]=1`. Still, without #341 - you would get an error ;)

amitaibu:
Closed in favor of #341

PR #341, opened 2014-12-23
Get the correct DB column name on filter/ sort by property

amitaibu:
#340

### DIFF
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -112,13 +112,15 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
 
     $sorts = $sorts ? $sorts : $this->defaultSortInfo();
 
-    foreach ($sorts as $sort => $direction) {
+    foreach ($sorts as $public_field_name => $direction) {
       // Determine if sorting is by field or property.
-      if (empty($public_fields[$sort]['column'])) {
-        $query->propertyOrderBy($public_fields[$sort]['property'], $direction);
+      $property_name = $public_fields[$public_field_name]['property'];
+      if (field_info_field($property_name)) {
+        $query->fieldOrderBy($public_fields[$public_field_name]['property'], $public_fields[$public_field_name]['column'], $direction);
       }
       else {
-        $query->fieldOrderBy($public_fields[$sort]['property'], $public_fields[$sort]['column'], $direction);
+        $column = $this->getColumnFromProperty($property_name);
+        $query->propertyOrderBy($column, $direction);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


amitaibu:
Next is tests

amitaibu:

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -151,16 +153,38 @@ protected function queryForListFilter(\EntityFieldQuery $query) {
...
+   * we make sure to get the real column name.
+   *
+   * @param string $property_name
+   *   The property name.
+   *
+   * @return string
+   *   The column name.
+   */
+  protected function getColumnFromProperty($property_name) {
+    $property_info = entity_get_property_info($this->getEntityType());
```
e0ipso, line 191: Does this relate to the `'method'` in the public field definition? Should we leverage that?

tests/RestfulListTestCase.test
```diff
@@ -259,28 +263,41 @@ class RestfulListTestCase extends RestfulCurlBaseTestCase {
       $roles = array_map('trim', explode(',', $item['roles']));
       $this->assertTrue(in_array($role_name, $roles), format_string('%role role found in content list.', array('%role' => $role_name)));
     }
+
+    // Sort by an entity metadata wrapper property that is different from the
+    // DB column name.
+    $handler = restful_get_restful_handler('articles', 1, 5);
+    $request = array('sort' => '-user',);
+    $result = $handler->get('', $request);
+    $this->assertEqual($result[0]['user']['id'], 3, 'List sorted by the "author" entity metadata wrapper property, which maps to the "uid" DB column name.');
```
e0ipso, line 272: This test is relying on auto-created concrete user ids. We should only care about the ordering:

``` php
$this->assertEqual($result[0]['user']['id'], max($user1->uid, $user2->uid), 'List sorted by the "author" entity metadata wrapper property, which maps to the "uid" DB column name.');
```

tests/RestfulListTestCase.test
```diff
@@ -373,6 +392,20 @@ class RestfulListTestCase extends RestfulCurlBaseTestCase {
...
+    $request = array(
+      'filter' => array(
+        'user' => array(
+          'value' => $user1->uid,
+        ),
+      ),
+    );
+
+    $result = $handler->get('', $request);
+    $this->assertEqual(count($result), 2, 'List filtered by the "author" entity metadata wrapper property, which maps to the "uid" DB column name.');
```
e0ipso, line 429: ~~Same as above:~~

e0ipso, line 429: Sorry, I mis-read the test. This is perfectly fine. :+1:

plugins/restful/RestfulDataProviderEFQ.php
amitaibu, line 191: > Does this relate to the 'method' in the public field definition?

No, we are just getting the real `schema field` from the entity metadata wrapper property definition.

amitaibu:
I'll re-roll

amitaibu:
Merged.

Issue #339, opened 2014-12-23
Wrong count with custom filter.


nedSaf:
I added a custom field filter to a resource with `getQueryForList`,
The count of the resource is not changing even when applying the filter (which is filtering correctly):

![localhost_capacity4more_www_api_activity_stream](https://cloud.githubusercontent.com/assets/7369740/5539138/771dc366-8ac6-11e4-8e10-16e2ae3ad9d3.png)

![localhost_capacity4more_www_api_activity_stream_group_24](https://cloud.githubusercontent.com/assets/7369740/5539151/94b454bc-8ac6-11e4-9c55-b0075c7ef5c4.png)

amitaibu:
@nader77 can you look at it in your contribution hour please

e0ipso:
@nader77 if you override `getQueryForList` you should also override `getTotalCount`.

Maybe a better idea would be to provide a filter query param? https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/docs/api_drupal.md#filter

amitaibu:
@nader77 The problem seems to be somewhere in `\RestfulDataProviderEFQ::getQueryCount`. We try to filter the values (`$this->queryForListFilter($query);`) but apparently it isn't working as expected.

nedSaf:
Using a filter query param.

PR #343, opened 2014-12-24
Wrong count with custom filter.

nedSaf:
#339

@amitaibu @mateu-aguilo-bosch 

I have added a custom filtering condition to the count as well.
selection_040

This solved the count issue for me.

### DIFF
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -186,6 +186,13 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
 
     $this->queryForListFilter($query);
 
+    // Add custom filters.
+    if($conditions = $this->getQueryForList()->fieldConditions) {
+      foreach($conditions as $condition) {
+        $query->fieldCondition($condition['field']['field_name'], $condition['column'], $condition['value']);
+      }
+    }
+
     return $query->count();
   }
 


plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -186,6 +186,13 @@ public function getQueryCount() {
 
     $this->queryForListFilter($query);
 
+    // Add custom filters.
+    if($conditions = $this->getQueryForList()->fieldConditions) {
```
amitaibu, line 190: this looks wrong - as I would expect `$this->queryForListFilter($query);` to already do the filtering.

amitaibu:
(btw, I have a feeling it's related to the fact that the filter is done on a field which is a `resource`)

amitaibu:
> (btw, I have a feeling it's related to the fact that the filter is done on a field which is a resource)

No, that's not the case.

nedSaf:
@amitaibu 
If i ditch the `getQueryForList` custom filter and add a public field and filter through that one like:
`?filter[group]=7`, It works without any  modifications to the restful code and with the proper count.
You think i should change my approach ?

nedSaf:
From what i saw in the code `$this->queryForListFilter($query);` Doesn't do custom filtering, Only filters according to public fields and a filter query param.

amitaibu:
Seems the mistake is that you should use `?filter[group]=24`.

PR #342, opened 2014-12-24
Add support for multiple filters on the same field.

e0ipso:
Allow things like getting all the articles created between 2 dates.

```
curl https://www.example.org/api/articles?filter[created][value][0]=12345&filter[created][operator][0]=">="&filter[created][value][1]=67890&filter[created][operator][1]="<="
```

Maybe the following is clearer.

``` php
$request['filter'] = array(
  'created' => array(
    'value' => array(12345, 67890),
    'operator' => array('>=', '<='),
    'conjuction' => 'AND',
  ),
);
```

### DIFF
--- a/docs/api_drupal.md
+++ b/docs/api_drupal.md
@@ -51,12 +51,41 @@ array(
 
 
 ### Filter
-Use the `'filter'` key to filter the list.
+Use the `'filter'` key to filter the list. You can provide as many filters as
+you need.
 
 ```php
 $handler = restful_get_restful_handler('articles');
 // Single value property.
 $request['filter'] = array('label' => 'abc');
+$result = $handler->get('', $request);
+```
+
+Additionally you can provide multiple filters for the same field. That is
+specially useful when filtering on multiple value fields. The following example
+will get all the articles with the integer multiple field that contains all 1, 3
+and 5.
+
+```php
+$handler = restful_get_restful_handler('articles');
+// Single value property.
+$request['filter'] = array('integer_multiple' => array(
+  'values' => array(1, 3, 5),
+));
+$result = $handler->get('', $request);
+```
+
+You can do more advanced filtering by providing values and operators. The
+following example will get all the articles with an integer value less than 5
+and another equal to 10.
+
+```php
+$handler = restful_get_restful_handler('articles');
+// Single value property.
+$request['filter'] = array('integer_multiple' => array(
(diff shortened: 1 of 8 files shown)


amitaibu:
For that you could use the `BETWEEN`, no?

e0ipso:
yeah… maybe that was a bad example.

``` php
$request['filter'] = array(
  'tags' => array(
    'value' => array(1, 2, 67890),
    'operator' => array('=', '=', '>'),
    'conjuction' => 'OR',
  ),
);
```

e0ipso:
Also, there could be shorthand methods like:

```
https://www.example.org/api/articles?filter[tag]=1,2,3,4 // AND
https://www.example.org/api/articles?filter[tag]=1+2+3+4 // OR
```

e0ipso:
@amitaibu this is ready for review.

docs/api_drupal.md
```diff
@@ -60,6 +61,34 @@ $request['filter'] = array('label' => 'abc');
...
+$handler = restful_get_restful_handler('articles');
+// Single value property.
+$request['filter'] = array('integer_multiple' => array(
+  'values' => array(1, 3, 5),
+));
+$result = $handler->get('', $request);
+```
+
+You can do more advanced filtering by providing values and operators. The
+following example will get all the articles with an integer value less than 5
```
amitaibu, line 79: `less than` => I think you meant `more then` ?

docs/api_url.md
```diff
@@ -97,6 +97,23 @@ articles after a certain date:
 curl https://example.com/api/articles?filter[created][value]=1417591992&filter[created][operator]=">="
 ```
 
+Additionally you can provide multiple filters for the same field. That is
```
amitaibu, line 100: the fact that we need to duplicate the same text indicates it might be divided incorrectly?

Anyway, not a blocker, but worth thinking about.

amitaibu:
You are crazy fast. Lovely PR. Added minor comments, after that it can be merged. Thanks!

docs/api_url.md
e0ipso, line 100: Yeah, I was actually thinking about dropping the docs for using the API from Drupal and just leaving a paragraph or two saying that you can use the API from Drupal and how to transform the URL params in the request array.

e0ipso:
Merged.

PR #349, opened 2014-12-27
Remove recommendation to use HAL+JSON be default

amitaibu:
I don't think it's really recommended, I personally find the simple JSON to be much better for _real life_ use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,7 @@ consuming client doesn't need to know if an entity is a node or a term, nor will
 * Resource versioning is built-in, so that resources can be reused with multiple
 consumers.  The versions are at the resource level, for more flexibility and
 control.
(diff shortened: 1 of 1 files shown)


amitaibu:
ready for review

e0ipso:
I'm not sure how I feel about recommending the proprietary SimpleJSON over the standard HAL+JSON.

I am not sure how _real life_ use cases apply in this particular case. Meaning that it's just an output format, so choosing one or the other will, probably, be based on preferences. My only argument is standard vs proprietary.

That being said, since this is a matter of _preference_ I am more than glad to let you take your preference.

e0ipso:
Merge at will.

amitaibu:
Since not you nor me are really using the HAL, I think it's ok not to request "newbies" to do it :)

Issue #351, opened 2015-01-06
Can't autocomplete users due to a missing bundle key in the user schema


RoySegall:
I have user named john. When i try to get results for users using the autocomplete in the get request i'm not getting any results. The path is: `api/v1.0/users?autocomplete[string]=joh`. I also have blog that contains the word john and when i try to autocomplete blog through `api/v1.0/blog?autocomplete[string]=joh` i get results.

The problem come from [here](https://github.com/RESTful-Drupal/restful/blob/ce9e0300c17b0aae7ee0e005ab44f829bc0667c2/plugins/restful/RestfulEntityBase.php#L223-L225). The user entity don't have a key for bundle in the users schema.

amitaibu:
@mateu-aguilo-bosch I think you had some plans for decoupling auto-complete from the RestfulEntityBase, maybe it's a good time to tackle this as-well? Can you share what you had in mind?

e0ipso:
It is possible that the autocomplete functionality can be decoupled. I don't have anything done yet, but I can look at it.

I am thinking about using a strategy pattern when interacting with different behaviours and a decorator pattern to extrapolate functionalities like render caching, rate limits, …

I think that we should contain the different URL modifiers under an interface and find a common way to interact with those too.

As you can see there are many changes. I would rather see this version fly for a while and then start to think about a version 7.x-2.x and _rewrite_ the code when the requirements have stabilized.

@amitaibu what do you think?

PR #352, opened 2015-01-06
Autocomplete should support entities without bundle key.

RoySegall:
#351

I'll start working on a test for that.

### DIFF
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -47,6 +47,18 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
   }
 
   /**
+   * Get the entity info for the current entity the endpoint handling.
+   *
+   * @param null $type
+   *   The entity type. Optional.
+   * @return array
+   *   The entity info.
+   */
+  public function getEntityInfo($type = NULL) {
+    return entity_get_info($type ? $type : $this->getEntityType());
+  }
+
+  /**
    * Constructs a RestfulDataProviderEFQ object.
    *
    * @param array $plugin
@@ -350,7 +362,7 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
     $query = $this->EFQObject();
     $entity_type = $this->getEntityType();
     $query->entityCondition('entity_type', $entity_type);
-    $entity_info = entity_get_info($entity_type);
+    $entity_info = $this->getEntityInfo();
     if ($this->getBundle() && $entity_info['entity keys']['bundle']) {
       $query->entityCondition('bundle', $this->getBundle());
     }
--- a/restful.module
+++ b/restful.module
@@ -197,16 +197,6 @@ function restful_menu() {
 }
 
 /**
- * Implements hook_entity_info_alter().
- *
- * Define the name property as the label of the user. We need this inorder for
- * users endpoint to be able to autocomplete users.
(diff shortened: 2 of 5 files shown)


RoySegall:
@amitaibu This is ready for review.

restful.module
```diff
@@ -197,6 +197,16 @@ function restful_menu() {
 }
 
 /**
+ * Implements hook_entity_info_alter().
+ *
+ * Define the name property as the label of the user. We need this inorder for
+ * users endpoint to be able to autocomplete users.
+ */
+function restful_entity_info_alter(&$entity_info) {
```
amitaibu: I don't think we would like to add that - it's too general. I think it's better to override the autocomplete function in the user resource, and making sure it queries by the `name` property.

RoySegall:
@amitaibu Can you have a second look?

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -47,6 +47,18 @@ public function getEntityType() {
...
   /**
+   * Get the entity info for the current entity the endpoint handling.
+   *
+   * @param null $type
+   *   The entity type. Optional.
+   * @return array
+   *   The entity info.
+   */
+  public function getEntityInfo($type = NULL) {
+    return entity_get_info($type ? $type : $this->getEntityType());
```
e0ipso, line 58:
``` php
return entity_get_info($type ?: $this->getEntityType());
```

e0ipso:
I'm not sure I understand the problem. Will other entities (that don't have bundles and don't have resources shipped by RESTful core) suffer from the same problem? Will they have to override getEntityInfo?

If so, we should find a way to avoid doing that since users won't know about it and will file bug reports for that. Is that possible?

plugins/restful/RestfulDataProviderEFQ.php
RoySegall, line 58: `return entity_get_info($type ?: $this->getEntityType());` won't work in php 5.3

RoySegall:
Entities without bundles are vocabulary and user. User is the only entity that don't define the label key and there are three ways i can think that will solved this:
1. Patch Drupal 7 but i don't think any one will merge this one in.
2. Use hook_entity_info_alter and just fix the user entity as i deed at the beginning
3. Use the current solution - the entity get info method. This look kind of a hack but in the long term if some one want to change the label key for the automplete he can do this in the controller.

plugins/restful/RestfulDataProviderEFQ.php
e0ipso, line 58: Oh! I thought it was available for php5.3

http://php.net/manual/en/language.operators.comparison.php#114872

e0ipso:
> Entities without bundles are vocabulary and user

There can be custom defined entities that don't have bundles, right?

> User is the only entity that don't define the label key

Is there a solution to fix this problem for all the entities that don't define the label key instead of only fixing it for the user entity?

RoySegall:
> There can be custom defined entities that don't have bundles, right?

Sure. For non fieldable entities a bundle is not mandatory.

> Is there a solution to fix this problem for all the entities that don't define the label key instead of only fixing it for the user entity?

I can't think on generic way to solve this. For example message don't define a label for the message entity. I can't see an autocomplete for messages but this is one example for the problem that maybe controllers would like to override the entity info.

e0ipso:
Sorry it took so long to merge. It's merged now.

Thanks!!!!

PR #356, opened 2015-01-08
196 data providers

raphaeltraviss:
Added data provider for variables, and switched dbQuery to use drupal_write_record().


modules/restful_example/plugins/restful/db_query/watchdog/1.0/RestfulWatchdogResource.class.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+/**
+ * @file
+ * Contains RestfulEntityBase.
+ */
+
+/**
+ * An abstract implementation of RestfulEntityInterface.
```
e0ipso: This class is not abstract ;-)

modules/restful_example/plugins/restful/db_query/watchdog/1.0/watchdog.inc
```diff
@@ -0,0 +1,15 @@
...
+  'resource' => 'watchdog',
+  'name' => 'watchdog',
+  'data_provider_options' => array(
+    'table_name' => 'watchdog',
+    'id_column' => 'wid',
+  ),
+  'description' => t('Expose watchdog entries to the REST API.'),
+  'class' => 'RestfulWatchdogResource',
+  'authentication_types' => TRUE,
+  'authentication_optional' => TRUE,
```
e0ipso: Let's remove the `authentication_optional`.

e0ipso: ~~Actually we should add an access check to this endpoint, since it can contain sensitive information. We wouldn't want a security issue because someone left `restful_example` enabled.~~

I see that you already did that :-)

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -99,7 +113,7 @@ public function __construct(array $plugin, \RestfulAuthenticationManager $auth_m
 
     $this->tableName = $options['table_name'];
     $this->idColumn = $options['id_column'];
-    $this->primary = empty($plugin['primary']) ? NULL : $this->primary = $plugin['primary'];
+    $this->primary = !empty($plugin['primary']) ? $options['primary'] : $options['id_column'];
```
e0ipso: What's this for?

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -341,37 +355,35 @@ protected function replace($id) {
...
+
+    $record = array();
     foreach ($public_fields as $public_property => $info) {
       // If this is the primary field, skip.
       if ($this->isPrimaryField($info['property'])) {
         continue;
       }
-      // Check if the public property is set in the payload.
-      if (!isset($request[$public_property])) {
-        if ($full_replace) {
```
e0ipso, line 370: Are we losing this functionality?

`$full_replace` is what separates a `PUT` request from a `PATCH` request.

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -341,37 +355,35 @@ protected function replace($id) {
...
     }
-    if (empty($fields)) {
+    if (empty($record)) {
       return $this->view($id);
     }
 
-    // Once the update array is built, execute the query.
-    $query->fields($fields)->execute();
+    // Add the id column values into the record.
+    foreach ($this->getIdColumn() as $index => $column) {
```
e0ipso, line 408: ~~Throw an exception if `$this->getIdColumn()` is empty.~~

Wait, that won't happen.

mateu-aguilo-bosch: Since it would become an insert in an update function call.

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -341,37 +355,35 @@ protected function replace($id) {
...
+      $record[$column] = current($this->getColumnFromIds(array($id), $index));
+    }
+
+    // Once the record is built, write it.
+    if (!drupal_write_record($this->getTableName(), $record, $id_columns)) {
+      throw new \RestfulServiceUnavailable('Record could not be updated to the database.');
```
e0ipso, line 414: :+1:

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -379,51 +391,38 @@ public function update($id, $full_replace = FALSE) {
       'cl' => implode(',', $this->getIdColumn()),
       'id' => $id,
     ));
-    return $this->view($id, TRUE);
+
+    // @todo: do we need to re-form $id in case some of the values have changed?
```
e0ipso: I don't think we need. What scenario are you thinking about?

e0ipso: Either way, we need to create a follow up issue in GitHub and remove the `@todo` in code or just remove the comment.

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -379,51 +391,38 @@ public function update($id, $full_replace = FALSE) {
...
+      // Handle multiple id columns.
+      $id_values = array();
+      foreach ($id_columns as $id_column) {
+        $id_values[$id_column] = $record[$id_column];
+      }
+      $id = implode(self::COLUMN_IDS_SEPARATOR, $id_values);
 
-    // Once the update array is built, execute the query.
-    if ($id = $query->fields($fields)->execute()) {
       return $this->view($id, TRUE);
```
e0ipso: Do we want to just: `$this->view($id);`?

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+   * @param array $plugin
+   *   Plugin definition.
+   * @param RestfulAuthenticationManager $auth_manager
+   *   (optional) Injected authentication manager.
+   * @param DrupalCacheInterface $cache_controller
+   *   (optional) Injected cache backend.
+   * @param string $language
+   *   (optional) The language to return items in.
+   */
+  public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL, $language = NULL) {
```
e0ipso: If we're just calling the parent, then we don't need to define the constructor.

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+    parent::__construct($plugin, $auth_manager, $cache_controller, $language);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function controllersInfo() {
```
e0ipso: Is there any need to re-define the controller names? Can't we just use the usual ones? I think it helps with setting expectation on what's in a data provider class.

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+    // Get the sorting options from the request object.
+    $sorts = $this->parseRequestForListSort();
+
+    $sorts = $sorts ? $sorts : $this->defaultSortInfo();
+
+    foreach ($sorts as $public_field_name => $direction) {
+      if (isset($public_fields[$public_field_name]['property'])) {
+        $property_name = $public_fields[$public_field_name]['property'];
+        if ($property_name == 'name') {
+          uksort($variables, array(__CLASS__, 'nameSort'));
```
e0ipso: I didn't know that you could use the keyword `__CLASS__` as a substitute for `$this` in the callable context :-)

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Sort the list of variables by name.
+   *
+   * @param string $a
+   *   The first variable name for the comparison.
+   * @param string $b
+   *   The second variable name for the comparison.
+   */
+  public function nameSort($a, $b) {
```
e0ipso: If this is not using the class' state, can we make it an static method?

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+   * If you are comparing variables with the same value, note that the PHP
+   * sorting algorithm is not stable, and their order may be changed.
+   *
+   * @param mixed $a
+   *   The first data structure for the comparison.
+   * @param mixed $b
+   *   The second data structure for the comparison.
+   */
+  public function valueSort($a, $b) {
+    return;
```
e0ipso: Are we not sorting?

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+
+    $sorts = $sorts ? $sorts : $this->defaultSortInfo();
+
+    foreach ($sorts as $public_field_name => $direction) {
+      if (isset($public_fields[$public_field_name]['property'])) {
+        $property_name = $public_fields[$public_field_name]['property'];
+        if ($property_name == 'name') {
+          uksort($variables, array(__CLASS__, 'nameSort'));
+        }
+        elseif ($property_name == 'value') {
```
e0ipso: Instead of an `if/else`, would it be possible to have a generic handle of this that would not need to separate by `$property_name`?

Maybe having a generic `$this->stringCompare` method?

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+  protected function applyListPagination(array &$variables) {
+    list($offset, $range) = $this->parseRequestForListPagination();
+    $variables = array_slice($variables, $offset, $range);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getTotalCount() {
+    return array('restful_count' => count($this->getVariablesForList()));
```
e0ipso: The count should not include pagination. This should be something like `count($GLOBALS['conf'])`.

Also, I'm not sure I get why you are returning an array. This is what I see for other data providers:

``` php
  /**
   * Helper method to get the total count of entities that match certain
   * request.
   *
   * @return int
   *   The total number of results without including pagination.
   */
  public function getTotalCount();
```

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
+<?php
+
+/**
+ * @file
+ * Contains \RestfulDataProviderDbQuery
+ */
+
+abstract class RestfulDataProviderVariable extends \RestfulBase implements \RestfulDataProviderInterface {
```
e0ipso, line 8: Maybe we need to create a `RestfulDataProviderVariableInterface`. Just like we have `RestfulDataProviderCToolsPluginsInterface`.

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewMultiple(array $names) {
+    if (empty($names)) {
+      return array();
+    }
+    $variables = array_keys($this->getVariablesForList());
```
e0ipso: `getVariablesForList` will return the paginated list, we should access the complete list of variables. `global $conf` instead?

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+    $output = array();
+
+    // Build output according to filtered/sorted list of variables.
+    foreach ($variables as $variable) {
```
e0ipso: Can we revers the order of `foreach`? There are potentially lots and lots of variables. Possibly something like:

``` php
global $conf;
foreach ($names as $name) {
  if (!empty($conf[$name])) {
    $output[] = $this->view($variable);
  }
}
```

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+    $return = array();
+    foreach ($this->getPublicFields() as $public_field_name => $info) {
+      if ($info['property'] == 'name') {
+        $public_field_value = $name;
+      }
+      elseif ($info['property'] == 'value') {
+        $public_field_value = variable_get($name);
+      }
+      else {
+        continue;
```
e0ipso: Throw an exception explaining that the only possible properties for the variable resource are: `'name'` and `'value'`. Only if the `$info['property']` is populated.

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+      }
+
+      // Modify the public field value using a callback, if supplied.
+      $public_field_value = $info['callback'] ? static::executeCallback($info['callback'], array($value)) : $public_field_value;
```
e0ipso: The `$info['callback']` provides an way to wire a callback to your public field. This should be something like:

``` php
      if ($info['callback']) {
        $public_field_value = static::executeCallback($info['callback'], array($name));
      }
```

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  /**
+   * Alias for $this->setVariable().
+   *
+   * @param string $name
+   *  The name of the variable to set a value for.
+   */
+  public function update($name, $full_update = TRUE) {
+    $this->setVariable($name);
```
e0ipso: We need to use the `$full_update` in order to tell `PUT` and `PATCH` methods apart.

With the `PUT` method if we don't provide a value, then NULL is assumed to be the value to set. With `PATCH` if we don't provide a value then the current value is not touched.

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+    foreach ($public_fields as $public_property => $info) {
+      if ($info['property'] == 'name' && isset($request[$public_property])) {
+        $name = $request[$public_property];
+      }
+      elseif ($info['property'] == 'value' && isset($request[$public_property])) {
+        $value = $request[$public_property];
+      }
+    }
+    if (isset($name)) {
+      variable_set($name, $value);
```
e0ipso: It's possible that there is no `$value` set (see above for `PUT` vs `PATCH`)

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+      // Clear the rendered cache before calling the view method.
+      $this->clearRenderedCache(array(
+        'tb' => 'variable',
+        'id' => $name,
+      ));
+
+      return $this->view($name);
+    }
+    else {
+      throw new Exception('No name property supplied');
```
e0ipso: The RESTful module always throws its own exceptions so we can do fancy things with them later. In this case you should throw a `RestfulBadRequestException`. Although in this case we don't need it since we will always have a name (because it's the ID for the update method).

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+   * Sets a variable value.
+   *
+   * This method is used for both 'create' and 'update' contexts.
+   *
+   * @param string $name
+   *  The name of the variable to set a value for.
+   *
+   * @todo: get $name from the request: make it optional for create()/update().
+   */
+  public function setVariable() {
```
e0ipso: `public function setVariable($name) {`

plugins/restful/RestfulDataProviderVariable.php
```diff
@@ -0,0 +1,306 @@
...
+
+  /**
+   * Sets a variable value.
+   *
+   * This method is used for both 'create' and 'update' contexts.
+   *
+   * @param string $name
+   *  The name of the variable to set a value for.
+   *
+   * @todo: get $name from the request: make it optional for create()/update().
```
e0ipso: `$name` is ignored in the payload for the update (because it's the ID of the resource). Although it should be mandatory on create.

Please remove the `@todo` comment.

tests/RestfulDbQueryTestCase.test
```diff
@@ -68,19 +65,20 @@ class RestfulDbQueryTestCase extends DrupalWebTestCase {
       'integer' => $mock_data['int_field'],
       'string' => $this->randomName(),
     );
+
     $handler->put($id, $mock_data2);
     $result = $handler->get($id);
-    // Remove nullified fields.
-    $result = array_filter($result[0]);
-    $this->assertEqual($result, $mock_data2, 'The record was updated with PUT successfully.');
```
mateu-aguilo-bosch: Why do we not need this anymore?

tests/modules/restful_test/restful_test.install
```diff
@@ -54,5 +54,37 @@ function restful_test_schema() {
     'primary key' => array('pid'),
   );
 
+  // Defining table via hook_install() due to drupal_write_record().
```
e0ipso, line 57: Does this mean that if we are not declaring the table via `hook_schema` we cannot use `drupal_write_record`?

tests/RestfulDbQueryTestCase.test
```diff
@@ -68,19 +65,20 @@ class RestfulDbQueryTestCase extends DrupalWebTestCase {
...
+    // Make mock data resemble the intended response.
+    $mock_data2['serialized'] = $mock_data['serialized_field'];
```
e0ipso: Actually this is the same `PUT` and `PATCH` situation as before. If we do not provide a _serialized_ and we call `PUT` that means that the _serialized_ should be nullified.

e0ipso: Please, see that tests reflect that.

tests/RestfulVariableTestCase.test
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+class RestfulVariableTestCase extends DrupalWebTestCase {
+
+  /**
+   * The name of the test table.
+   *
+   * @var string
+   */
+  protected $tableName = 'restful_test_variable';
```
e0ipso, line 15: Copy & paste leftover?

tests/RestfulVariableTestCase.test
```diff
@@ -0,0 +1,127 @@
...
+   */
+  function setUp() {
+    parent::setUp('restful_example');
+  }
+
+  /**
+   * Test authenticating a user.
+   */
+  function testCrudOperations() {
+    // Populate the table with some variables.
```
e0ipso: `// Populate with some variables.`

tests/RestfulVariableTestCase.test
```diff
@@ -0,0 +1,127 @@
...
+    $this->assertTrue(variable_get('variable_5'), 'The variables have been set.');
+
+    // Get the handler.
+    $handler = restful_get_restful_handler('variables');
+
+    // Testing read
+    $results = $handler->get('variable_5');
+    $result = $results[0];
+    $this->assertEqual($result['variable_name'], 'variable_5', 'The variable name was successfully retrieved.');
+    $this->assertNotNull($result['variable_value'], 'The variable value was successfully retrieved.');
```
e0ipso: We should store the random number when it gets generated and compared here:

``` php
$this->assertEqual($result['variable_value'], $numbers[5], 'The variable value was successfully retrieved.');
```

tests/RestfulVariableTestCase.test
```diff
@@ -0,0 +1,127 @@
...
+    $inResults = FALSE;
+    foreach ($results as $result) {
+      if ($result['variable_name'] == 'variable_5') {
+        $inResults = TRUE;
+      }
+    }
+    $this->assertTrue($inResults, 'All the content listed successfully.');
+
+    // Testing sort.
+    variable_set('zzz', array('some_property' => $this->randomName()));
```
e0ipso: Imagine a site that has a variable named `zzz_counter` (for the imaginary zzz module). This can lead to false positives. We should get all the variable names, then get the last (in strcmp order) and generate one that would compare after that.

tests/RestfulVariableTestCase.test
```diff
@@ -0,0 +1,127 @@
...
+    // Testing create.
+    $request = array(
+      'variable_name' => 'created_variable',
+      'variable_value' => $this->randomName(),
+    );
+    $handler->post('', $request);
+    $results = $handler->get('created_variable');
+    $initial_value = $results[0]['variable_value'];
+    $this->assertEqual($results[0]['variable_name'], 'created_variable', 'The variable was created.');
+    $this->assertTrue($initial_value, 'The created variable value is present.');
```
e0ipso:
``` php
$this->assertEqual($initial_value, $request['variable_value'], 'The variable was successfully created.');
```

e0ipso:
@rfx1007 this is wonderful work!

Thanks a lot for this contribution, I can only imagine all the effort you put for this pull request. I made a thorough peer review out of respect for your awesome work.

You'll see that there are some questions, some suggestions and some clarifications. If you don't feel like implementing those, let me know and I'll pop in a PR on top of your work.

respect

raphaeltraviss:
Thanks for the in-depth review; I'll take care of the remaining issues and complete the pull request.

plugins/restful/RestfulDataProviderDbQuery.php
raphaeltraviss, line 370: My thinking here was that `drupal_write_record()` always performs a partial update.  But, come to think of it, if we are doing `$full_replace`, we may be able to feed it empty values for the non-supplied fields to get it to work for a `PUT` context.  I'll look into this.

plugins/restful/RestfulDataProviderDbQuery.php
raphaeltraviss: This was a left over comment from before I understood how the id processing worked; will be deleted.

plugins/restful/RestfulDataProviderVariable.php
raphaeltraviss: By default, we don't do any sorting on the variable values, since their data structures can be completely different.  This method exists as a placeholder, to be overridden by the plugin author.  If they have some specific variables that they know the data structure of, and they want to sort by, then they can override this method in their plugin.  Maybe I should replace the algorithm comment with something more useful?

plugins/restful/RestfulDataProviderVariable.php
raphaeltraviss: Question: is `viewMultiple()` expected to obey the sort order of the passed-in `$names`, or the sort defined by the plugin?

plugins/restful/RestfulDataProviderVariable.php
raphaeltraviss: That is my attempt at sorting logic; I looped through `$variables` because they were in the correct sort order from `getVariablesForList()`.  So, this depends on whether we want this method to sort according to the plugin definition, or the passed-in `$names`.

plugins/restful/RestfulDataProviderVariable.php
raphaeltraviss: Another left-over, will remove.

tests/modules/restful_test/restful_test.install
raphaeltraviss, line 57: Correct.  `drupal_write_record()` uses `drupal_get_schema()` to find the names of the fields for the supplied table, and if the table is not defined in the schema, it will return `FALSE`.

amitaibu, line 57: :+1:

plugins/restful/RestfulDataProviderDbQuery.php
e0ipso: Maybe you could drop a comment on how it works with your own words, it will help the next time we revisit this.

plugins/restful/RestfulDataProviderVariable.php
e0ipso: :+1: to adding a `@code` implementation example in the docblock.

plugins/restful/RestfulDataProviderVariable.php
e0ipso: There's no guarantee that the output order is the same of the order of the ids. This means that the output order is the one defined by the plugin.

plugins/restful/RestfulDataProviderVariable.php
e0ipso: Can we build the array and then pass the results through `uasort`?

e0ipso:
@rfx1007 You are welcome!

plugins/restful/RestfulDataProviderDbQuery.php
raphaeltraviss: This will set `$primary` from the id column, in case the `'primary'` key is omitted from the plugin definition.  This is used in `update()` to skip updating the primary and id fields.

plugins/restful/RestfulDataProviderDbQuery.php
raphaeltraviss, line 370: Yes, this works, adding in next commit.

plugins/restful/RestfulDataProviderVariable.php
raphaeltraviss: Yes, it could go either way.  But I think the tie-breaker is that fact that you need to know the properties in order to know if you should run the sort function on the keys, or the values of the array, i.e. `uksort()` vs `uasort()`.  And, since we need to know which property we're sorting, that allows us to break them out into two separate methods: `nameSort()`, which is essentially string comparison, and `valueSort()`, which compares data structures in some specific way.

plugins/restful/RestfulDataProviderVariable.php
raphaeltraviss: My thinking was that we needed the exception for `POST` requests, where `name` might not have been sent in the body, since we'd have nothing to create and nothing to view.

EDIT: ah, ok, `create` contexts set the name from the payload, `update` and `replace` provide it as part of the method call.

rfx1007:
I've implemented all of the items from the code review, but now, many of the tests are failing: I've got about 75 exceptions and 15 failures in Node Create, CSRF, and others.  Many of the exceptions are from `entity_metadata_no_hook_node_access()`.  I've tried using entity API dev version with the two patches applied, and entity API from git with the one patch it still needs, same result.  Any hints as to what might be wrong?

raphaeltraviss:
Test lights are finally green!
- removed variable value sort, since it was a weird concept to begin with, and could not possibly be supplied via URL
- the variable array is now indexed, with each element containing the property, similar to how the other data providers work.  Reworked view/viewMultiple
- integrated passthrough commit with dbquery provider and got the tests working

modules/restful_token_auth/tests/RestfulTokenAuthenticationTestCase.test
```diff
@@ -11,13 +11,12 @@ class RestfulTokenAuthenticationTestCase extends DrupalWebTestCase {
     return array(
       'name' => 'Token Authentication',
       'description' => 'Test the request authentication with a token.',
-      'group' => 'Restful',
+      'group' => 'RESTful',
```
e0ipso, line 14: :+1:

tests/RestfulVariableTestCase.test
e0ipso, line 15: I can remove this.

e0ipso:
Merged!

Many thanks for this great contribution. I'll make sure to mention your effort in the upcoming release notes.

amitaibu:
Thanks @rfx1007 !

raphaeltraviss:
Whew!  Finally bagged this one!  Time for a drink.

PR #359, opened 2015-01-11
2.x plug authentication providers

e0ipso:
This PR is the first of an attempt to mock the possible design of 2.x. 7.x-2.x will be as similar as it can be to a D8 module, this way module upgrade will be a bliss.

This PR will:
- Start using D8 plugins (via the plug module) for the authentication providers.
- Start using D8 plugin managers (`AuthenticationPluginManager`) instead of the current custom code.
- Will use composition to create an `AuthenticationManager` that substitutes `RestfulAuthenticationManager`.
- Will use YAML based plugin definitions (possibly changeable to annotation based plugin definitions).
- Will use a D8 plugin collection to hold all the different authentication plugins.

At this point the module is still functional, but tests are broken.


src/Authentication/AuthenticationManager.php
```diff
@@ -26,48 +37,50 @@ class RestfulAuthenticationManager extends \ArrayObject {
...
   /**
-   * Set the authentications' "optional" flag.
-   *
-   * @param boolean $is_optional
-   *   Determines if the authentication is optional.
+   * @param AuthenticationPluginManager $manager
+   *   The authentication plugin manager.
+   */
+  public function __construct(AuthenticationPluginManager $manager = NULL) {
+    $manager = $manager ?: AuthenticationPluginManager::create();
```
amitaibu, line 44: lets still keep this php 5.3 `$manager ? $manager : AuthenticationPluginManager::create();`

amitaibu:
Looks like a great start.

> Will use YAML based plugin definitions

Why not go only with annotations?

e0ipso:
> Why not go only with annotations?

I was trying to keep the UX as similar as possible to 1.x, but I was certainly tempted.

src/Authentication/AuthenticationManager.php
e0ipso, line 44: The plugin system in D8 requires traits, which means that we need php 5.4.

e0ipso:
Closing, this should be against 7.x-2.x

amitaibu:
> UX as similar as possible to 1.x

I think it's fine to go all the way with annotations

src/Authentication/AuthenticationManager.php
amitaibu, line 44: ok, I can live with that :)

PR #360, opened 2015-01-11
2.x plug authentication providers

e0ipso:
This PR is the first of an attempt to mock the possible design of 2.x. 7.x-2.x will be as similar as it can be to a D8 module, this way module upgrade will be a bliss.

This PR will:
- Start using D8 plugins (via the plug module) for the authentication providers.
- Start using D8 plugin managers (`AuthenticationPluginManager`) instead of the current custom code.
- Will use composition to create an `AuthenticationManager` that substitutes `RestfulAuthenticationManager`.
- Will use YAML based plugin definitions (possibly changeable to annotation based plugin definitions).
- Will use a D8 plugin collection to hold all the different authentication plugins.

At this point the module is still functional, but tests are broken.


e0ipso:
Merging into 2.x.

PR #363, opened 2015-01-11
Add a 'passthrough' option for public fields

amitaibu:


plugins/restful/RestfulEntityBase.php
```diff
@@ -307,6 +313,11 @@ public function viewEntity($id) {
         $value = static::executeCallback($info['callback'], array($wrapper));
       }
       else {
+        if (!$info['property'] && $info['create_or_update_passthrough']) {
```
e0ipso: What about view modes, callbacks, … Do we need if/else cases for that?

plugins/restful/RestfulEntityBase.php
```diff
@@ -1220,24 +1241,15 @@ public function publicFieldsInfo() {
...
+  protected function addDefaultValuesToPublicFields(array $public_fields = array()) {
+    $public_fields = parent::addDefaultValuesToPublicFields($public_fields);
     // Set defaults values.
     foreach (array_keys($public_fields) as $key) {
-      // Set default values.
+      // Set default values specific for entities.
       $info = &$public_fields[$key];
       $info += array(
         'access_callbacks' => array(),
-        'callback' => FALSE,
```
e0ipso, line 1238: why are we dropping this?

amitaibu, line 1238: not dropping, it's moving to the parent class. I'm trying to refactor a bit getPublicFields/ setPublicFields -- so it always adds defaults values to the configuration array (for easier testing).

plugins/restful/RestfulEntityBase.php
amitaibu: yes, I'll cover all the places it's needed.

plugins/restful/RestfulEntityBase.php
e0ipso, line 1238: gotcha

amitaibu:
Update: currently the update DB provider is failing, and after that it should be ready

(btw, The amount of code duplication shows some refactoring will be needed.)

amitaibu:
That was harder then anticipated ;)

Ready for review.

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -99,7 +99,7 @@ public function __construct(array $plugin, \RestfulAuthenticationManager $auth_m
 
     $this->tableName = $options['table_name'];
     $this->idColumn = $options['id_column'];
-    $this->primary = empty($plugin['primary']) ? NULL : $this->primary = $plugin['primary'];
+    $this->primary = empty($plugin['data_provider_options']['primary']) ? NULL : $this->primary = $plugin['data_provider_options']['primary'];
```
e0ipso, line 102: Is this an unrelated bugfix?

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -349,25 +349,43 @@ public function update($id, $full_replace = FALSE) {
...
+    $save = FALSE;
+    $original_request = $request;
+
     $public_fields = $this->getPublicFields();
     $fields = array();
-    foreach ($public_fields as $public_property => $info) {
+    foreach ($public_fields as $public_field_name => $info) {
+      if (!empty($info['create_or_update_passthrough'])) {
+        // Allow passing the value in the request.
+        if (!empty($info['create_or_update_passthrough_required']) && !isset($request[$public_field_name])) {
```
e0ipso: About `'create_or_update_passthrough_required'`. We are deferring validation (for entities) to _Entity Validation_. So I'm not sure why we need to validate input data for the _passthrough_ fields.

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -349,25 +349,43 @@ public function update($id, $full_replace = FALSE) {
...
         if ($full_replace) {
           $fields[$info['property']] = NULL;
         }
       }
       else {
-        $fields[$info['property']] = $request[$public_property];
+        $fields[$info['property']] = $request[$public_field_name];
       }
+
+      unset($original_request[$public_field_name]);
```
e0ipso, line 378: I don't follow what is `$original_request` for.

plugins/restful/RestfulDataProviderDbQuery.php
e0ipso: This applies to later cases too.

plugins/restful/RestfulDataProviderDbQuery.php
amitaibu, line 102: yes - once I've added the `id` to the public field, any update would fail without this fix.

plugins/restful/RestfulDataProviderDbQuery.php
amitaibu: > We are deferring validation (for entities) to Entity Validation. So I'm not sure why we need to validate input data for the passthrough fields.

You mean you don't like the `required` property? In that case where do you think we should we move it?

plugins/restful/RestfulDataProviderDbQuery.php
amitaibu, line 378: this is a a bad copy/ paste - in the RestfulEntityBase we use $original_request to make sure all properties were mapped properly before saving them. I'll fix this

plugins/restful/RestfulDataProviderDbQuery.php
amitaibu: maybe we should have a `dataValidate()` instead of `entityValidate()` - so it applies also to non-entities?

amitaibu: I mean, we should defiantly have a `dataValidate()` instead of `entityValidate()` :)

But the required can indeed by a regular property.

e0ipso: Right now there is no place in RESTful where the user specifies if the data being submitted is valid to be saved or not.

The only way of doing so is with entities via the usage of _Entity Validator_. I like this pattern, and the fact that RESTful is not opinionated about data validation (beyond calling the _Entity Validator_).

Additionally I think that @RoySegall was working on getting generic data validations for _Entity Validator_, so maybe the required passthrough could be deferred to that.

plugins/restful/RestfulDataProviderDbQuery.php
e0ipso, line 378: :+1:

plugins/restful/RestfulDataProviderDbQuery.php
RoySegall: @mateu-aguilo-bosch @amitaibu The entity validator can handle schema validation and generic objects. You can look [here](https://github.com/Gizra/entity_validator/blob/7.x-1.x/modules/entity_validator_example/plugins/validator/entity_validator_example/EntityValidatorExampleObjectValidator.class.php)

e0ipso: That's fantastic @RoySegall! @amitaibu that means that we can ignore `create_or_update_passthrough_required` and call, as you suggest `dataValidate()`. In turn, `dataValidate()` could use _Entity Validator_ (with the example provided by @RoySegall).

amitaibu:
I'll remove the `passthourgh_required` property as indeed validation can be done else where

amitaibu:
Comments fixed, ready for review

amitaibu:
@mateu-aguilo-bosch ^^

e0ipso:
Looking good!

e0ipso:
Merged.

PR #362, opened 2015-01-11
2.x token auth plug

e0ipso:
Depends on https://github.com/RESTful-Drupal/restful/pull/360.

Converts the token auth provider to the D8 plugin system.

PR #364, opened 2015-01-11
2.x Rate Limits

e0ipso:
Depends on #362.

This PR migrates the Rate Limit plugins and the Rate Limit Manager to the D8 plugin system.

Note that this time I created the plugins using named annotations, so you can see the difference between declaring an annotated plugin or a yaml based plugin. I think I like annotations better, but it's less consistent with the current DX… I'm unsure about this. We should decide about one type of plugin discovery and stick with it. Opinions?

This is [the new code](https://github.com/mateu-aguilo-bosch/restful/compare/mateu-aguilo-bosch:2.x-token-auth-plug...mateu-aguilo-bosch:2.x-rate-limits?expand=1) since the last PR.


amitaibu:
>  but it's less consistent with the current DX… I'm unsure about this

For me annotations make more sense, since it's happening on the code itself. A YAML is like staying with CTools' `inc` file

e0ipso:
Merging into 2.x.

PR #365, opened 2015-01-12
WIP: 2.x Fix tests

e0ipso:


restful.info
```diff
@@ -1,8 +1,9 @@
 name = RESTful
 description = Turn Drupal to a RESTful server, following best practices.
 core = 7.x
-dependencies[] = ctools
+php = 5.4
 dependencies[] = entity
+dependencies[] = plug
```
amitaibu, line 6: nice :)

src/RateLimit/RateLimitManager.php
```diff
@@ -41,29 +55,32 @@ public function getAccount() {
...
+    foreach ($plugin_options as $plugin_id => $rate_options) {
+      // Set the instance id to articles::request and specify the plugin id.
+      $instance_id = $resource->getResourceName() . '::' . $plugin_id;
+      $options[$instance_id] = array(
+        'id' => $plugin_id,
+        'resource' => $resource,
+      );
+      $options[$instance_id] += $rate_options;
+    }
+    $this->plugins = new RateLimitPluginCollection($manager, $options);
```
amitaibu, line 83: @mateu-aguilo-bosch can you explain this part?

e0ipso, line 83: This is the D8 recommended way of dealing with arrays of plugins.

From the code documents:

> A plugin collection is used to contain plugins that will be lazily instantiated. The configurations of each potential plugin are passed in, and the configuration key containing the plugin ID is specified by self::$pluginKey.

e0ipso, line 83: By lazily instantiated, it means that the actual plugin objects will be instantiated when looping through `$this->plugins` for the first time.

e0ipso, line 83: Also, see that in this case (but not in auth providers) we can potentially have multiple plugin instances of the same plugin definition at the same time. Hence the fact that the passed `$plugin_id` and the `$instance_id` are different.

src/Plugin/AuthenticationPluginManager.php
```diff
@@ -0,0 +1,61 @@
...
+   * Constructs AuthenticationPluginManager.
+   *
+   * @param \Traversable $namespaces
+   *   An object that implements \Traversable which contains the root paths
+   *   keyed by the corresponding namespace to look for plugin implementations.
+   * @param \DrupalCacheInterface $cache_backend
+   *   Cache backend instance to use.
+   */
+  public function __construct(\Traversable $namespaces, \DrupalCacheInterface $cache_backend) {
+    $this->discovery = new YamlDiscovery('authentication', Module::getDirectories());
```
amitaibu, line 42: are we going with YAML, or it's just a leftover?

e0ipso, line 42: I wanted to have an implementation of both so we could decide if we wanted to use YAML or annotations.

It seems clear that we both prefer annotations, so this will need to get converted. Luckily enough the only part that changes is the discovery :-)

e0ipso:
This should be against 7.x-2.x branch.

PR #367, opened 2015-01-14
355 Bad HAL:JSON

adaddinsane:
The initial issue was that sub-resources were not being output in HAL:JSON at all, then it was discovered (when fixed) they were always output (as embedded resources) at the outermost level and not associated with the resource to which they belonged.

The fix ensures that sub-resources are embedded with their parent resource and nested correctly.

### DIFF
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -6,6 +6,18 @@
  */
 
 class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulFormatterInterface {
+
+  /**
+   * Determine whether this is an associative array.
+   *
+   * @param array $array
+   *
+   * @return boolean
+   */
+  static function is_assoc(array $array) {
+    return (bool)count(array_filter(array_keys($array), 'is_string'));
+  }
+
   /**
    * Content Type
    *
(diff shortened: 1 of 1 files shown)


plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
```diff
@@ -6,6 +6,18 @@
...
 class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulFormatterInterface {
+
+  /**
+   * Determine whether this is an associative array.
+   *
+   * @param array $array
+   *
+   * @return boolean
+   */
+  static function is_assoc(array $array) {
```
e0ipso, line 17: We already have `\RestfulBase:isArrayNumeric`, maybe you can use that instead.

e0ipso:
@adaddinsane it seems that the `View HAL+JSON` test is failing:

```
Test RestfulHalJsonTestCase->testHalEmbeddedResources() failed: Undefined index: _embedded
```

thebruce:
Yeah, i see what is happening here, we want the individual resources on listings to have their own embeds, and in that context "_embedded" is at the right level, the root of each item, not at the resource root.

However, the test is looking at an individual item, and the above nesting level is carrying over there on an individual item, nested within the main object but it should be at the resource root with _links on an individual resource.

I'll try and look at this today.

thebruce:
I think this can be closed this as it was solved through #422

Issue #378, opened 2015-01-18
Enable support for sub-property setter


bricel:
In case of fields with sub properties the setter fail on create and update the values:
example:

```
$public_fields['period'] = array(
      'property' => 'field_work_length',
      'sub_property' => 'period',
    );
```

In such case when saving the entity wrapper should use the following format:

```
$wrapper->{$property_name}->{$sub_property_name}->set($field_value);
```

e0ipso:
Closing in favor of #379.

PR #379, opened 2015-01-18
Fix sub entity setter. 

bricel:
From: #378.

In case of fields with sub properties the setter fail on create and update the values:
example:

```
$public_fields['period'] = array(
      'property' => 'field_work_length',
      'sub_property' => 'period',
    );
```

In such case when saving the entity wrapper should use the following format:

```
$wrapper->{$property_name}->{$sub_property_name}->set($field_value);
```

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -640,6 +640,11 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
       }
 
       $property_name = $info['property'];
+      $sub_property_name = FALSE;
+      // Set sub-property name, in case of nid bypass, since is is not required.
+      if (isset($info['sub_property']) && $info['sub_property'] != 'nid') {
+        $sub_property_name = $info['sub_property'];
+      }
       if (!isset($request[$public_field_name])) {
         // No property to set in the request.
         if ($null_missing_fields && $this->checkPropertyAccess('edit', $public_field_name, $wrapper->{$property_name}, $wrapper)) {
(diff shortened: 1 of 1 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -640,6 +640,11 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
       }
 
       $property_name = $info['property'];
+      $sub_property_name = FALSE;
+      // Set sub-property name, in case of nid bypass, since is is not required.
+      if (isset($info['sub_property']) && $info['sub_property'] != 'nid') {
```
e0ipso, line 645: When are having `'nid'` as a _sub_property_?

Also, this `'nid'` logic wouldn't scale for other entity types. Maybe consider using `entity_get_info($this->getEntityType())` to get the entity id key.

plugins/restful/RestfulEntityBase.php
```diff
@@ -655,7 +660,19 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
 
       $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
 
-      $wrapper->{$property_name}->set($field_value);
+      // Support sub property saving.
+      if ($sub_property_name) {
+        if (!$wrapper->{$property_name}->value()) {
+          // Init field.
+         $wrapper->{$property_name}->set(array($sub_property_name => ''));
```
e0ipso, line 667: Is it safe to init this to a string? Should this be a NULL?

Do we need to init the field?

e0ipso:
Thanks for the PR!

There are some comments along the code.

Also it seems that some tests are failing (maybe they need updating or maybe the new code is breaking something):
- Create Entity
- Entity Validator
- Sub-requests

amitaibu:
Hi Bruce! The `sub_property` already exists -- on which field type doesn't it work for you?

bricel:
I see it exist as a getter, but not as a setter. The use case I had is using the field_interval, which has 2 properties.

amitaibu:
I'll be working on this as it's indeed not clear where its best to add this part.

amitaibu:
closing in favor of #380

PR #386, opened 2015-01-21
After clean install update db 7100 is pending

e0ipso:
After clean install of restfull I get a pending update.
image


e0ipso:
Did you run it?

bricel:
Yes, but i suppose it should not be pending after a clean install, right ?

e0ipso:
@bricel, @amitaibu fix attached.

amitaibu:
Thanks!

PR #387, opened 2015-01-21
Fix missing entities returning HTTP 500 instead of a 4XX.

deviantintegral:
This fixes missing entity GETs returning a 500, because the exception class wasn't extending RestfulException. I wasn't totally sure where to put the test, but this seemed like a good start.

### DIFF
--- a/exceptions/RestfulUnprocessableEntityException.php
+++ b/exceptions/RestfulUnprocessableEntityException.php
@@ -5,7 +5,7 @@
  * Contains RestfulUnprocessableEntityException
  */
 
-class RestfulUnprocessableEntityException extends Exception {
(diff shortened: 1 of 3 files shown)


amitaibu:
Hi @deviantintegral !

Thanks for the PR. I think a better place for would be [here](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/tests/RestfulExceptionHandleTestCase.test)

e0ipso:
After merging https://github.com/deviantintegral/restful/pull/1 this can be merged.

deviantintegral:
Thanks @mateu-aguilo-bosch !

amitaibu:
Merged, thanks!

PR #385, opened 2015-01-22
defaultSort fails with custom resource properties returns SQL error

e0ipso:
I have a custom property added to a custom resource that extends `RestfulEntityBaseNode` like so:

```
<?php

class Example extends RestfulEntityBaseNode {

    $public_fields['weight'] = array(
      'callback' => array($this, 'customWeightCallback'),
    );
  }

  public function defaultSortInfo() {
    return array(
      'weight' => 'ASC',
    );
  }

  protected function customWeightCallback($entity) {
    return 0; // Could return anything
  }
}

```

According to the documentation this should be supported, from plugin.md:

> ## Defining a default sort
> 
>    You can also define default sort fields in your plugin, by overriding `defaultSortInfo()` in your class definition.
> 
> This method should return an associative array, with each element having a key that matches a field from `publicFieldsInfo()`, and a value of either 'ASC' or 'DESC'.

But what happens is that this throws an error:

```
{"type":"http:\/\/www.w3.org\/Protocols\/rfc2616\/rfc2616-sec10.html#sec10.5.1",
"title":"SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near \u0027ASC\nLIMIT 50 OFFSET 0\u0027 at line 2","status":500}
```

It appears that the current sort functionality perhaps only allow things that are actually queried out of the database, not calculated properties

Are custom properties supported as a sort parameter?

### DIFF
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -140,7 +140,9 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
 
     foreach ($sorts as $public_field_name => $direction) {
       // Determine if sorting is by field or property.
-      $property_name = $public_fields[$public_field_name]['property'];
+      if (!$property_name = $public_fields[$public_field_name]['property']) {
+        throw new \RestfulBadRequestException('The current sort selection does not map to any entity property or Field API field.');
+      }
       if (field_info_field($property_name)) {
         $query->fieldOrderBy($public_fields[$public_field_name]['property'], $public_fields[$public_field_name]['column'], $direction);
       }
@@ -165,7 +167,9 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
     $public_fields = $this->getPublicFields();
     foreach ($this->parseRequestForListFilter() as $filter) {
       // Determine if filtering is by field or property.
-      $property_name = $public_fields[$filter['public_field']]['property'];
+      if (!$property_name = $public_fields[$filter['public_field']]['property']) {
+        throw new \RestfulBadRequestException('The current filter selection does not map to any entity property or Field API field.');
+      }
       if (field_info_field($property_name)) {
         if (in_array(strtoupper($filter['operator'][0]), array('IN', 'BETWEEN'))) {
           $query->fieldCondition($public_fields[$filter['public_field']]['property'], $public_fields[$filter['public_field']]['column'], $filter['value'], $filter['operator'][0]);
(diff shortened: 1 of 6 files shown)


e0ipso:
@bmcmurray it seems that the documentation may be wrong in that case. You can only sort on public fields that map to an entity property or to a Field API field.

The rule of thumb is, if your public field defines a `'property'` element you can use EFQ to sort.

## Solution

Sorting the output array will only sort the current page of results, so it will not be acceptable. Also, that same sorting could be done in client side.

Therefore, when an unsortable public field is selected, throw an exception explaining the situation.

e0ipso:

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -140,7 +140,9 @@ protected function queryForListSort(\EntityFieldQuery $query) {
 
     foreach ($sorts as $public_field_name => $direction) {
       // Determine if sorting is by field or property.
-      $property_name = $public_fields[$public_field_name]['property'];
+      if (!$property_name = $public_fields[$public_field_name]['property']) {
```
amitaibu, line 143: I think we should use `\RestfulDataProviderEFQ::getColumnFromProperty` here (and maybe improve it to return FALSE of no `schema field` found) - as EMW allows defining "computed" properties that have no real schema field.

e0ipso, line 143: I'm not sure I follow. `getColumnFromProperty` will be run later. This check only makes sure that there is a `'property'` key in the public field definition. This is the same thing we were doing before, but now it's wrapped in an if clause.

amitaibu, line 143: well, probably as a follow up, we should throw an excpetion, if `getColumnFromProperty` returned FALSE (as some EMW properties may be "computed"). I'll open an issue for that.

tests/RestfulListTestCase.test
```diff
@@ -427,6 +436,19 @@ class RestfulListTestCase extends RestfulCurlBaseTestCase {
...
+    $request = array('filter' => array(
+      'static' => array(
+        'value' => 0,
+      )
+    ));
+    try {
+      $handler->get('', $request);
+      $this->fail('Exception not thrown for invalid filter.');
+    }
+    catch (\RestfulBadRequestException $e) {
```
amitaibu, line 449: maybe worth adding, like in other tests, another `catch (\Exception $e)` that will `$this->fail()` - to assert we have the correct exception thrown.

e0ipso, line 449: If there is an uncatched exception, the tests will fail and simpletest will provide more information than with a cached exception.

amitaibu, line 449: ok, in that case we'll need to remove those `catch`s from other tests ;) (but of course not on this PR).

amitaibu:
Merged, thanks.

tests/RestfulListTestCase.test
e0ipso, line 449: I'll open an issue.

PR #388, opened 2015-01-22
2.x Convert formatter plugins

e0ipso:


e0ipso:
@amitaibu tests are green.

plugins/restful/RestfulBase.php
```diff
@@ -543,17 +562,7 @@ public static function isValidMethod($method, $strict = TRUE) {
...
-
-  /**
-   * Get the formatter handler for the current restful formatter.
-   *
-   * @return \RestfulFormatterInterface
-   *   The formatter handler.
-   */
-  protected function formatter() {
-    return \RestfulManager::outputFormat($this);
+    return $this->formatterManager->format($data, $this->getPluginKey('formatter'));
```
amitaibu, line 565: I think it would look a bit cleaner to have a `$this->getFormatterManager()`

src/Formatter/FormatterManager.php
```diff
@@ -0,0 +1,136 @@
...
+   */
+  public function __construct($resource = NULL) {
+    $this->resource = $resource;
+    $manager = FormatterPluginManager::create();
+    $options = array();
+    foreach ($manager->getDefinitions() as $plugin_id => $plugin_definition) {
+      // Since there is only one instance per plugin_id use the plugin_id as
+      //instance_id.
+      // Add the restful resource to the plugin options.
+      // TODO: Remove this coupling.
```
amitaibu, line 38: can you explain the issue here, it's not clear to me.

src/Formatter/FormatterManager.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+    $this->plugins = new FormatterPluginCollection($manager, $options);
+  }
+
+
+  /**
+   * {@inheritdoc}
+   */
+  public function format(array $data, $formatter_name = NULL) {
+    $accept = empty($GLOBALS['_SERVER']['HTTP_ACCEPT']) ? NULL : $GLOBALS['_SERVER']['HTTP_ACCEPT'];
```
amitaibu, line 51: we should avoid globals inside the plugin.

e0ipso, line 51: Yes! I just submitted the PR that will make this possible.

src/Plugin/formatter/FormatterJson.php
```diff
@@ -27,17 +39,17 @@ public function prepare(array $data) {
 
     $output = array('data' => $data);
```
amitaibu, line 40: not in the scope of this PR (and sorry for doing the review post-merge), but we should allow the restful resource inject not only the 'data' property but also other properties (e.g. `'total posts by author'`)

src/Formatter/FormatterManager.php
e0ipso, line 38: Ideally the formatter should not need to be composed with the resource. We are creating circular dependencies (since the resource is composed with the formatter via the formatter manager), and thus coupling those classes very tightly.

Not sure how to deal with this as of now, that's why I put a TODO.

plugins/restful/RestfulBase.php
e0ipso, line 565: Would you have the `formatterManager` be accessible via a public method?

src/Plugin/formatter/FormatterJson.php
e0ipso, line 40: I agree 100% on this. This would be a path to remove the coupling between the two objects.

Care to create an issue and link here?

src/Formatter/FormatterManager.php
amitaibu, line 38: > since the resource is composed with the formatter via the formatter manager

Why do we do this - do we use the formatter from within the resource?

src/Plugin/formatter/FormatterJson.php
amitaibu, line 40: Yes, it's something I'm thinking to do for 1.x as-well

src/Formatter/FormatterManager.php
e0ipso, line 38: 1.x has the formatter in order to call `$handler->format()`. In 2.x the process will be a bit different:

_Manager_ will have: a _resource_ to get the data, a _formatter_ to format it and a _response_ to return it.

amitaibu, line 38: 2.x makes more sense :wink:

PR #395, opened 2015-01-27
2.x Request object

e0ipso:
This PR breaks RESTful pretty badly. Still I wanted to throw it out there to show progress and architecture.

Additions:
- `Http` component:
  - `HttpHeader` class that deals with a single header.
  - `HttpHeaderBag` class that deals with a collection of headers.
  - `Request` class that deals with everything from the incoming request.

Ideally there will be a `RestfulManager` that will have a request object. The `RestfulManager` would be available from everywhere and therefore the request (and the header bag, and therefore each header).


amitaibu:
This has a symfony feeling to it :) So crazy idea, since we're already using composer manager. Why not use symfony here? :)

e0ipso:
> This has a symfony feeling to it :)

That's good!

> Why not use symfony here? :)

I pondered the idea, but we can't just ignore Drupal's API at the HTTP level. See how there is nothing there that relates to the hostname, or methods to deal with query string parsing, …

Using the symfony component would add more work, since we'd have to override a lot of implementations.

amitaibu:
yeah, it was more of a crazy idea :)

src/http/HttpHeaderBag.php
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Add a header to the bag.
+   *
+   * @param array|HttpHeaderInterface $header
+   *   The header object or an associative array with the name and value.
+   *
+   * @throws \RestfulServerConfigurationException
+   */
+  public function bag($header) {
```
amitaibu: Why allow both types, and not inforce `HttpHeaderInterface`?

amitaibu: I think the method name should describe its action. `addHeader()`?

src/http/HttpHeader.php
```diff
@@ -0,0 +1,129 @@
...
+   * {@inheritdoc}
+   */
+  public function get() {
+    return $this->values;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function contents() {
```
amitaibu: maybe `getString()` or `getFormatted()` (since just by the name it's not clear the difference between `get()` and `contents()`). (btw, Is there a Symfony equivalent - I didn't find one)

src/Http/Request.php
```diff
@@ -0,0 +1,332 @@
...
+    if (!static::isValidMethod($method, FALSE)) {
+      throw new \RestfulBadRequestException('Unrecognized HTTP method.');
+    }
+    return new static($path, $query, $method, $headers, $via_router, $csrf_token, $cookies, $files, $server);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function createFromGlobals() {
```
amitaibu, line 149: where is this being called from  - the menu callback?

src/Http/Request.php
```diff
@@ -0,0 +1,332 @@
...
+    $headers = static::parseHeadersFromGlobals();
+    $csrf_token = $headers->get('x-csrf-Token')->contents();
+
+    return static::create($path, $query, $method, $headers, $via_router, $csrf_token, $_COOKIE, $_FILES, $_SERVER);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function isWriteMethod($method, $strict = TRUE) {
```
amitaibu: I don't remember why we have the $strict, I think it had to do something with testing -- can we remove it?

src/http/HttpHeaderBag.php
amitaibu: I see that we use `$bag->bag(HttpHeader::create($name, $value));` - so I think it's ok to change it to 

``` php
public function bag(HttpHeaderInterface $header) {
```

src/Http/Request.php
```diff
@@ -0,0 +1,332 @@
...
+    ));
+  }
+
+  /**
+   * Get the credentials based on the $_SERVER variables.
+   *
+   * @return array
+   *   A numeric array with the username and password.
+   */
+  protected static function getCredentials() {
```
amitaibu, line 298: I think understand why you moved it to `Request`, since it's dealing with `$_SERVER`, but it feels it shouldn't be here

src/Http/Request.php
```diff
@@ -0,0 +1,332 @@
...
+      $_SERVER['PHP_AUTH_PW'] = $password;
+    }
+
+    return array($username, $password);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getUser() {
```
amitaibu, line 319: `getUserName`

src/Http/Request.php
```diff
@@ -0,0 +1,332 @@
...
+    list($user,) = static::getCredentials();
```
amitaibu: $user => $account (as best practice, to avoid touching the `global $user` somewhere down the line).

src/Http/Request.php
amitaibu, line 319: or maybe better `getAuthUserName`

src/Http/RequestInterface.php
```diff
@@ -0,0 +1,85 @@
...
+   */
+  public static function isValidMethod($method, $strict = TRUE);
+
+  /**
+   * Returns the user.
+   *
+   * @return
+   *   string|null
+   */
+  public function getUser();
```
amitaibu, line 69: do we need this in the interface?

amitaibu:
This PR makes a lot of sense to me, RESTful is definitely going in the right direction..

src/http/HttpHeaderBag.php
e0ipso: :+1:

src/http/HttpHeader.php
e0ipso: ~~It could be the `__toString` method, with an alias of `getString()`.~~

src/Http/Request.php
e0ipso, line 149: It's not called from anywhere yet. It will be called from the `RestfulManager` that will get instantiated on the menu callback.

src/Http/Request.php
e0ipso: Sure we can.

src/Http/Request.php
e0ipso, line 298: The Symfony request component also has it in the request class.

src/Http/Request.php
e0ipso, line 319: I'd rather keep the same name as in the Symfony request component.

src/Http/Request.php
e0ipso: :+1:

src/Http/RequestInterface.php
e0ipso, line 69: Any public method in the class should be surfaced via the interface. If it's not in the interface we shouldn't be declaring it public.

Anyways, this is also a public method in the Symfony component.

src/http/HttpHeaderBag.php
e0ipso: > I think the method name should describe its action. addHeader()

I think that since the object is a `HeaderBag` adding the word _header_ in the method name is not necessary. I was going for bag, as in _to bag_ (like putting something inside a bag). I will change it to _add_.

e0ipso:
@amitaibu feedback implemented in 176fd8e.

e0ipso:
@amitaibu after addressing your concerns I'm going to go ahead and merge this one. Anything extra can be done in a new pr.

PR #398, opened 2015-01-28
2.x Restful manager

e0ipso:
This PR mocks some classes and introduces the new `RestfulManager` service.

PR #399, opened 2015-01-28
2.x Resource manager

e0ipso:
~~Depends on #398.~~


e0ipso:
@amitaibu you can see that we are moving in the right direction because we are moving stuff from `RestfulBase` into the appropriate subsystems.

I feel that I cannot avoid porting the resource plugins much longer… But there are several things first that can be done:
- Namespace the exceptions.
- Create the response object.
- …

amitaibu:
> I feel that I cannot avoid porting the resource plugins much longer

What is holding you? apart of the crazy amount of work ;)

e0ipso:
> apart of the crazy amount of work

_Just_ that.

amitaibu:
> Just that.

No worries you won't have to do it alone :)

PR #295, opened 2015-01-28
Filtering nodes as anonymous user

e0ipso:
Hi,

For some weird reason, when I do a query with filters on my API which uses the restful module, I get the following error:

<code>SQLSTATE[42703]: Undefined column: 7 ERROR: column field_data_field_operation_type0.nid does not exist LINE 11: ....realm = 'all') ))AND (na.grant_view >= '1') AND (field_data... ^</code>

I've found out that the issue comes from the <a href="https://api.drupal.org/api/drupal/modules!node!node.module/function/_node_query_node_access_alter/7">_node_query_node_access_alter</a> function, which, when not logged in as an admin, overrides the EntityFieldQuery.

I haven't found any way to fix this issue yet, but was wondering if someone had run into it at some point...

### DIFF
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -252,10 +252,10 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
       $query->entityCondition('entity_id', $ids, 'IN');
     }
 
+    $this->queryForListFilter($query);
+
     $this->addExtraInfoToQuery($query);
     $query->addTag('restful_count');
-
-    $this->queryForListFilter($query);
 
     return $query->count();
   }
@@ -278,9 +278,13 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
   protected function addExtraInfoToQuery($query) {
     parent::addExtraInfoToQuery($query);
     $entity_type = $this->getEntityType();
-    // Add a generic entity access tag to the query.
-    $query->addTag($entity_type . '_access');
-    $query->addMetaData('restful_handler', $this);
+    // The only time you need to add the access tags to a EFQ is when you don't
+    // have fieldConditions.
+    if (empty($query->fieldConditions)) {
+      // Add a generic entity access tag to the query.
+      $query->addTag($entity_type . '_access');
+      $query->addMetaData('restful_handler', $this);
+    }
   }
 
   /**
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -283,8 +283,8 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
     if (empty($query->fieldConditions)) {
       // Add a generic entity access tag to the query.
       $query->addTag($entity_type . '_access');
-      $query->addMetaData('restful_handler', $this);
     }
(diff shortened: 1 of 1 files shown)


e0ipso:
I'm seeing this same error.

e0ipso:
``` php
$query = new \EntityFieldQuery();

$results = $query
  ->entityCondition('entity_type', 'file')
  ->propertyCondition('bundle', 'video')
  ->fieldCondition('field_permalink', 'value', 'Whatever…')
  ->addTag('file_access')
  ->range(0, 1);

try {
  $query->execute();
  drush_log('Everything is fine. File.', 'ok');
}
catch (\PDOException $e) {
  drush_log('Womp, womp. Something went wrong. File.', 'error');
}
```

This code results on this same error.

amitaibu:
The problem is probably that Drupal doesn't add the base table. The work around that I tried to do, is doing something like this:

``` php
$query
  ->entityCondition('entity_type', 'file')
  ->entityCondition('id', 0, '>') // <-- this work around
  ...
```

Just to force the base table to be added. @mateu-aguilo-bosch Can you check that?

amitaibu:
(although the work around did _not_ work when I tried it with @guillaumev steps)

e0ipso:
Got off a call with Dave Reid and he told me that we do not need to add the access tag, since EFQ will add it by default.

I will make a PR asap.

amitaibu:
> since EFQ will add it by default.

Are you sure? We are adding a tag in Entity reference for example.

e0ipso:
@amitaibu I meant when you have field conditions applied to it.

e0ipso:
@amitaibu pr submitted.

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -278,9 +278,13 @@ public function getTotalCount() {
...
     parent::addExtraInfoToQuery($query);
     $entity_type = $this->getEntityType();
-    // Add a generic entity access tag to the query.
-    $query->addTag($entity_type . '_access');
-    $query->addMetaData('restful_handler', $this);
+    // The only time you need to add the access tags to a EFQ is when you don't
+    // have fieldConditions.
+    if (empty($query->fieldConditions)) {
+      // Add a generic entity access tag to the query.
+      $query->addTag($entity_type . '_access');
```
amitaibu, line 285: so the problem is that tag was added twice?

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -278,9 +278,13 @@ public function getTotalCount() {
...
+      $query->addMetaData('restful_handler', $this);
```
amitaibu: I think this should stay out of the IF

plugins/restful/RestfulDataProviderEFQ.php
e0ipso, line 285: There is no documentation around this (security!) issue. But from what I gather you are not supposed to add access tags to entities when using fieldConditions. On the contrary you need to add the `DANGEROUS_ACCESS_CHECK_OPT_OUT` if you want to avoid them.

The `field_sql_storage` module will add the tag `entity_field_access` when doing a field condition.

e0ipso:
@amitaibu moved out of the if statement.

amitaibu:
Ok, I'll have to reproduce the original error, and see if the PR fixed it.

amitaibu:
Merged.

amitaibu:
Thanks :)

PR #400, opened 2015-01-29
2.x namespace exceptions

e0ipso:
~~Depends on #399.~~

This will autoload exceptions namespacing them under `Drupal\restful\Exception` and removing the initial _Restful_ from the class name.


amitaibu:
Looks good

PR #401, opened 2015-01-30
2.x Response object.

e0ipso:
@amitaibu minor break through!

After this commit you will be able to make a request for an API that that does not exits and get a fully formed and formatted error response.

This is handy to see how the `Request` + `HttpHeader…` + `RestfulManager` + `FormatterManager` + `Response`  classes interact to the life cycle of a request.

### DIFF
--- a/src/Http/Response.php
+++ b/src/Http/Response.php
@@ -120,6 +120,48 @@ class Response implements ResponseInterface {
   );
 
   /**
+   * Constructor.
+   *
+   * @param mixed $content
+   *   The response content, see setContent()
+   * @param int $status
+   *   The response status code
+   * @param array $headers
+   *   An array of response headers
+   *
+   * @throws UnprocessableEntityException
+   *   When the HTTP status code is not valid
+   */
+  public function __construct($content = '', $status = 200, $headers = array()) {
+    $this->headers = new HttpHeaderBag($headers);
+    $this->setContent($content);
+    $this->setStatusCode($status);
+    $this->setProtocolVersion('1.0');
+    if (!$this->headers->has('Date')) {
+      $this->setDate(new \DateTime(NULL, new \DateTimeZone('UTC')));
+    }
+  }
+
+  /**
+   * Returns the Response as an HTTP string.
+   *
+   * The string representation of the Response is the same as the
+   * one that will be sent to the client only if the prepare() method
+   * has been called before.
+   *
+   * @return string
+   *   The Response as an HTTP string
+   *
+   * @see prepare()
+   */
+  public function __toString() {
+    return
+      sprintf('HTTP/%s %s %s', $this->version, $this->statusCode, $this->statusText) . "\r\n" .
+      $this->headers . "\r\n" .
+      $this->getContent();
+  }
+
+  /**
    * Is the response empty?
    *
    * @return bool
@@ -164,9 +206,6 @@ class Response implements ResponseInterface {
    *
    * @param Request $request
    *   A Request instance
-   *
-   * @return Response
-   *   The current response.
    */
   public function prepare(Request $request) {
     $headers = $this->headers;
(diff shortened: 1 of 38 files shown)


src/Http/Response.php
```diff
@@ -3,8 +3,396 @@
...
+  public function getHeaders() {
+    return $this->headers;
+  }
+
+  /**
+   * Checks if we need to remove Cache-Control for SSL encrypted downloads when using IE < 9.
+   *
+   * @link http://support.microsoft.com/kb/323308
+   */
+  protected function ensureIEOverSSLCompatibility(Request $request) {
```
amitaibu, line 366: is this taken from Symfony?

amitaibu, line 366: answering myself - [yes](https://api.drupal.org/api/drupal/core%21vendor%21symfony%21http-foundation%21Symfony%21Component%21HttpFoundation%21Response.php/function/Response%3A%3AensureIEOverSSLCompatibility/8) :)

src/Http/Response.php
```diff
@@ -3,8 +3,396 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setContent($content) {
+    if ($content !== NULL && !is_string($content) && !is_numeric($content) && !is_callable(array($content, '__toString'))) {
+      throw new InternalServerErrorException(sprintf('The Response content must be a string or object implementing __toString(), "%s" given.', gettype($content)));
+    }
+    $this->content = (string) $content;
+    $this->headers->add(HttpHeader::create('Content-Length', strlen($this->content)));
```
amitaibu, line 262: Seems we are copying much of Symfony's response (which is good), but doing small overrides like this line (which I'm not clear why is needed here).

So maybe it's back to the question, why not extend Symfony's `Response` and just override what's needed?

e0ipso, line 262: We're just copying some of methods, I don't think that justifies importing the whole http foundation component. I copied the gotchas ☺

amitaibu, line 262: :+1:

src/http/HttpHeaderBag.php
```diff
@@ -37,28 +59,97 @@ public function get($key) {
     if (array_key_exists($key, $this->values)) {
       return $this->values[$key];
     }
-    return NULL;
+    // Return a NULL object on which you can still call the HttpHeaderInterface
+    // methods.
+    return HttpHeaderNull::create(NULL, NULL);
```
amitaibu, line 64: this looks a little weird - why do we need this null implementation?

e0ipso, line 64: This is actually kind of cool (from how I see it). It will allow us doing things like:

``` php
if ($value = $headers->get('header-that-does-not-exist')->getValueString()) { … }
```

Instead of the more verbose (and awkard):

``` php
if ($header = $headers->has('header-that-does-not-exist')) {
  $value = $headers->get('header-that-does-not-exist')->getValueString();
  …
}
```

https://en.wikipedia.org/wiki/Null_Object_pattern

I don't feel to attached to it if you don't like it, though.

amitaibu, line 64: Or maybe you could have already gotten the string/ array from the `get()` - `function get($string = TRUE)`

``` php
$string = $headers->get('header-that-does-not-exist');
$values = $headers->get('header-that-does-not-exist', FALSE);
```

e0ipso, line 64: The problem of that is that it muddies the interface by having:

``` php
/**
  * @return string|HttpHeaderInterface
  */
```

Which means that your method chaining is cut depending on the passed in arguments. I'd rather have methods returning a single interface or `NULL`.

amitaibu, line 64: alright :)

e0ipso:
@amitaibu this last commit fixed the request alter hook (along with some other minor fixes).

amitaibu:
Looks good, can be merged.

PR #407, opened 2015-02-12
2.x resource conversion

e0ipso:
This PR contains incomplete work of how the new classes are going to interact with each other. There is a lot still missing, and none of the tests apply yet.

@amitaibu I wanted to show some progress. Maybe you can start reviewing so we catch implementation problems early on.


amitaibu:
@mateu-aguilo-bosch wow, 86 files? Is this updated with the latest merges of 2.x?

e0ipso:
@amitaibu it seems it has not.

e0ipso:
There we go.

amitaibu:
Ok, 24 files review is something I can do :)

amitaibu:
(I'll do the review tomorrow)

e0ipso:
Great!

amitaibu:
@mateu-aguilo-bosch can you a few (non-working even) code snippets, so it's easier to understand the PR?

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -0,0 +1,321 @@
...
+      return;
+    }
+
+
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function viewEntity($id) {
```
e0ipso, line 115: This is a leftover and will go away in the next PR.

e0ipso:
@amitaibu this PR will basically introduce 3 classes that will replace the `RestfulBase` classes.
1. Adds `Resource` class. It's the plugin class (the one that gets the user configuration via annotations). It's in charge of the controllers and decorates the data provider for the CRUD operations. The appropriate data provider type is declared by overriding the `dataProviderFactory` method (entity, DB query, Plugins, …). The `Resource` class is abstract, that means that we'll need more concrete implementations to work with, like `EntityResource` and `DBResource` with the abstract `dataProviderFactory` implemented.
2. Adds `DataProvider` class. That will be in charge of performing the CRUD operations with the storage (the database, the filesystem, external services, …).
3. Adds `ResourceField` class family. That encapsulate the field declaration that will drive the `DataProvider`. For instance, a `ResourceFieldEntity` will have all the properties needed (`method_on_wrapper`, `property`, `column`, …) for the `DataProviderEntity` to use `\EntityMetadataWrapper` to get/set that particular field. Every subtype of `ResourceField` can have different properties. All the basic properties are included in the base class `ResourceField`.

A basic implementation would be:

``` php
use Drupal\restful\Plugin\resource\DataProvider\DataProvider;

/**
 * Class ArticleResource
 * @package Drupal\restful\Plugin\resource
 *
 * @Resource(
 *   name = "blah",
 *   dataProvider = {
 *     entity_type = "node",
 *     bundles = [ "article" ]
 *   },
 *   …
 * )
 */
class ArticleResource extends Resource {

  /**
   * Overwrites Resource::dataProviderFactory().
   */
  protected function dataProviderFactory() {
    $plugin_definition = $this->getPluginDefinition();
    return new DataProviderEntity($this->getRequest(), $this->fieldDefinitions, $this->getAccount(), $plugin_definition['dataProvider'], $GLOBALS['language']->language);
  }

  /**
   * Overwrites Resource::fieldDefinitionInfo().
   */
  protected function fieldDefinitionInfo() {
    // TODO: This has to be put in place yet.
    $fields['label'] = array(
      'wrapper_method' => 'label',
      'method_on_wrapper' => TRUE,
      'process callbacks' => array(
        array(array($this, 'prefix'), array('Label: ')),
      ),
    );
    …
    return $fields;
  }
}
```

Note that `fieldDefinitionInfo` has not been wired yet. There will be a method that will take the result of `fieldDefinitionInfo` and turn it into a `ResourceFieldCollection`.

e0ipso:
Also, it has not been removed yet but the passthrough feature has been replaced with a more flexible alternative.

Basically for every field you can specify which methods it's active. That means that we can have `POST` only fields and any other combination.

src/Http/Request.php
```diff
@@ -258,6 +275,9 @@ public function getParsedBody() {
    *   The parsed body.
    */
   protected static function parseBody($method) {
+    if (!static::isWriteMethod($method)) {
```
amitaibu, line 278: what's the difference between `parseBody` and `parseInput`?

src/Http/Request.php
```diff
@@ -283,6 +303,16 @@ protected static function parseBody($method) {
   }
 
   /**
+   * Parses the input data.
+   *
+   * @return array
+   *   The parsed input.
+   */
+  protected static function parseInput($method) {
+    return $_GET;
```
amitaibu, line 312: I assume it's wrong :)

src/Plugin/resource/CrudInterface.php
```diff
@@ -0,0 +1,77 @@
...
+   * @param mixed $object
+   *   The thing that will be set.
+   * @param bool $replace
+   *   TRUE if the contents of $object will replace $subject entirely. FALSE if
+   *   only what is set in $object will replace those properties in $subject.
+   *
+   * @return array
+   *   An array of structured data for the thing that was updated.
+   */
+  public function update($subject, $object, $replace = TRUE);
```
amitaibu, line 67: $subject => $identifier

src/Plugin/resource/DataProvider/DataProviderInterface.php
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * @file
+ * Contains \Drupal\restful\Plugin\resource\DataProvider\DataProviderInterface.
+ */
+
+namespace Drupal\restful\Plugin\resource\DataProvider;
+
+use Drupal\restful\Plugin\resource\CrudInterface;
+
+interface DataProviderInterface extends CrudInterface {}
```
amitaibu, line 12: Why do we need the `CrudInterface` in the first place?

src/Plugin/resource/Field/ResourceFieldBase.php
```diff
@@ -0,0 +1,222 @@
...
+ * @file
+ * Contains \Drupal|restful\Plugin\resource\Field\ResourceFieldBase.
+ */
+
+namespace Drupal\restful\Plugin\resource\Field;
+
+use Drupal\restful\Exception\ServerConfigurationException;
+use Drupal\restful\Http\Request;
+
+abstract class ResourceFieldBase implements ResourceFieldInterface {
```
amitaibu, line 13: Maybe we should call it `ResourceEntityFieldBase` as it specifically deals with field/ properties that can be wrapped with $wrapper?

src/Plugin/resource/Field/ResourceFieldBase.php
```diff
@@ -0,0 +1,222 @@
...
+   *   );
+   *
+   * @var array
+   */
+  protected $resource = array();
+
+  /**
+   * The HTTP methods where this field applies.
+   *
+   * This replaces the create_or_update_passthrough feature. Defaults to all.
```
amitaibu, line 88: :+1:

src/Plugin/resource/Field/ResourceFieldCollection.php
```diff
@@ -0,0 +1,160 @@
...
+      $field = ResourceField::create($field_info);
+      $this->fields[$field->id()] = $field;
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function factory(array $fields = array()) {
+    return new static($fields);
```
amitaibu, line 100: I'm not clear on `factory` and `create` - what are we trying to do here?

src/Plugin/resource/Field/ResourceFieldCollection.php
```diff
@@ -0,0 +1,160 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\restful\Plugin\resource\Field\ResourceFieldCollection.
+ */
+
+namespace Drupal\restful\Plugin\resource\Field;
+
+class ResourceFieldCollection implements ResourceFieldCollectionInterface {
```
amitaibu, line 10: This one seems to be related to entity fields. The problem I see is for other providers (e.g. db), we will have to repeat some of the definitions

amitaibu, line 10: or is intended to be the only class, and it should be decorated per data provider?

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * {@inheritdoc}
+   */
+  public function addDefaults() {
+    // Almost all the defaults come are applied by the object's property
+    // defaults.
+
+    // Set the defaults from the decorated.
+    $this->setResource($this->decorated->getResource());
```
amitaibu, line 189: I'm not clear on this - what are we trying to do here?

src/Plugin/resource/Resource.php
```diff
@@ -0,0 +1,303 @@
...
+  abstract protected function dataProviderFactory();
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getRequest() {
+    if (isset($this->request)) {
+      return $this->request;
+    }
+    if (!$this->request = $this->configuration['request']) {
```
amitaibu, line 75: when can this happen?

src/Plugin/resource/Resource.php
```diff
@@ -0,0 +1,303 @@
...
+    $path = $this->getPath();
+
+    return ResourceManager::executeCallback($this->getControllerFromPath($path), array($path));
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function contollersInfo() {
+    // Provide sensible defaults for the HTTP methods. These methods (index,
```
amitaibu, line 142: move this to the PHPdocs

src/Plugin/resource/Resource.php
```diff
@@ -0,0 +1,303 @@
...
+   */
+  public function index($path) {
+    return $this->getDataProvider()->index();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function view($path) {
+    // TODO: Compare this with 1.x logic.
```
amitaibu, line 194: explain?

src/Plugin/resource/ResourceEntity.php
```diff
@@ -0,0 +1,69 @@
...
+   * @var string
+   */
+  protected $entityType;
+
+  /**
+   * The entity bundles.
+   *
+   * @var array
+   */
+  protected $bundles;
```
amitaibu, line 28: this changes 1.x logic where the base is assuming a single bundle?

src/Resource/ResourceManager.php
```diff
@@ -109,6 +109,19 @@ public static function executeCallback($callback, array $params = array()) {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  public static function isValidCallback($callback) {
+    if (!is_callable($callback)) {
+      if (is_array($callback) && count($callback) == 2 && is_array($callback[1])) {
```
amitaibu, line 116: worth adding a comment or two with an example.

src/Resource/ResourceManagerInterface.php
```diff
@@ -52,4 +52,35 @@ public function negotiate();
    */
   public static function executeCallback($callback, array $params = array());
 
+  /**
+   * Is valid callback?
```
amitaibu, line 56: => `Determine if a callback is valid.`

src/http/Request.php
```diff
@@ -127,10 +127,17 @@ class Request implements RequestInterface {
   /**
    * Holds the parsed body.
    *
+   * @var array
```
amitaibu, line 130: this file seems to be duplicated

`src/http/Request.php` vs `src/Http/Request.php`

src/Http/Request.php
e0ipso, line 278: Technically the body of the request is what comes after the headers in the http request.

In 7.x-1.x we have the `$request` variable that can hold the body and the query string input arguments. 7.x-2.x separates the two concepts.

src/Http/Request.php
e0ipso, line 312: I think it is not. This returns an structured array of the user input via query string.

src/Plugin/resource/CrudInterface.php
e0ipso, line 67: :+1:

src/Plugin/resource/DataProvider/DataProviderInterface.php
e0ipso, line 12: The `DataProviderInterface` **will** have more stuff in the future, not just the `CrudInterface`.

Sorry, there are too many moving pieces. but submitting a complete thing would have been a nightmare to review.

src/Plugin/resource/Field/ResourceFieldBase.php
e0ipso, line 13: Actually, there is `ResourceFieldEntity` for that. In this class we only deal with the basics (all that applies to all data providers). `'property'` is supposed to act as a mapping to the _field_ identifier:
- Field API field name. Entity provider.
- Column name. DB provider.
- Array key. Plugin provider.

src/Plugin/resource/Field/ResourceFieldCollection.php
e0ipso, line 100: In this case, factory is just a proxy method. When doing testing we may want to create an object from an array of fields instead of forcing the use of `static::getInfo()`. In that situation we can use this `factory` method.

src/Plugin/resource/Field/ResourceFieldCollection.php
e0ipso, line 10: `ResourceFieldCollection` can be seen as a glorified array of `ResourceFieldInterface` objects. But with the ability to add methods to the collection.

It can hold _resource fields_ of any kind (even mixing types in a collection). A _resource field_ is every one of the entries of the mapping array in 7.x-1.x + the public name.

src/Http/Request.php
amitaibu, line 312: but doesn't it care about the $method?

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 189: This sets the resource from `ResourceFieldBase` to be the resource for `ResourceFieldEntity`. This is just a way to simulate inheritance without having to use a complete decorator.

We want to avoid inheritance because we could get in these situations were:
- A contrib module adds a class `AcmeResourceFieldEntity` that inherits from `ResourceFieldEntity`.
- In parallel we inherit from `ResourceFieldEntity` to create `CachedResourceFieldEntity`.

Now there are 2 different inheritance branches, and no `AcmeCachedResourceFieldEntity` is possible without a rewrite.

src/Plugin/resource/Resource.php
e0ipso, line 75: Not sure it can happen. It's just being thorough in error handling.

Maybe for mocked classes in tests?

src/Plugin/resource/Resource.php
e0ipso, line 142: :+1:

src/Plugin/resource/Resource.php
e0ipso, line 194: Just a note to make sure that the end result of a view is the same in 7.x-1.x and 7.x-2.x.

src/Plugin/resource/ResourceEntity.php
e0ipso, line 28: Yeah, I could not remember the reason behind a single bundle. I figured that it would be easier to move from single to multiple than the other way around.

src/Resource/ResourceManagerInterface.php
e0ipso, line 56: Yeah… I'm not a fan of Symfony2 style comments either.

e0ipso, line 56: :+1:

src/Resource/ResourceManager.php
e0ipso, line 116: :+1:

src/Plugin/resource/ResourceEntity.php
amitaibu, line 28: > Yeah, I could not remember the reason behind a single bundle

For `create` - you must know the bundle. Also, now with more experience, indeed most resources are bundle specific.

src/http/Request.php
e0ipso, line 130: :hankey:

e0ipso, line 130: That file is not in my local :-(

I tried removing it and re-creating it, but it did not help…

e0ipso:
@amitaibu comments have been addressed.

amitaibu:
cool, feel free to merge :)

src/http/Request.php
amitaibu, line 130: it can probably be removed in the `git merge --squash`

e0ipso:
Merged.

PR #409, opened 2015-02-17
Set json to be the default formatter on admin page

amitaibu:
restful___site-install


amitaibu:
@mateu-aguilo-bosch ^^

e0ipso:
Hmmm. Didn't we do this already?

PR #410, opened 2015-02-17
Switch global user by user authenticated by RESTful

amitaibu:
This PR started as a naive feature to allow user to `POST https://example.com/api/foo?access_token=foo`.

But it ended with RESTful needing to switch the `global $user` before the actual processing, as different places (e.g. `entity_metadata_file_access()`) assume the current user is the global $user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,9 +28,8 @@ control.
 
 ## Module dependencies
 
-  * [Entity API](https://drupal.org/project/entity), with the following patches:
-  * [$wrapper->access() might be wrong for single entity reference field](https://www.drupal.org/node/2264079#comment-8911637)
-  * [Prevent notice in entity_metadata_no_hook_node_access() when node is not saved](https://drupal.org/node/2086225#comment-8768373)
+  * [Entity API](https://drupal.org/project/entity), with the following patche:
+  * [Prevent notice in entity_metadata_no_hook_node_access() when node is not saved](https://www.drupal.org/node/2086225#comment-9627407)
 
 ## Recipes
 Read even more examples on how to use the RESTful module in the [module documentation
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ control.
 
 ## Module dependencies
 
-  * [Entity API](https://drupal.org/project/entity), with the following patche:
+  * [Entity API](https://drupal.org/project/entity), with the following patch:
   * [Prevent notice in entity_metadata_no_hook_node_access() when node is not saved](https://www.drupal.org/node/2086225#comment-9627407)
 
 ## Recipes
--- a/modules/restful_token_auth/restful_token_auth.module
+++ b/modules/restful_token_auth/restful_token_auth.module
@@ -49,8 +49,16 @@ function restful_token_auth_restful_parse_request_alter(&$request) {
   $plugin = restful_get_authentication_plugin('token');
   $param_name = $plugin['options']['param_name'];
 
+  $value = \RestfulManager::getRequestHttpHeader($param_name);
+  if (!$value && !empty($_GET[$param_name])) {
+    // If no access token found on the HTTP header, check if it's in the URL
+    // itself. This allows to do a POST request to for example:
+    // https://example.com/api/file-upload?access_token=foo
+    $value = $_GET[$param_name];
+  }
+
   $request['__application'] += array(
-    $param_name => \RestfulManager::getRequestHttpHeader($param_name),
+    $param_name => $value,
   );
 }
 
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -139,13 +139,13 @@ class RestfulAuthenticationManager extends \ArrayObject {
     // Determine if this is the first time we enter this method, so we capture
     // the original user, even if the authentication manager was initialized
     // multiple times.
-    // This value has to persist across calls to drupal_static_reset(), since a
-    // potentially wrong or disallowed session would be written otherwise.
+    // We don't use a static variable, as we need to make sure the original user
+    // session doesn't change even if the authentication manager was initialized
+    // multiple times
+    global $restful_switch_user;
 
-    static $first_time = TRUE;
-
-    if ($first_time) {
-      $first_time = FALSE;
+    if (empty($restful_switch_user)) {
+      $restful_switch_user = TRUE;
 
       $session = drupal_save_session();
       $this->setOriginalUserSession(array(
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -135,9 +135,18 @@ class RestfulAuthenticationManager extends \ArrayObject {
   public function switchUser() {
     global $user;
 
-    // If this is the first time we enter this method we capture the original
-    // user.
-    if (!$this->getOriginalUserSession()) {
+    // @todo: Check if this is correct.
+    // Determine if this is the first time we enter this method, so we capture
+    // the original user, even if the authentication manager was initialized
+    // multiple times.
+    // We don't use a static variable, as we need to make sure the original user
+    // session doesn't change even if the authentication manager was initialized
+    // multiple times
+    $restful_switch_user = &drupal_static(__FUNCTION__, FALSE);
+
+    if (empty($restful_switch_user) || !$this->getOriginalUserSession()) {
+      $restful_switch_user = TRUE;
+
       $session = drupal_save_session();
       $this->setOriginalUserSession(array(
         'user' => $user,
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -78,6 +78,7 @@ class RestfulAuthenticationManager extends \ArrayObject {
    */
   public function getAccount(array $request = array(), $method = \RestfulInterface::GET, $cache = TRUE) {
     global $user;
+
     // Return the previously resolved user, if any.
     if (!empty($this->account)) {
       return $this->account;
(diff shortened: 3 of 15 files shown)


amitaibu:
this will also probably fix the private image styles can't be used with token auth.

amitaibu:
Tests unsurprisingly are hating this change... :/

@mateu-aguilo-bosch before I fight with it, does the approach sound ok to you?

amitaibu:
btw, Seems all the failing tests are the ones extending `RestfulCurlBaseTestCase`

amitaibu:
Ok, failing tests were not related to the user switch :)

amitaibu:
Tests are green. Next is adding a test.

amitaibu:
actually, I probably need to move the swtichUser to the `authenticationManager`, so on setAccount() a switch will occur.

amitaibu:
> actually, I probably need to move the swtichUser to the authenticationManager, so on setAccount() a switch will occur.

After some work, I think the current implementation is better

amitaibu:
I have no clue how to test the user switch :/

e0ipso:
I remember a conversation in Amsterdam with @klausi when he suggested that we better stay away of logging users, and I agree. Any flaw in the RESTful authentication system leads to access to protected resources (which is bad enough), but if we log the user in it means that that impostor has access to the system as the impersonated user.

Can we circumvent these problems without poisoning the `$user` global?

amitaibu:
Yes, this PR _looks_ risky, but I _think_ we're doing it right

we're not really logging the user - we just override the global user for that specific request. Subtle difference, but an important one.

We are basically following the [best practice](https://www.drupal.org/node/218104) to do it, so even if RESTful will throw an exception, that user will _not_ be logged in.

The only case where we allow an actual login, is in the [login](https://github.com/RESTful-Drupal/restful/pull/410/files#diff-e54485f08c34e5e0a615f1ab1f660adeR53), where it is in fact the required behavior.

amitaibu:
but yeah, I'd love if @klausi could have a quick look ;)

amitaibu:
Ok, I think I know how to test this. I'll send a request and will prove a user is switch, and then switched back.

amitaibu:
test added

amitaibu:
Back to WIP, as I'm moving the `switchUser()` to a better place. I think this PR is much needed, as I've seen other cases where the user authenticated by RESTful was missing in other parts.

rbran100:
@amitaibu I like it, it *looks risky but the only way to handle complicated permissions without re-implementing them in full.  less risky than a ULI link in hand.

amitaibu:
`\RestfulAuthenticationTestCase::testSwitchUser` seems to pass locally for me (php 5.5). @RoySegall were you able to reproduce the fail on your local?

RoySegall:
Not that amount and types of failures.

amitaibu:
Ok, I was able to reproduce on once `\RestfulAuthenticationTestCase::testAuthentication` is also executed, so I'll check.

amitaibu:
Ok, I'm now reseting the `global $restful_switch_user;`

amitaibu:
Update: FInally a green test. Doing a few more tests to try and improve.

amitaibu:
@mateu-aguilo-bosch brave enough to have a look there ;)

amitaibu:
Another benefit of this PR, is that we can use RESTful auth to help us with downloading private files.

``` php
/**
 * Implements hook_restful_file_download_access().
 *
 * Get the user account from the RESTful authentication.
 */
function example_restful_file_download_access($file_item, $entity_type, $entity) {
  $auth_plugins = array_keys(restful_get_authentication_plugins());

  $auth_manager = new RestfulAuthenticationManager();

  // Register all authentication plugins.
  foreach ($auth_plugins as $auth_plugin_name) {
    $auth_manager->addAuthenticationProvider($auth_handler);
    $auth_manager->setIsOptional(TRUE);
  }

  $request = restful_parse_request();
  $auth_manager->getAccount($request);
}
```

So now an "anonymous" with an access token can have access to a file (e.g. `http://restful.dev/system/files/some-file.png?access_token=ovhsd-ntcdJHR8cPs95T1Becy1SKfyOsxGxgVrgrgWw`

In fact maybe we should add this piece of code into RESTful itself?

README.md
```diff
@@ -28,8 +28,8 @@ control.
 
 ## Module dependencies
 
-  * [Entity API](https://drupal.org/project/entity), with the following patch:
-  * [Prevent notice in entity_metadata_no_hook_node_access() when node is not saved](https://drupal.org/node/2086225#comment-8768373)
+  * [Entity API](https://drupal.org/project/entity), with the following patche:
```
e0ipso: _patche_ :arrow_forward: _patch_

modules/restful_token_auth/restful_token_auth.module
```diff
@@ -49,8 +49,16 @@ function restful_token_auth_restful_parse_request_alter(&$request) {
   $plugin = restful_get_authentication_plugin('token');
   $param_name = $plugin['options']['param_name'];
 
+  $value = \RestfulManager::getRequestHttpHeader($param_name);
+  if (!$value && !empty($_GET[$param_name])) {
```
e0ipso, line 53: Should we be using `$request[$param_name]` instead of `$_GET[$param_name]`?

restful.module
```diff
@@ -749,3 +768,21 @@ function restful_date_time_format_element_validate($element, &$form_state) {
...
   }
 }
+
+/**
+ * Determine if this is the first time we try to switch the user.
+ *
+ * @return bool
+ *   TRUE if this static function was already called.
+ */
+function restful_is_user_switched() {
```
e0ipso, line 778: This belongs to the `AuthenticationManager` class as a `public static` method.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -116,6 +126,74 @@ public function getAccount(array $request = array(), $method = \RestfulInterface
...
+      $this->setOriginalUserSession(array(
+        'user' => $user,
+        'session' => $session,
+      ));
+
+      // Don't allow a session to be saved. Provider that require a session to
+      // be saved, like the cookie provider, need to explicitly set
+      // drupal_save_session(TRUE).
+      // @see \RestfulUserLoginCookie::loginUser().
+      drupal_save_session(FALSE);
```
e0ipso, line 154: > Don't allow a session to be saved.

~~Why?~~ Nevermind, thanks for https://www.drupal.org/node/218104!

restful.module
```diff
@@ -472,15 +481,21 @@ function restful_menu_process_callback($resource_name, $version = NULL) {
     $handler->setHttpHeaders('Access-Control-Allow-Origin', $allowed_origin);
   }
 
-  $path = implode('/', $path);
-
   $method = strtoupper($_SERVER['REQUEST_METHOD']);
 
   if ($method == \RestfulInterface::POST && \RestfulManager::getRequestHttpHeader('X-HTTP-Method-Override')) {
-    $method = strtoupper(\RestfulManager::getRequestHttpHeader('X-HTTP-Method-Override'));
+    $method = \RestfulManager::getRequestHttpHeader('X-HTTP-Method-Override');
```
e0ipso, line 487: Why are you deleting `strtoupper`? There is nothing that ensures that the value of the `X-HTTP-Method-Override` HTTP header will be uppercase.

restful.module
```diff
@@ -426,6 +426,20 @@ function restful_menu_access_callback($resource_name, $version = NULL) {
     return;
   }
 
+  // Set the request and method on the handler, so access callbacks have full
+  // access to the request and account.
```
e0ipso, line 430: When is all this going to be used? Is it for `$handler->access()`?

plugins/restful/RestfulBase.php
```diff
@@ -815,7 +815,16 @@ public function process($path = '', array $request = array(), $method = \Restful
       $this->getRateLimitManager()->checkRateLimit($request);
     }
 
-    return $this->{$method_name}($path);
+    $return = $this->{$method_name}($path);
+
+    if (empty($request['__application']['rest_call'])) {
```
e0ipso, line 820: We are always switching the user when doing a `setAccount`, but we are only switching it back if the call is from cURL?

restful.module
amitaibu, line 778: the reason I did it here, is to prevent cases where multiple `AuthenticationManager`(s) are initialised. A little ugly, but feels more secure

restful.module
amitaibu, line 487: because two lines below we convert it to `strtolower` :)

restful.module
amitaibu, line 430: It's for the auth manager, to have the full context of the request

e0ipso:
I am still a bit wary about this, but I am willing to do a trust fall in this one :-)

@amitaibu some comments before merging this.

plugins/restful/RestfulBase.php
amitaibu, line 820: > but we are only switching it back if the call is from cURL?

No, we only switch back inside an API call. We don't do it on cURL as:
1) The user might have logged in using the cookie login
2) There's no need since the session isn't saved, and the request will terminate "soon".

amitaibu:
> I am still a bit wary about this

I can understand why ;) However, I think that without this the token auth isn't really useful/ hard to use

e0ipso:
> In fact maybe we should add this piece of code into RESTful itself?

Do you mean for the file resource? If so, I'm up for it but maybe in a follow up to reduce scope of this.

restful.module
e0ipso, line 778: I think that it's OK. You are not supposed to create multiple `AuthenticationManager`. If you do, you need to be aware of the insights of doing so.

restful.module
e0ipso, line 487: :astonished: … :ok:

restful.module
amitaibu, line 778: I agree in general to the statement, but for the sake of security, I think it would be ok to make the exception. If you feel strongly about it - I can change

restful.module
e0ipso, line 430: But you're setting them in the resource handler (https://github.com/RESTful-Drupal/restful/blob/token-get/restful.module#L411). As I understand, the `process` method will end up doing this anyways.

amitaibu, line 430: But `process` isn't being called during the `$handler->access()`

restful.module
e0ipso, line 778: What if we made `AuthenticationManager` to be a singleton? https://gist.github.com/mateu-aguilo-bosch/884b3c997f9fb3eef07c

That way we can be **sure** that there is only one `AuthenticationManager`.

restful.module
e0ipso, line 430: Right, then it is for `$handler->access()`. Sorry, I misunderstood your last comment. Carry on.

modules/restful_token_auth/restful_token_auth.module
amitaibu, line 53: No, because the $request might be the values from the `$_POST`, and we want to explicetly check also the `$_GET` for the presence of access_token

plugins/authentication/RestfulAuthenticationManager.php
amitaibu, line 154: good idea - I'll add the link to the comment

restful.module
amitaibu, line 778: I thought about, but wasn't sure if we wanted to do it in this PR. do we? :)

modules/restful_token_auth/restful_token_auth.module
e0ipso, line 53: Right! Sorry, I got confused by 2.x where we have them both separated.

restful.module
e0ipso, line 778: Fair enough. Let's create a follow up and deal with that later.

amitaibu:
Merged. What can possibly go wrong :wink:

rbran100:
I felt like something really important happened 30 minutes ago, I guess this was it.  I will need to use this and cant wait.

PR #415, opened 2015-02-19
414 check to see if additional hateoas function has set _links before declaring array()

### LINKED ISSUE #414: hal _links supplied by additionalHateoas() in hal JSON formatter is overwritten by addHateoas
Handlers are allowed to supply additional Hateoas elements via the additionalHateoas() function in ln 47 of RestfulFormatterHalJson.class.php however later when $this->addHateoas is called it overwrites any previously set "_links".

thebruce:
This PR adds code to check to see if $data['_links'] has already been set by the additionalHateoas function in the JSON formatter.

closes #414 
- checks to see if $data['_links'] has been set
- checks to see that previously set $data['_links'] is the type we expect
- allows addtionalHateoas function defined at the handler level to contribute _links

### DIFF
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -77,7 +77,9 @@ class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulF
     }
     $request = $this->handler->getRequest();
 
-    $data['_links'] = array();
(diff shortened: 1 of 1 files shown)


thebruce:
Yeah, was over thinking and added in the nonsense is_array check. Oops did i say "over" thinking I meant "under" thinking.

e0ipso:
Hahaha! I love you sense of humor.

Thanks for contributing David! This has been merged.

Issue #413, opened 2015-02-18
Avoid SA-CONTRIB-2015-052 - RESTful Web Services - Access Bypass


amitaibu:
https://www.drupal.org/node/2428863

Here's the commit that deals with it - http://cgit.drupalcode.org/restws/commit/?id=01724e0

(cc @klausi)

PR #416, opened 2015-02-19
WIP: Avoid SA-CONTRIB-2015-052 - RESTful Web Services - Access Bypass

amitaibu:
#413

Currently just a simple copy/ paste form the [original](http://cgit.drupalcode.org/restws/commit/?id=01724e0) restws commit.

### DIFF
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -102,6 +102,11 @@ class RestfulAuthenticationManager extends \ArrayObject {
         $account = $user->uid ? user_load($user->uid) : $account;
       }
     }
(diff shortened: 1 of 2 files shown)


amitaibu:
Actually, our cache layer is completely different. So we should probably just keep the test

e0ipso:
Have you checked `\RestfulManager::pageFooter()`? (talking from memory).

amitaibu:
I'll first try to reproduce the RESTws bug with RESTful, and then we can also add a simpleTest just to be extra sure we won't get the same "cache pollution" issue.

amitaibu:
This PR should probably include only the tests part - so we validate we don't have this cache poisoning.

PR #418, opened 2015-02-23
2.x Decorated resources.

e0ipso:
@amitaibu this PR will focus on creating the decorated resources. These will act as a regular resource, but they're are a fairly dumb shell/wrapper around a regular resource. This wrapper will add a small piece of functionality:
- Authentication.
- Rate Limits.
- Cache.

This means that any resource that we implement can get all the features for free.

A special word about cached resources. The caching functionality belongs to the data provider, since it will skip or not the data loading. All you need to do to have your data provider to allow caching is implement `getContext`. That will return a keyed array that will be used to build the _weird_ cache id for smarter cache invalidation. Finally, the caching feature is bubbled up to the `Resource` layer.

All these objects will –not implemented yet– be instantiated (or not) based on the plugin configuration. If there is presence of `rateLimit`, then instead of returning the newly created `$resource`, return `new RateLimitedResource($resource)` (which is a valid `ResourceInterface`).


amitaibu:
Haven't forgotten about this - I'm going over it..

e0ipso:
@amitaibu awesome. Thanks!

amitaibu:
Ok, lets get this in :)

e0ipso:
Hopefully this is going to start to make sense in no time soon.

[image: http://]
Mateu Aguiló Bosch
[image: http://]about.me/e0ipso
http://about.me/e0ipso

PR #419, opened 2015-03-02
2.x entity data provider (take 3)

e0ipso:
This PR is built on top of #418. This is [the delta](https://github.com/RESTful-Drupal/restful/compare/mateu-aguilo-bosch:2.x-entity-data-provider-2...mateu-aguilo-bosch:2.x-entity-data-provider-3?expand=1).

This finalizes the view method for the entity data provider. It also introduces the `DataProviderResource`, that it will take a resource and use it to fetch data.

Next step will be having the code for the rest of methods. That will be (hopefully) a single PR and a simpler job.


amitaibu:
@mateu-aguilo-bosch can you please re-roll this with latest 2.x

e0ipso:
@amitaibu done.

This can be reviewed now. This contains CRUD for entities w00t!

After this, the next step will be providing an example resource. That should help putting many pieces in place. And the real testing will begin!

amitaibu:
Now that we have #432 , lets get this in, as it will make it easier to follow the code.

Issue #411, opened 2015-02-17
Wrong count on un-accessible nodes


amitaibu:
When anon does api/articles but has no permission to "view published content", no data appears but the count is wrong

![postman-2](https://cloud.githubusercontent.com/assets/125707/6236291/08a56aa8-b6f2-11e4-9db6-246b9219c046.png)

rw355:
RestfulDataProvider::getTotalCount is by-passing the access check. Or, the access check should be implemented at RestfulFormatter::prepare

amitaibu:
I think it would be best to make sure getTotalCount() is using an access check, similar to what we do in `\RestfulDataProviderEFQ::addExtraInfoToQuery`

PR #420, opened 2015-03-02
Wrong count on un-accessible nodes

rw355:
#411

Fix by judging the returned $data. I don't think it's necessary to run a checkEntityAccess again at formatter. Better to have the checkEntityAccess result passed into formatter, then we don't rely on $data.


e0ipso:
Re-rolled in #586.

PR #422, opened 2015-03-05
355 bad hal (remix)

### LINKED ISSUE #355: User resource not appearing in node JSON
I am getting no data returned when using this code, even from `restful_example`:

``` php
      'user' => array(
        'property' => 'author',
        'resource' => array(
          // The bundle of the entity.
          'user' => array(
            // The name of the resource to map to.
            'name' => 'users',
            // Determines if the entire resource should appear, or only the ID.
            'full_view' => TRUE,
          ),
        ),
```

Switching to `'full_view' => FALSE` makes no difference.

Similarly I'm not getting anything with an `entityreference` field where I've defined a resource for the referenced entity (the resource works on its own). Accessing the user resource directly also works.

I've tried lots of different approaches and read the documentation every which way.  I've tried creating content with other users (just in case).

If I remove the 'resource' part I get the entire user (or other entity) returned as expected.

If I use:

``` php
        'resource' => array(
          'users' => 'users',
        ),
```

I get `user: null` (no other variant of this generates anything).

There are no errors or other messages in the error log.

I was originally using the standard release but then switched to the dev release but it made no difference.

I'm feeling pretty stupid (even though I've been a pro Drupal dev for 8 years) so any guidance would be appreciated. (I could use a callback but that would be defeating the object.)

thebruce:
Fixes #355 
Closes #367 

This PR picks up on the awesome work that @adaddinsane did to help the hal_json formatter output _embedded resources for invidual items within list requests.  That PR #367 has been open since January 15th and I definitely would like to see the work make it to the main repo.

I've kept a lot of the work @adaddinsane including working with an embedded array within the prepareRow call and removing the final function (moveMetadataResource).

What i've done is tested for whether or not this call is a list request. That's what is going to either put embedded resources into the property level or allow them to float up the resource root in a single item request.

So for listings:

```
hal:thing [
    {
        property:"",
        _embedded {specific embeds for this thing},
       _links
    },
    {
        property:"",
        _embedded {specific embeds for this thing},
       _links
    }
],
_embedded {listing embedded probably empty},
_links
```

For Single Items:

```
hal:thing [
    {
        property:"",
    }
],
_embedded {specific embeds for this thing},
_links
```

There are a few code styling things from Drupal phpcs but nothing shocking there.

### DIFF
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -8,17 +8,6 @@
 class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulFormatterInterface {
 
   /**
-   * Determine whether this is an associative array.
-   *
-   * @param array $array
-   *
-   * @return boolean
-   */
-  static function is_assoc(array $array) {
-    return (bool)count(array_filter(array_keys($array), 'is_string'));
-  }
-
-  /**
    * Content Type
    *
    * @var string
@@ -36,13 +25,15 @@ class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulF
       return $data;
     }
     // Here we get the data after calling the backend storage for the resources.
+    $curies_resource = $this->withCurie($this->handler->getResourceName());
+
+    $output = array();
 
     foreach ($data as &$row) {
-      $row = $this->prepareRow($row);
+      $row = $this->prepareRow($row, $output);
     }
 
-    $curies_resource = $this->withCurie($this->handler->getResourceName());
-    $output = array($curies_resource => $data);
+    $output[$curies_resource] = $data;
 
     if (!empty($this->handler)) {
       if (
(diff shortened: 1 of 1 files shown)


thebruce:
@mateu-aguilo-bosch - Local simpletests pass for this code however Travis is failing before it ever gets to thes simpletests on this entity wrapper patch

$ patch -p1 /home/travis/build/restful/drupal/sites/all/modules/entity/includes/entity.wrapper.inc < 2264079-entity-wrapper-access-single-entity-reference-2.patch

patching file /home/travis/build/restful/drupal/sites/all/modules/entity/includes/entity.wrapper.inc

Reversed (or previously applied) patch detected! Assume -R? [n]

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated

amitaibu:
Test seems to fail.

thebruce:
Weird. Both Hal JSON tests were coming back fine locally - now this has #423 in it which was causing it to die in Travis, so i'll at least be able to watch this there and see what's happening.

thebruce:
OK, everything came back green - I think it was just #423 where it was hanging.

amitaibu:
Ok, great. I think we should also add some tests, to assert the functionality and prevent regressions (also the tests help with the review, since there's lots going on in the HAL formatter..)

thebruce:
Yeah, that makes sense. I should have some time next week, i think.

e0ipso:
@thebruce, should we merge this and have create a follow up for the tests?

thebruce:
@mateu-aguilo-bosch - yeah, i don't mind writing them, but it might be a week or two and it'd be nice to get this working correctly.

e0ipso:
Merged.

Thanks for contributing!!

Issue #429, opened 2015-03-10
How to get uid on token login?


uniquename:
Maybe a stupid question, cause i missed something... 

Getting the access_token with basic auth works just fine, but how can I receive user data or uid to get it from the users resource, after I got the token?

...btw. fantastic module!-)

amitaibu:
You mean `$this->getAccount()` ?

uniquename:
Sorry, maybe I wasn't clear enough... when I send a GET with basic auth credentials to api/login-token I receive the token and infos about it, as defined in RestfulAccessTokenAuthentication.class.php line 61-63.
Lets say, I have an app that logs in a user this way, after a successfulI login I want to get some information about the logged in user, but therefore I would need the uid to query api/user. So I was thinking that api/login-token should also give the uid, or to avoid another request, some more information about the user.

amitaibu:
I see two options:

Add `uid` to login-token, or add a `/api/me` resource

uniquename:
hehe... just thought the same about api/me, but it would be another request. I think I will go for more information on api/login-token. I could make a pull request later, if this is desired for the module design...

amitaibu:
How about providing both options in your PR, and wining some extra Karma :wink:

uniquename:
...k, i'll go for that;-)

e0ipso:
Closed in favor of https://github.com/RESTful-Drupal/restful/pull/430

jeff-h:
For anyone stumbling across this issue, there's now a recipe in the wiki for creating a `user/me` resource.

PR #430, opened 2015-03-11
optional provide info about token owner

uniquename:
like discussed here #429

### DIFF
--- a/modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/access_token/1.0/RestfulAccessTokenAuthentication.class.php
+++ b/modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/access_token/1.0/RestfulAccessTokenAuthentication.class.php
@@ -59,7 +59,21 @@ class RestfulAccessTokenAuthentication extends \RestfulTokenAuthenticationBase {
 
     $output = $this->viewEntity($id);
 
+    // various options to provide info about the token owner
+    $owner_resource = variable_get('restful_token_auth_owner_resource', '');
+
+    if ($owner_resource !== '<none>') {
+      if ($owner_resource == '<id>') {
+        $output['owner'] = $access_token->uid;
+      } else {
+        $handler = restful_get_restful_handler('users') ?: restful_get_restful_handler($owner_resource) ?: FALSE;
+        if ($handler) {
+          $user = $handler->get($access_token->uid);
+          $output['owner'] = array_pop($user);
+        }
+      }
+    }
+
     return $output;
   }
-
 }
--- a/modules/restful_token_auth/restful_token_auth.admin.inc
+++ b/modules/restful_token_auth/restful_token_auth.admin.inc
@@ -24,6 +24,12 @@ function restful_token_auth_admin_settings($form, &$form_state) {
     '#default_value' => variable_get('restful_token_auth_expiration_period', 'P1D'),
     '#element_validate' => array('restful_date_time_format_element_validate'),
   );
+  $form['advanced']['restful_token_auth_owner_resource'] = array(
+    '#type' => 'textfield',
+    '#title' => t('Owner resource'),
+    '#description' => t('If a resource endpoint is specified or the resource "users" exists the token endpoint will also provide information about the token owner. Use !none if you have the users resource enabled but no information should be send. Use !id to just provide the owner uid wihout querying any resource.', array('!none' => '&lt;none&gt;', '!id' => '&lt;id&gt;')),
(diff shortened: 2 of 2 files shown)


e0ipso:
> just thought the same about api/me, but it would be another request. I think I will go for more information on api/login-token.

I think that making an extra request is fine in this case. We don't want to:
- Put in all the data we can in a very basic endpoint. Other use cases may not need all the user data back from the token endpoint.
- Complicate with even more configuration options.

I would just add the `uid` in the token endpoint.

modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/access_token/1.0/RestfulAccessTokenAuthentication.class.php
```diff
@@ -59,7 +59,21 @@ public function getOrCreateToken() {
 
     $output = $this->viewEntity($id);
 
+    // various options to provide info about the token owner
+    $owner_resource = variable_get('restful_token_auth_owner_resource', '');
+
+    if ($owner_resource !== '<none>') {
+      if ($owner_resource == '<id>') {
+        $output['owner'] = $access_token->uid;
```
e0ipso, line 67: Just keep this line.

uniquename:
I'm using this module with a mobile app, thats why I'm like concerned about extra requests and I had thoughts like, why sending the owner id if there is no endpoint to get more information... that's why I came up with all that configuration... for me line 70 is the solution.

e0ipso:
Thanks for getting back @uniquename. I am concerned about the module itself.

I understand, and share, your concern in your particular use case.

My recommendation would be to:
- Update this PR to make it simpler :smile: 
- Create `v2` (in a custom module) of the token endpoint. In that I would inherit from `RestfulAccessTokenAuthentication` and overwrite `getOrCreateToken` to inject your logic.
- Write a recipe of how you did that in the recipe book :-P (https://www.drupal.org/node/2380679)

I know I'm asking a lot, but I think that it's a good think to spread Open Source knowledge.

Thanks for your efforts!!

uniquename:
Ok, I will see when I find time for that. Specially writing a recipe about it... cause that section really needs some love ;-)

modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/access_token/1.0/RestfulAccessTokenAuthentication.class.php
amitaibu, line 67: or even better, we can declare the owner in the `\RestfulTokenAuthenticationBase::publicFieldsInfo`, probably something like this:

``` php
    if (variable_get('restful_enable_users_resource', TRUE)) {
      $public_fields['owner'] = array(
        'property' => 'uid',
        'resource' => array(
          'user' => array(
            'name' => 'users',
            'full_view' => FALSE,
          ),
        ),
      );
    }
    else {
      $public_fields['owner'] = array(
        'property' => 'uid',
      );
    }
```

uniquename:
Ok, kept on thinking about this for a while... and: just close the pull request ;-) I see just too many ifs and thens for different use cases... one other thought, shouldn't it be possible to enable/disable the csrf token endpoint? For mobile apps it's just not needed. And actually the same for the token endpoints, so one could enable the module to have the classes without having the endpoint.

e0ipso:
@uniquename you can disable the endpoint if you set `hook_menu = FALSE` in the plugin definition and you don't provide a menu item. That will leave the endpoint available to use programatically but not hittable from the exterior world.

I believe that there must be a `hook_somethig_ctools_something_alter` to alter the plugin definition without having to hack `restful_token_auth`.

That would also be a great recipe :wink: wink.

uniquename:
Then I would go for closing the endpoint with the htaccess ;-) ...sorry, joking... but why should it go this way for the token endpoint, when user and cookie endpoint are configurable in the backend? Like the comment in the your code says "By checking the variable, we allow implementing modules to easily implement their own "users" resource."

modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/access_token/1.0/RestfulAccessTokenAuthentication.class.php
e0ipso, line 67: @uniquename does this provide the info that you need?

e0ipso:
I'm doing triage. @uniquename I will close this issue since it seems to be stalled.

Do not hesitate to re-open it.

Thanks for contributing!

PR #433, opened 2015-03-14
2.x Data Sources

e0ipso:
Depends on #432.

This PR introduces data sources. Data sources are the objects that the DataProviders use to interact with the data store (in our case the DB, the file system, etc.). These are mainly introduced for consistency in the `callback` signature (see `getEntitySelf()`).

Also the ResourceFields are now able to get the value for the field they represent! Yay for encapsulation!


src/Annotation/Resource.php
```diff
@@ -0,0 +1,162 @@
...
+   *
+   * 'string' => 'foo',
+   * 'operator' => 'STARTS_WITH',
+   *
+   * @var array
+   */
+  public $autocomplete = array();
+
+  /**
+   * Options. Used mainly for file resources.
```
amitaibu: Maybe `Arbitrary options that can be passed to the resource`

src/Annotation/Resource.php
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Access control using the HTTP Access-Control-Allow-Origin header.
+   *
+   * @var string
+   */
+  public $allowOrigin;
+
+  /**
+   * Discoverable.
```
amitaibu: `Determines if a resource should be discoverable, and appear under /api.`

src/Annotation/Resource.php
```diff
@@ -0,0 +1,162 @@
...
+   * @var bool
+   */
+  public $discoverable = FALSE;
+
+  /**
+   * URL parameters.
+   *
+   * @var array
+   */
+  public $urlParams = array();
```
amitaibu, line 152: what is this used for?

src/Annotation/Resource.php
```diff
@@ -0,0 +1,162 @@
...
+ */
+
+namespace Drupal\restful\Annotation;
+
+use Drupal\Component\Annotation\Plugin;
+
+/**
+ * Defines a Resource annotation object.
+ *
+ * @ingroup plug_example_api
```
amitaibu: is this correct?

src/Annotation/Resource.php
```diff
@@ -0,0 +1,162 @@
...
+   *
+   * @var array
+   */
+  public $urlParams = array();
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getId() {
+    // Our ID property is 'name', not 'id'.
```
amitaibu: Maybe `// The ID of the resource plugin is its name.`

src/Formatter/FormatterManager.php
```diff
@@ -23,9 +24,18 @@ class FormatterManager implements FormatterManagerInterface {
   protected $plugins;
 
   /**
+   * The resource.
+   *
+   * @todo: Remove this coupling.
```
amitaibu, line 29: why? I think it would be ok to have access to the resource from the formatter, no?

src/Http/Request.php
```diff
@@ -351,9 +351,9 @@ protected static function parseHeadersFromGlobals() {
   /**
    * {@inheritdoc}
    */
-  public function getPath() {
+  public function getPath($strip = TRUE) {
```
amitaibu, line 354: what's the use case for strip=false?

src/Plugin/formatter/Formatter.php
```diff
@@ -41,4 +44,20 @@ public function getContentTypeHeader() {
...
+    if (isset($this->resource)) {
+      return $this->resource;
+    }
+
+    // Get the resource from the instance configuration.
+    $instance_configuration = $this->getConfiguration();
+    if (empty($instance_configuration['resource'])) {
+      return NULL;
+    }
+    return $instance_configuration['resource'] instanceof ResourceInterface ? $instance_configuration['resource'] : NULL;
```
amitaibu, line 60: this part is not clear.

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -198,29 +199,13 @@ public function view($identifier) {
...
-        $resource_data_provider = DataProviderResource::init($this->getRequest(), $resource['name'], array(
-          $resource['major_version'],
-          $resource['minor_version'],
-        ));
-        $value = $resource_data_provider->view($identifier);
-      }
-      else {
-        $value = $this->getValue($wrapper, $resource_field);
-      }
+      $source = new DataSourceEMW($this->getAccount(), $wrapper);
```
amitaibu: I'm not sold on the `DataSource` name in general, in fact `DataSource` is probably a better name then `DataProvider`. 

Maybe we  `DataInterpreter` (not sure about it as-well)?

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -484,15 +469,22 @@ protected function getEntityInfo($type = NULL) {
...
   protected function getQueryForList() {
     $query = $this->getEntityFieldQuery();
-    if ($path = $this->request->getPath()) {
-      $ids = explode(',', $path);
-      if (!empty($ids)) {
-        $query->entityCondition('entity_id', $ids, 'IN');
-      }
+
+    // If we are trying to filter or sort on a computed field, just ignore it
+    // and log an exception.
```
amitaibu, line 472: why ignore, and not throw an error?

src/Plugin/resource/DataSource/DataSourceInterface.php
```diff
@@ -0,0 +1,31 @@
...
+   * @return object
+   *   The fully loaded account.
+   */
+  public function getAccount();
+
+  /**
+   * Get the wrapper.
+   *
+   * @return mixed
+   *   The entity metadata wrapper describing the entity. Every data source will
```
amitaibu: `The entity metadata wrapper describing the entity` is just an example of a wrapper, so I think we can remove this line.

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -140,6 +145,183 @@ class_implements($class_name)
   /**
    * {@inheritdoc}
    */
+  public function value(DataSourceInterface $source) {
+    // Return if this field is a callback.
```
amitaibu: not clear - if this is a callback then why is it getting into `ResourceFieldEntity`?

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -140,6 +145,183 @@ class_implements($class_name)
...
+      // Now it does not need to be keyed by bundle since you don't even need
+      // an entity to use the resource based field.
+
+      $request = Request::create('', array(), Request::METHOD_GET);
+      $resource_data_provider = DataProviderResource::init($request, $resource['name'], array(
+        $resource['majorVersion'],
+        $resource['minorVersion'],
+      ));
+      // FIXME: $embedded_identifier needs to be fetched first!!!
+      return $resource_data_provider->view($embedded_identifier);
```
amitaibu, line 165: To se if I understand correctly - embedded_identifier should be the ID for the sub-resource? e.g. `$resource->getIdentifier`?

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -140,6 +145,183 @@ class_implements($class_name)
...
+   * @throws \EntityMetadataWrapperException
+   */
+  public function access($op, DataSourceInterface $source) {
+    // Perform basic access checks.
+    if (!$this->decorated->access($op, $source)) {
+      return FALSE;
+    }
+
+    if (!$this->getProperty()) {
+      // If there is no property we cannot check for property access.
```
amitaibu, line 204: example?

modules/restful_example/src/Plugin/resource/Articles__1_0.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+namespace Drupal\restful_example\Plugin\resource;
+
+use Drupal\restful\Plugin\resource\ResourceEntity;
+use Drupal\restful\Plugin\resource\ResourceInterface;
+
+/**
+ * Class Articles
+ * @package Drupal\restful\Plugin\formatter
```
amitaibu: formatter?

modules/restful_example/src/Plugin/resource/Articles__1_0.php
```diff
@@ -0,0 +1,55 @@
...
+ * @Resource(
+ *   name = "articles:1.0",
+ *   resource = "articles",
+ *   label = "Articles",
+ *   description = "Export the article content type with cookie authentication.",
+ *   authenticationTypes = TRUE,
+ *   authenticationOptional = TRUE,
+ *   dataProvider = {
+ *     "entityType": "node",
+ *     "bundles": {
```
amitaibu, line 26: Lets discuss this shift from single bundle by default to multiple in regards to `create` - do you have an idea how to deal with this (I saw you had a todo about it in one of the PRs)

amitaibu:
btw, I tried to enable this on my local, but on clear cache I get

```
WD menu: Drupal\restful\Exception\BadRequestException: Unrecognized  [error]
HTTP method. ( in Drupal\restful\Http\Request::create() (line 180 of
/.../sites/all/modules/restful/src/Http/Request.php).
```

amitaibu:
> but on clear cache I get

Actually it happened only on `drush cc all` -- worked fine when clear cached from UI.

amitaibu:
localhost_d7_dev_api_articles

This "small" JSON is simply amazing. Crazy amount of great work @mateu-aguilo-bosch !!

amitaibu:
Added few comments - the next part which is still not totally clear to me which I'd love to see a code example is the decoration of the resources (e.g. rate limit and cache).

Let me know when you want the Gizra team to jump and help with the tests :)

src/Annotation/Resource.php
e0ipso: :+1:

src/Annotation/Resource.php
e0ipso: :+1:

src/Annotation/Resource.php
e0ipso, line 152: This is used to disable the query string parameters that control:
- The sort
- The filter by field
- The field listing
- The loadById

In case that this field is:

``` php
array(
  'sort' => FALSE,
);
```

Then hitting `api/articles?sort=label` will result in an exception.

e0ipso, line 152: Also, this is present in 1.x as `ulr_params` in the CTools plugin.

src/Annotation/Resource.php
e0ipso: Nope. Good catch.

src/Annotation/Resource.php
e0ipso: :+1:

src/Formatter/FormatterManager.php
e0ipso, line 29: I feel that the resource plugin should deliver the renderable data to the formatter. Then the formatter only has to create a return string in the right format.

If the formatter needs further information for complex formatting, then we should provide a context object of some sort, with that information, without coupling both objects. My argument for that is that the resource is much more than the context for formatting output, therefore it seems an unnecessary coupling between the objects and an overkill.

Anyways, as you see it's just a TODO. And not the most important one.

src/Http/Request.php
e0ipso, line 354: We use the `menu_router` table to detect the resource name based on a URL. In that table we have the URLs with like `api/v1.0/articles`, `api/v1/articles` and `api/articles`. If we remove `api/v1.0` from the beginning then we won't match any row in that table when querying.

We cannot derive the resource name by just exploring the URL, since a plugin may declare `menu_item = foo/bar/resource/big` where none of its parts is the name of the resource.

src/Plugin/formatter/Formatter.php
e0ipso, line 60: This is just type checking. If you use `setResource('ha!')` then `getResource()` will yield `NULL` and not `ha!`. This can go away and let the implementor mess with their own code.

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso: `DataInterpreter` sounds good to me.

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso, line 472: We are providing sort by ID by default, but, for entities, ID is a computed field (`getIdentifier()`). I feel that it is a very common case to have computed fields.

I'm not too sold on this either, but I don't see a clear exit…

e0ipso, line 472: Also, now I realize that we need a `isComputed()` method on the `ResourceFieldBase` class.

src/Plugin/resource/DataSource/DataSourceInterface.php
e0ipso: :+1:

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso: I'm just waterproofing our code. Imagine some newbie that has one of the keys on the plugin that decide it to be `ResourceFieldEntity` (like `sub_wrapper`), but at the same time they have a `callback`. The ideal outcome is that we return the callback.

``` php
/**
  * @Resource
  * "callback" = "getStuff",
  * "sub_property" = "this_is_a_leftover",
  * …
  */
```

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 165: Yes! This is solved in the next PR.

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 204: Meaning that we can only check for field access or property access for things like `$file->filemime` or `$node->field_tags`. We denote those by using the `'property'` property in the plugin definition.

modules/restful_example/src/Plugin/resource/Articles__1_0.php
e0ipso: Wrong. It's a copy & pasta error :spaghetti: Good catch.

src/Plugin/resource/Field/ResourceFieldEntity.php
amitaibu: > The ideal outcome is that we return the callback.

I think it would be best to throw an error with a good explanation, to educate the poor developer ;)

modules/restful_example/src/Plugin/resource/Articles__1_0.php
e0ipso, line 26: Yeah, that's the trickiest part. I have two ideas:
1. If there is more than one different bundle, then `create` is disabled.
2. We define some extra predefined key that allows you to select the appropriate bundle.

I like both options, but I see the difficulties. Let's see what you think of this.

Maybe it's worth mentioning that when posting to a resource that has embedded resources (ex: a node with tags) we are already allowing a special `'request'` key to provide info for the sub-request when posting the tags. That may be some precedence in favor of 2. Something like sending a payload like:

``` php
// PUT request.
array(
  'title' => 'Article',
  'body' => 'Lorem ipsum',
  'tags' => array(
    'request' => array('method' => 'POST', …),
    'values' => array('alpha', 'beta'),
  ),
);
```

e0ipso:
> the next part which is still not totally clear to me

I would like to make sure all of this is working:
- [x] Listing Field API fields.
- [x] Listing entity properties.
- [x] Listing embedded resources.
- [x] Listing reference fields with their ID (either by not using `'resource'` or using it with `'full_view' = FALSE`.
- [x] Using process callbacks.
- [x] Using field access callbacks.
- [x] Using method based field access (disabling fields for the GET method). This is the new alternative for `create_or_update_passthrough`.
- [x] Filtering by property/field.
- [x] Hiding fields from output via url.
- [x] Sorting  by property/field.
- [x] Disabling url params.
- [x] Showing fields by using a field formatter.
- [x] Showing an entity with a view mode.

Basically, making sure that we can read from an entity based endpoint with full features.

modules/restful_example/src/Plugin/resource/Articles__1_0.php
amitaibu, line 26: I'm in favor of 2 as-well.

amitaibu, line 26: and indeed, if no extra info is passed in the request, we should error out with a clear message

modules/restful_example/src/Plugin/resource/Articles__1_0.php
```diff
@@ -0,0 +1,55 @@
...
+ *       "article"
+ *     },
+ *   },
+ *   majorVersion = 1,
+ *   minorVersion = 0
```
amitaibu, line 31: btw, is this mandatory, or can it be omitted?

e0ipso:
> Let me know when you want the Gizra team to jump and help with the tests :)

That would be awesome. At some point I think that we should port the same examples from 1.x to 2.x and have the same / similar tests. That way we accomplish:
- We have a ground for comparison between 1.x and 2.x.
- Devs know how to port their resources because they have examples.
- We are familiar with the existing examples.
- We can reuse a lot of the testing logic.

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso: That works for me.

:+1:

modules/restful_example/src/Plugin/resource/Articles__1_0.php
e0ipso, line 31: I see where you are going. This is redundant if we are using `name = "articles:1.0",`.

I'd like to remove one or the other in the future. Any preferences?

modules/restful_example/src/Plugin/resource/Articles__1_0.php
e0ipso, line 26: Cool! I'm leaving TODOs scattered with the new things that we will need to document. This shall be one of them too.

src/Plugin/resource/DataSource/DataSourceInterface.php
e0ipso: Actually, on second thought I think that being verbose in the comment providing some more context is a good thing.

modules/restful_example/src/Plugin/resource/Articles__1_0.php
amitaibu, line 31: > Any preferences?

Keep the majorVersion/ minorVersion -- so we can keep doing `api/sort=-majorVersion`

e0ipso:
@amitaibu implemented feedback.

amitaibu:
Looks good

Issue #435, opened 2015-03-16
README.md example does not work


iStryker:
I follow 'Declaring a REST Endpoint' in the README.md, and I get 

RestfulServiceUnavailable: Restful plugin class (articles) was not found. in restful_get_restful_handler_by_name() (line 303 of /var/www/mysite/sites/all/modules/contrib/restful/restful.module).

Examples inside restful/modules work.

iStryker:
The name of the file must match the name of the plugin.  If it does not, then you will get the RestfulServiceUnavailable plugin.   

Took me 4 hours to figure this out.  It was not obvious as the myplugin.inc still gets read by ctools.

NorthWindH:
Thank you for posting this as it helped me figure my stuff out. Will you be submitting a PR to the project with regard to the README?

raphaeltraviss:
Just to clarify, the example in the README does work, but it's not clear what some of the parameters in `$plugin` do.  I think some clarification in the docs would do some good.  I've commented on the pull request.

mikemccaffrey:
I was seeing this error when trying to create a new api endpoint, even when I am sure that the 'name' attribute in the plugin declaration matches the name of the file (minus the .inc).

Turns out that the even though the class name in the plugin settings was the same as the class filename, I forgot to change the class name within the class.php file. If you are getting this error, make sure the class names match up in all three places.

PR #436, opened 2015-03-16
Issue #435: README.md exmaple does not work

iStryker:


e0ipso:
Many thanks for the contribution!

README.md
```diff
@@ -63,12 +63,12 @@ dependencies[] = restful
  }
 ```
 
-####restful\_custom/plugins/restful/myplugin.inc
+####restful\_custom/plugins/restful/thenameoftheplugin.inc
```
amitaibu, line 66: maybe just change this to `restful\_custom/plugins/restful/articles.inc`, and keep the comment about the plugin name == the `name` property?

iStryker, line 66: I was thinking along the same lines, however articles.inc is a little bit confusing with 'resource' and 'name' being the same.  It doe not emphasize that the name of the file has to be the same as the name of the plugin.   

Maybe changing it to articles.inc and having 'The name of your plugin must match your file' comment I added is good enough for some. Personally, I would not notice it right away, however once I received the thrown error, I would re-read the instruction and noticed the comment right away. 

If you are not happy with 'thenameoftheplugin' you could change it to your liking, like 'plugin_name'

amitaibu:
Seems #446 already took care of this. Re-open if you think it's needed.

Issue #376, opened 2015-01-18
RESTful service with field collection


rivkitzipory:
Hi
I am using RESTFUL Module in drupal 7 And I try to create Service with Content type and field collection (nodes and field collection).
I can create a Service with the field collection, but when i try to add to the service with the content type thos rows:

$public_fields['my_field_collection'] = array(
      'property' => 'my_field_collection',
      'resource' => array(
        'my_resource' => 'my_resource',
      ),
    );

I get that error:
Field @property is not an entity reference or taxonomy reference field

What's the solution?

amitaibu:
> What's the solution?

RESTful doesn't know field collection yet. You could provide a simple PR that will fix it -- by getting `\RestfulEntityBase::getTargetTypeFromEntityReference` to recognise that type of field.

rivkitzipory:
Hi
I tried this solution (allready)
I Created class myRestfulEntityBase and i did the the following changes:

1 Add in `RestfulEntityBase::getTargetTypeFromEntityReference`:

``` php
elseif ($field['type'] == 'field_collection') {                    
        return 'field_collection_item';
      }
```

2 In `propertyValuesPreprocess`:

``` php
case 'field_collection':
        return $this->propertyValuesPreprocessReference($property_name, $value, $field_info, $public_field_name);
```

3 In `formSchemaHasAllowedValues`:

``` php
$field_types = array(
      'entityreference',
      'taxonomy_term_reference',
       'field_collection'
    );
```

Now, I get my service:

But I think there is missing method like setPropertyValues in RestfulEntityBaseTaxonomyTerm.

e0ipso:
FWIW you should only have in the `myRestfulEntityBase` class the methods that are different from the parent `RestfulEntityBase`. If not, any possible future bugfixes in `RestfulEntityBase` won't be applied automatically to your resource.

rivkitzipory:
Obviously, this is only in the development

e0ipso:
@rivkitzipory sorry for stating the obvious, I hope you know didn't mean to offend :-P

amitaibu:
@rivkitzipory after the code changes in https://github.com/RESTful-Drupal/restful/issues/376#issuecomment-70403578  what error are you getting?

rivkitzipory:
I get a Service like this:
{
id: 1,
label: "aaa",
field_collection: [
null,
null,
null,
null
]
},
{
id: 2,
label: "bbb",
field_collection: [
null,
null,
null
[...]

PR #438, opened 2015-03-18
Support for field collections

kzoltan:
Based on the #376 issue, the solution suggested by @rivkitzipory


e0ipso:
This looks good to me.

kzoltan:
Does this mean that its now in there in the dev version of the drupal.org module?
What is your strategy on the releases? When will this be part of a stable release?

Is there somewhere a documentation about this just that I didn't read it?

e0ipso:
Should be there in no time ☺

lukewertz:
It looks like this PR merged some, but not all of the code necessary to support field collections. It looks like @rivkitzipory added a file here: https://github.com/rivkitzipory/restful/commit/96ed9ea0f82505dcb95eeee0016753f5c19fa178

Should that not have been included in the PR?

PR #439, opened 2015-03-18
2.x writing entities

e0ipso:
@amitaibu this PR depends on https://github.com/RESTful-Drupal/restful/pull/437.

Like with the read, this only checks that POST can do basic posting. With this code you can do:

```
curl -i -X POST \
   -H "X-API-Version:1.0" \
   -H "Accept:application/json" \
   -H "Content-Type:application/json" \
   -d \
'{
  "label": "POST title 2!",
  "body": "POST body 2"
}' \
 'http://restful.local/api/articles'
```

And that will create an article with a title and a body.

Next steps will be making sure embedded posting works, that we can upload files, … Basically making sure all the advanced functionalities are in place.

### DIFF
--- a/modules/restful_example/src/Plugin/resource/Articles__1_0.php
+++ b/modules/restful_example/src/Plugin/resource/Articles__1_0.php
@@ -49,6 +49,14 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
       'self' => array(
         'callback' => array($this, 'getEntitySelf'),
       ),
+      'tags' => array(
+        'property' => 'field_tags',
+        'resource' => array(
+          'name' => 'tags',
+          'majorVersion' => 1,
+          'minorVersion' => 0,
+        ),
+      ),
     );
   }
 
--- a/modules/restful_example/src/Plugin/resource/Articles__1_0.php
+++ b/modules/restful_example/src/Plugin/resource/Articles__1_0.php
@@ -64,6 +64,15 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
         'property' => 'status',
         'methods' => array(Request::METHOD_POST, Request::METHOD_PUT),
       ),
+      'body' => array(
+        'property' => 'body',
+        'formatter' => array(
+          'type' => 'text_summary_or_trimmed',
+          'settings' => array(
+            'trim_length' => 100,
+          ),
+        ),
+      ),
     );
   }
 
--- a/modules/restful_example/src/Plugin/resource/Articles__1_0.php
+++ b/modules/restful_example/src/Plugin/resource/Articles__1_0.php
@@ -7,6 +7,9 @@
 
 namespace Drupal\restful_example\Plugin\resource;
 
+use Drupal\restful\Http\Request;
+use Drupal\restful\Plugin\resource\Field\ResourceFieldBase;
+use Drupal\restful\Plugin\resource\Resource;
 use Drupal\restful\Plugin\resource\ResourceEntity;
 use Drupal\restful\Plugin\resource\ResourceInterface;
 
@@ -44,6 +47,9 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
       'label' => array(
         'wrapper_method' => 'label',
         'wrapper_method_on_entity' => TRUE,
+        'process_callbacks' => array(
+          array(array($this, 'addPrefix'), array('Label: ')),
+        ),
       ),
       'self' => array(
         'callback' => array($this, 'getEntitySelf'),
@@ -60,6 +66,21 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
         'property' => 'status',
       ),
     );
+  }
+
+  /**
+   * Helper function that adds a prefix.
+   *
+   * @param mixed $value
+   *   The input value to be prefixed.
+   * @param string $prefix
+   *   The prefix to add.
+   *
+   * @return string
+   *   The prefixed value.
+   */
+  public function addPrefix($value, $prefix) {
+    return $prefix . $value;
   }
 
 }
--- a/modules/restful_example/src/Plugin/resource/Articles__1_0.php
+++ b/modules/restful_example/src/Plugin/resource/Articles__1_0.php
@@ -8,8 +8,6 @@
 namespace Drupal\restful_example\Plugin\resource;
 
 use Drupal\restful\Http\Request;
-use Drupal\restful\Plugin\resource\Field\ResourceFieldBase;
-use Drupal\restful\Plugin\resource\Resource;
 use Drupal\restful\Plugin\resource\ResourceEntity;
 use Drupal\restful\Plugin\resource\ResourceInterface;
 
@@ -64,6 +62,7 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
       ),
       'status' => array(
         'property' => 'status',
+        'methods' => array(Request::METHOD_POST, Request::METHOD_PUT),
       ),
     );
   }
--- a/modules/restful_example/src/Plugin/resource/Articles__1_0.php
+++ b/modules/restful_example/src/Plugin/resource/Articles__1_0.php
@@ -8,6 +8,7 @@
 namespace Drupal\restful_example\Plugin\resource;
 
 use Drupal\restful\Http\Request;
+use Drupal\restful\Http\RequestInterface;
 use Drupal\restful\Plugin\resource\ResourceEntity;
 use Drupal\restful\Plugin\resource\ResourceInterface;
 
@@ -62,7 +63,7 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
       ),
       'status' => array(
         'property' => 'status',
-        'methods' => array(Request::METHOD_POST, Request::METHOD_PUT),
+        'methods' => array(RequestInterface::METHOD_POST, RequestInterface::METHOD_PUT),
       ),
       'body' => array(
         'property' => 'body',
--- a/modules/restful_example/src/Plugin/resource/Articles__2_0.php
+++ b/modules/restful_example/src/Plugin/resource/Articles__2_0.php
@@ -1,0 +1,51 @@
+<?php
+
+/**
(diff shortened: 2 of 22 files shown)


amitaibu:
Dude, you are too fast - you write faster then I'm able to review :wink:

modules/restful_example/src/Plugin/resource/Articles__1_0.php
```diff
@@ -39,17 +41,55 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
   protected function publicFields() {
     return array(
       'id' => array(
-        'wrapper_method' => 'getIdentifier',
-        'wrapper_method_on_entity' => TRUE,
+        'property' => 'nid',
```
amitaibu, line 44: minor,but I prefer the existing format of  `$public_fields['id'] = array(..` -- just because it make it easier to copy/ paster from one resource to another.

modules/restful_example/src/Plugin/resource/Articles__1_0.php
```diff
@@ -39,17 +41,55 @@ class Articles__1_0 extends ResourceEntity implements ResourceInterface {
...
+        'property' => 'field_tags',
+        'resource' => array(
+          'name' => 'tags',
+          'majorVersion' => 1,
+          'minorVersion' => 0,
+        ),
+      ),
+      'status' => array(
+        'property' => 'status',
+        'methods' => array(RequestInterface::METHOD_POST, RequestInterface::METHOD_PUT),
```
amitaibu, line 66: worth adding a comment, what it does.

ntw, AFAIK the status property of node isn't handled properly by Entity API - only admins can write it -- the node author doesn't have access to it.

modules/restful_example/src/Plugin/resource/Articles__2_0.php
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Class Articles
+ * @package Drupal\restful\Plugin\resource
+ *
+ * @Resource(
+ *   name = "articles:2.0",
+ *   resource = "articles",
+ *   label = "Articles",
+ *   description = "Export the article content type with cookie authentication.",
```
amitaibu, line 21: description isn't right if `authenticationTypes = TRUE`

modules/restful_example/src/Plugin/resource/Articles__2_0.php
```diff
@@ -0,0 +1,51 @@
...
+ *   majorVersion = 2,
+ *   minorVersion = 0
+ * )
+ */
+class Articles__2_0 extends ResourceEntity implements ResourceInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function publicFields() {
```
amitaibu, line 47: is it needed?

modules/restful_example/src/Plugin/resource/Tags__1_0.php
```diff
@@ -0,0 +1,79 @@
...
+        'callback' => array($this, 'getEntitySelf'),
+        'access_callbacks' => array(
+          array($this, 'evenAccess'),
+        ),
+      ),
+    );
+  }
+
+  /**
+   * Access callback example.
```
amitaibu, line 62: Access callback example; Allow access only to IDs with even number.

modules/restful_example/src/Plugin/resource/Tags__1_0.php
```diff
@@ -0,0 +1,79 @@
...
+   * @param ResourceFieldInterface $resource_field
+   *   The resource field definition object.
+   * @param DataInterpreterInterface $interpreter
+   *   The data interpreter.
+   *
+   * @return bool
+   *   TRUE for access granted.
+   */
+  public function evenAccess($op, ResourceFieldInterface $resource_field, DataInterpreterInterface $interpreter) {
+    $account = $interpreter->getAccount();
```
amitaibu, line 75: `$interpreter->getAccount();` - why does the $interpreter need this method. Also, don't we have `$this->getAccount()`?

modules/restful_example/src/Plugin/resource/Tags__1_0.php
```diff
@@ -0,0 +1,79 @@
...
+    $value = $interpreter->getWrapper()->getIdentifier() + $account->uid;
```
amitaibu, line 76: maybe simplify and just `return $interpreter->getWrapper()->getIdentifier() % 2;`

restful.module
```diff
@@ -23,6 +22,7 @@ function restful_ctools_plugin_directory($module, $plugin) {
   if ($module == 'restful') {
     return 'plugins/' . $plugin;
   }
+  return NULL;
```
amitaibu, line 25: this whole function will be removed once everything is converted to Plug, right?

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -207,6 +242,11 @@ public function access($op, DataInterpreterInterface $interpreter) {
 
     // Perform field API access checks.
     $property_wrapper = $this->propertyWrapper($interpreter);
+    if ($this->isWrapperMethodOnEntity() && $this->getWrapperMethod() && $this->getProperty()) {
+      // Sometimes we define fields as $wrapper->getIdentifier. We need to
+      // resolve that to $wrapper->nid to call $wrapper->nid->info().
+      $property_wrapper = $property_wrapper->{$this->getProperty()};
```
amitaibu, line 248: this part isn't clear, can you give an example.

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -421,6 +474,26 @@ public function getEntityType() {
...
+    // If there is no property try to get it based on the wrapper method and
+    // store the value in the decorated object.
+    $property = NULL;
+    // If entity metadata wrapper methods were used, then return the appropriate
+    // entity property.
+    if (!$this->isWrapperMethodOnEntity() || !($wrapper_method = $this->getWrapperMethod())) {
+      return NULL;
+    }
+    $entity_info = entity_get_info($entity_type);
+    if ($wrapper_method == 'label') {
```
amitaibu, line 487: can you also please explain this part? I'm not sure what it does, but maybe we also need to take care of the `$wrapper_method == 'url'` case?

src/Plugin/resource/Field/ResourceFieldEntityReference.php
```diff
@@ -149,4 +149,49 @@ protected static function subRequestId(array $value) {
     return $value['id'];
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function value(DataInterpreterInterface $interpreter) {
+    $value = $this->decorated->value($interpreter);
```
amitaibu, line 156: this seems to be duplicated, can we avoid it?

amitaibu:
@mateu-aguilo-bosch I've added a few minor comments/ questions -- it can probably be already merged.

@RoySegall @HelenaEksler  - in your contribution hours please start updating the plugins under restful_example...

modules/restful_example/src/Plugin/resource/Articles__2_0.php
e0ipso, line 21: :+1:

modules/restful_example/src/Plugin/resource/Articles__1_0.php
e0ipso, line 66: Myabe it's a bad example, I just wanted to show that it won't appear for GET requests.

modules/restful_example/src/Plugin/resource/Tags__1_0.php
e0ipso, line 62: :+1:

modules/restful_example/src/Plugin/resource/Tags__1_0.php
e0ipso, line 75: It needs it since the data in the DB will be made available to the user differently based on the account roles (it will perform access checks, etc).

modules/restful_example/src/Plugin/resource/Tags__1_0.php
e0ipso, line 76: :+1:

restful.module
e0ipso, line 25: Yes.

At this point everything is converted, but the cleanup work has not been done yet.

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 248: Sure.

This is trying to get the property that is behind `EntityDrupalWrapper` helper methods. For instance, for a node it will return `nid` if we are using `getIdentifier()` and `fid` for a file entity.

The idea behind this is that if we have a node resource with:

``` php
array(
  'title' => array(
    'method_wrapper' => 'label',
    'method_on_wrapper' => TRUE, // Don't remember the exact key.
  ),
);
```

We can read without problem, but we won't be able to map that back to the `title` column in the `node` table for writing. What this call does is return `title` for the `label` callback.

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 487: We should probably add here: `getBundle`. But I don't know about `\EntityDrupalWrapper::url()`.

modules/restful_example/src/Plugin/resource/Articles__1_0.php
e0ipso, line 44: :+1:

modules/restful_example/src/Plugin/resource/Articles__2_0.php
e0ipso, line 47: Yes it is.

`Resource::publicFields` is abstract, and there is no default implementation in `ResourceEntity`. Maybe we need a follow up ticket to add a default implementation in `ResourceEntity`?

modules/restful_example/src/Plugin/resource/Tags__1_0.php
e0ipso, line 75: Also, we do have `$this->getAccount()`.

PR #456, opened 2015-04-01
Only call prepareRow if $row is an array.

chasingmaxwell:
I'm not sure if this is the right way to handle this, but I thought I'd put something out there to get the ball rolling on a solution. This fixes https://github.com/RESTful-Drupal/restful/issues/331.


amitaibu:
`RestfulFormatterHalJson` isn't properly tested. @chasingmaxwell would you be up for adding some test coverage so we stop breaking it over and over again? :)

e0ipso:
@amitaibu, @chasingmaxwell merging this and creating a follow up to add testing.

PR #457, opened 2015-04-01
Support autocomplete when bundle plugin property is ommitted

chasingmaxwell:
This allows autocomplete on plugins that are intentionally including all bundles by excluding the bundle property. Otherwise getBundlesForAutocomplete() would return array(0) since the entity type has bundles but getBundle() returns 0.


amitaibu:
Would it be possible to add a test to that - e.g. to assert autocomplete for `user` entity works?

e0ipso:
@chasingmaxwell, @amitaibu merging. I will create a follow up to add tests.

PR #460, opened 2015-04-02
Allow plugins to be sorted by weight.

chasingmaxwell:
The use case for this:
We have extended RestfulFormatterHalJson with our own class and utilized it in a custom formatter plugin. However, if the request comes through with an Accept header of 'application/hal+json' the regular hal+json formatter is always used. This is because outputFormat() bails as soon as it finds a matching formatter and the regular hal+json formatter always comes first in the array. This pull request allows us to specify that our custom hal+json formatter should come first.

I also created an additional helper function to prevent from repeating myself in each of the others.

### DIFF
--- a/docs/plugin.md
+++ b/docs/plugin.md
@@ -146,8 +146,20 @@ $plugin = array(
 );
 ```
 
+
 ## Setting the default range
 The range can be specified by setting `$this->range` in your plugin definition.
+
+
+## Setting plugin precedence
+Plugins can be sorted relative to each other by specifying a weight. This is useful, for example, in the case of formatters where the first formatter matching the Accept request header is used.
+
+```php
+$plugin = array(
+  ...
+  'weight' => -1,
+);
+```
 
 
 ### Disabling the range parameter
--- a/modules/restful_example/plugins/formatter/hal_xml.inc
+++ b/modules/restful_example/plugins/formatter/hal_xml.inc
@@ -5,4 +5,5 @@ $plugin = array(
   'description' => t('Output in using the HAL conventions and XML format.'),
   'name' => 'hal_xml',
   'class' => 'RestfulFormatterHalXml',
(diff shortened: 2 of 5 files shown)


amitaibu:
Thanks. This requires updating the README with some info about the expected key in the plugin.

Also, a simple test could go a long way - just use `restful_get_plugins_by_type()` and assert you get the expected order.

amitaibu:
@chasingmaxwell ^^

chasingmaxwell:
Sorry @amitaibu, this week has been crazy. I'm hoping to knock this out today.

amitaibu:
:+1:

chasingmaxwell:
Okay, I've added a simple test. I hope it's okay that I added `weight: -1` to the hal_xml formatter plugin in restful_example. It made sense to test the most likely use case (ordering formatter plugins) and there was already a formatter plugin defined there.

amitaibu:
> in the case of formatters where the first formatter matching the Accept request header is used

After another reading, I believe what you want is not a weight system, but to be able to _disable_ the other plugin, correct?

In that case should we go with `hook_ctools_plugin_pre_alter`  or `hook_ctools_plugin_post_alter` to remove the one(s) you don't want?

(p.s. sorry for thinking about it, only after you've added the test)

chasingmaxwell:
That would also meet our use case. Are you imagining a way for a plugin to specify other plugins which should be disabled or just that I can implement those alter hooks to meet my specific use case? Also, I wonder if there might be other use cases for sorting plugins?

amitaibu:
> That would also meet our use case

Cool :)

> Also, I wonder if there might be other use cases for sorting plugins?

I don't think so, since we're always answering with a single resource/ formatter. @mateu-aguilo-bosch ?

amitaibu:
Closing this, as I believe there's already an existing solution.

PR #464, opened 2015-04-06
Add support for LIKE and STARTS_WITH operators

edhollinghurst:
This PR adds support for use of the LIKE and STARTS_WITH operators when adding a field condition to a query.

### DIFF
--- a/plugins/restful/RestfulDataProviderEFQ.php
+++ b/plugins/restful/RestfulDataProviderEFQ.php
@@ -175,6 +175,10 @@ abstract class RestfulDataProviderEFQ extends \RestfulBase implements \RestfulDa
           $query->fieldCondition($public_fields[$filter['public_field']]['property'], $public_fields[$filter['public_field']]['column'], $filter['value'], $filter['operator'][0]);
           continue;
         }
+        if (in_array(strtoupper($filter['operator'][0]), array('LIKE', 'STARTS_WITH'))) {
+          $query->fieldCondition($public_fields[$filter['public_field']]['property'], $public_fields[$filter['public_field']]['column'], $filter['value'][0], $filter['operator'][0]);
+          continue;
+        }
         for ($index = 0; $index < count($filter['value']); $index++) {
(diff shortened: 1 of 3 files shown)


amitaibu:
Thanks. This will also require a (simple) simpleTest :)

e0ipso:
We also need support for those operators on entity properties, not just fields.

edhollinghurst:
When I looked at this again I realised that the LIKE and STARTS_WITH operators are already handled correctly in RestfulDataProviderEFQ.php for both fields and properties. It's just a case of enabling these operators in RestfulBase.php.

I've added a couple of tests for these operators to RestfulListTestCase.test.

e0ipso:
@edhollinghurst actually by enabling it in `RestfulBase` you are enabling it as well for all of the data providers. Probably this will fail for other data providers, like for instance `RestfulDataProviderDbQuery`.

edhollinghurst:
I've now taken out the change from `RestfulBase` and overridden the function in `RestfulDataProviderEFQ`.

I also made a small correction to the error message that's thrown if the operator is not allowed. `$operators` was being passed as the token when this should be `$operator`.

edhollinghurst:
@mateu-aguilo-bosch Is there anything more you need me to do in terms of this?

plugins/restful/RestfulDataProviderEFQ.php
```diff
@@ -209,6 +209,34 @@ protected function getColumnFromProperty($property_name) {
...
+  protected static function isValidOperatorsForFilter(array $operators) {
+    $allowed_operators = array(
+      '=',
+      '>',
+      '<',
+      '>=',
+      '<=',
+      '<>',
+      '!=',
+      'IN',
```
amitaibu, line 223: There's also

```
+      'CONTAINS',
+      'IN',
+      'NOT IN',
+      'STARTS_WITH',
```

btw, the file chmod should be changed back to 644

amitaibu:
Should we also add this to `RestfulBase.php`?

This needs a re-roll.

e0ipso:
> Should we also add this to RestfulBase.php?

I don't think that we should. Those apply to all of the data providers, but only some of them support that list of operators.

e0ipso:

Thanks for everything!

PR #467, opened 2015-04-13
466 drupal static for parse restful

### LINKED ISSUE #466: parse_restful_request  - php://Input can only be read once in php 5.6 and below
After it's read once it disappears because previous to php 5.6 php://input was only intended to be read once. This is a useful function and the only part of it that changes if it is called more than once is the query string -but only on PUTS and PATCHES and other non POST || GET actions.

Because this is out of the $this space and code in the .module it'd be cool if we could use a drupal static on a helper function to allow it to return the same results even if it is called again - for whatever reason.

thebruce:
This Resolves #466 by adding a drupal static in place of using file_get_contents('php://input') which in php 5.6 and lower is only read 1 time.

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -556,7 +556,7 @@ function restful_parse_request() {
     $request = $_POST;
   }
 
-  if (!$request && $query_string = file_get_contents('php://input')) {
+  if (!$request && $query_string = _restful_get_query_string_from_php_input()) {
     // When trying to POST using curl on simpleTest it doesn't reach
     // $_POST, so we try to re-grab it here.
     // Also, sometimes the client might send the input still encoded.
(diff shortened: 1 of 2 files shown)


restful.module
```diff
@@ -786,3 +786,18 @@ function restful_is_user_switched() {
...
+/**
+ * Get the query string from php Input.
+ *
+ * @return string
+ *   The query string from request.
+ */
+function _restful_get_query_string_from_php_input() {
+    $query_string = &drupal_static(__FUNCTION__);
+    if (!isset($query_string)) {
+      // In php 5.6 and below php://input can only be read once.
```
amitaibu, line 798: That's surprising - have you seen it documented anywhere?

e0ipso, line 798: I just read about this :open_mouth: 

- manual 2015-04-14 14-30-38

e0ipso:
This was merged. Thanks yet again @thebruce!

PR #477, opened 2015-04-16
2.x Enable Render cache feature

e0ipso:
This PR will make sure that we can decorate resources to stack functionality to a resource. For now there are 2 decorators:
- `CachedResource`: adds caching capabilities.
- `RateLimitedResource`: adds rate limit capabilities.

That is specially good because it allows other modules to add new decorators and cherry pick the decorators to use (via the annotation configuration). Yes! (this was one of the big points of the re-architecture).

Also, this PR fixes render cache. With the increased complexity, that comes from the greater flexibility, the render cache is even more important. See a basic benchmark below.

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -12,7 +12,10 @@ use Drupal\restful\Exception\RestfulException;
 use Drupal\restful\Exception\ServiceUnavailableException;
 use Drupal\restful\Http\HttpHeader;
 use Drupal\restful\Http\RequestInterface;
+use Drupal\restful\Plugin\resource\CachedResource;
+use Drupal\restful\Plugin\resource\RateLimitedResource;
 use Drupal\restful\Plugin\resource\ResourceInterface;
+use Drupal\restful\RateLimit\RateLimitManager;
 use Drupal\restful\RestfulManager;
 
 /**
@@ -559,5 +562,20 @@ function restful_date_time_format_element_validate($element, &$form_state) {
       '%name' => $element['#title'],
       '!link' => l(t('\DateInterval format'), 'http://php.net/manual/en/class.dateinterval.php'),
     )));
+  }
+}
+
+/**
+ * Implements hook_restful_resource_alter().
+ */
+function restful_restful_resource_alter(ResourceInterface &$resource) {
+  $plugin_definition = $resource->getPluginDefinition();
+  // Check for the rate limit configuration.
+  if (!empty($plugin_definition['rateLimit'])) {
+    $resource = new RateLimitedResource($resource);
+  }
+  // Check for the rate limit configuration.
+  if (!empty($plugin_definition['renderCache'])) {
+    $resource = new CachedResource($resource);
   }
 }
--- a/restful.module
+++ b/restful.module
@@ -566,6 +566,9 @@ function restful_date_time_format_element_validate($element, &$form_state) {
 
 /**
  * Implements hook_restful_resource_alter().
+ *
+ * Decorate an existing resource with other services (e.g. rate limit and render
+ * cache).
  */
 function restful_restful_resource_alter(ResourceInterface &$resource) {
   $plugin_definition = $resource->getPluginDefinition();
@@ -574,7 +577,7 @@ function restful_restful_resource_alter(ResourceInterface &$resource) {
     $resource = new RateLimitedResource($resource);
   }
   // Check for the rate limit configuration.
-  if (!empty($plugin_definition['renderCache']) && !empty($plugin_definition['renderCache']['render'])) {
+  if (!empty($plugin_definition['renderCache']['render'])) {
     $resource = new CachedResource($resource);
   }
 }
--- a/restful.module
+++ b/restful.module
@@ -262,7 +262,7 @@ function restful_menu_access_callback($resource_name, $version_string = NULL) {
 
   try {
     return $resource_manager
-      ->getPlugin($resource_name . PluginBase::DERIVATIVE_SEPARATOR . implode('.', $versions))
+      ->getPlugin($resource_name . PluginBase::DERIVATIVE_SEPARATOR . implode('.', $versions), restful()->getRequest())
       ->access();
   }
   catch (RestfulException $e) {
@@ -575,7 +575,7 @@ function restful_restful_resource_alter(ResourceInterface &$resource) {
     $resource = new RateLimitedResource($resource);
   }
   // Check for the rate limit configuration.
-  if (!empty($plugin_definition['renderCache'])) {
+  if (!empty($plugin_definition['renderCache']) && !empty($plugin_definition['renderCache']['render'])) {
     $resource = new CachedResource($resource);
   }
 }
--- a/restful.module
+++ b/restful.module
@@ -12,10 +12,9 @@ use Drupal\restful\Exception\RestfulException;
 use Drupal\restful\Exception\ServiceUnavailableException;
 use Drupal\restful\Http\HttpHeader;
 use Drupal\restful\Http\RequestInterface;
-use Drupal\restful\Plugin\resource\CachedResource;
-use Drupal\restful\Plugin\resource\RateLimitedResource;
+use Drupal\restful\Plugin\resource\Decorators\CachedResource;
+use Drupal\restful\Plugin\resource\Decorators\RateLimitedResource;
 use Drupal\restful\Plugin\resource\ResourceInterface;
-use Drupal\restful\RateLimit\RateLimitManager;
 use Drupal\restful\RestfulManager;
 
 /**
--- a/restful.module
+++ b/restful.module
@@ -125,9 +125,9 @@ function restful_menu() {
     }
     $item = array(
       'title' => $plugin_definition['name'],
-      'access callback' => 'restful_menu_access_callback',
+      'access callback' => RestfulManager::FRONT_CONTROLLER_ACCESS_CALLBACK,
       'access arguments' => array($plugin_definition['resource']),
-      'page callback' => 'restful_menu_process_callback',
+      'page callback' => RestfulManager::FRONT_CONTROLLER_CALLBACK,
       'page arguments' => array($plugin_definition['resource']),
       'delivery callback' => 'restful_formatted_delivery',
       'type' => MENU_CALLBACK,
--- a/src/Authentication/AuthenticationManager.php
+++ b/src/Authentication/AuthenticationManager.php
@@ -10,6 +10,7 @@ namespace Drupal\restful\Authentication;
 use Drupal\restful\Exception\UnauthorizedException;
 use Drupal\restful\Http\RequestInterface;
 use Drupal\restful\Plugin\AuthenticationPluginManager;
+use Drupal\restful\RestfulManager;
 
 class AuthenticationManager implements AuthenticationManagerInterface {
 
@@ -103,7 +104,7 @@ class AuthenticationManager implements AuthenticationManagerInterface {
 
     if (!$account) {
 
-      if ($this->plugins->count() && !$this->getIsOptional()) {
+      if (RestfulManager::isRestfulPath($request) && $this->plugins->count() && !$this->getIsOptional()) {
         // User didn't authenticate against any provider, so we throw an error.
         throw new UnauthorizedException('Bad credentials');
       }
--- a/src/Plugin/resource/CachedResource.php
+++ b/src/Plugin/resource/CachedResource.php
@@ -337,4 +337,11 @@ class CachedResource extends PluginBase implements ResourceInterface {
     return $cache_info;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getResourceMachineName() {
+    return $this->subject->getResourceMachineName();
+  }
+
 }
--- a/src/Plugin/resource/CachedResource.php
+++ b/src/Plugin/resource/CachedResource.php
@@ -8,9 +8,10 @@
 namespace Drupal\restful\Plugin\resource;
 
 use Drupal\Component\Plugin\PluginBase;
-use Drupal\restful\Exception\NotImplementedException;
+use Drupal\restful\Http\HttpHeader;
 use Drupal\restful\Http\RequestInterface;
 use Drupal\restful\Plugin\resource\DataProvider\DataProviderInterface;
+use Drupal\restful\Resource\ResourceManager;
 
 class CachedResource extends PluginBase implements ResourceInterface {
 
@@ -47,6 +48,8 @@ class CachedResource extends PluginBase implements ResourceInterface {
     // TODO: Implement the ResourceManager factory to use the CachedResource.
     $this->subject = $subject;
     $this->cacheController = $cache_controller ? $cache_controller : $this->newCacheObject();
+    $cache_info = $this->defaultCacheInfo();
+    $this->pluginDefinition['renderCache'] = $cache_info;
   }
 
   /**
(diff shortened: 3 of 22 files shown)


e0ipso:
From the commit messages:

> Create hook to decorate plugins with extra functionality.
> 
> There is a hook that you can implement and based on the plugin
> definition, you can create a decorated resource that wraps around
> the resource. This is the standard way to stack functionality to
> your resources, do not use inheritance!

e0ipso:
Initial benchmarks show a ~40% performance increase in a very basic test. It is reasonable to think that the more complex the returned resource is, the bigger the performance improvement.

See:

```
$> ab -n 50 -c 3 -H 'X-Api-Version: v1.1' http://restful.local/api/articles/88
This is ApacheBench, Version 2.3 <$Revision: 1554214 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking restful.local (be patient).....done


Server Software:        Apache/2.4.7
Server Hostname:        restful.local
Server Port:            80

Document Path:          /api/articles/88
Document Length:        348 bytes

Concurrency Level:      3
Time taken for tests:   1.388 seconds
Complete requests:      50
Failed requests:        0
Total transferred:      48700 bytes
HTML transferred:       17400 bytes
Requests per second:    36.02 [#/sec] (mean)
Time per request:       83.280 [ms] (mean)
Time per request:       27.760 [ms] (mean, across all concurrent requests)
Transfer rate:          34.26 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    51   82  19.6     80     138
Waiting:       51   82  19.6     80     138
Total:         51   82  19.6     80     138

Percentage of the requests served within a certain time (ms)
  50%     80
  66%     88
  75%     98
  80%    103
  90%    109
  95%    113
  98%    138
  99%    138
 100%    138 (longest request)
```

```
$> ab -n 50 -c 3 -H 'X-Api-Version: v1.0' http://restful.local/api/articles/88
This is ApacheBench, Version 2.3 <$Revision: 1554214 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking restful.local (be patient).....done


Server Software:        Apache/2.4.7
Server Hostname:        restful.local
Server Port:            80

Document Path:          /api/articles/88
Document Length:        348 bytes

Concurrency Level:      3
Time taken for tests:   2.327 seconds
Complete requests:      50
Failed requests:        0
Total transferred:      48700 bytes
HTML transferred:       17400 bytes
Requests per second:    21.49 [#/sec] (mean)
Time per request:       139.604 [ms] (mean)
Time per request:       46.535 [ms] (mean, across all concurrent requests)
Transfer rate:          20.44 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    93  138  19.2    137     181
Waiting:       93  138  19.2    137     181
Total:         94  138  19.2    137     181

Percentage of the requests served within a certain time (ms)
  50%    137
  66%    148
  75%    152
  80%    154
  90%    163
  95%    168
  98%    181
  99%    181
 100%    181 (longest request)
```

restful.module
```diff
@@ -561,3 +564,18 @@ function restful_date_time_format_element_validate($element, &$form_state) {
     )));
   }
 }
+
+/**
+ * Implements hook_restful_resource_alter().
+ */
```
amitaibu, line 572: `Decorate an existing resource with other services (e.g. rate limit and render cache).`

restful.module
```diff
@@ -561,3 +564,18 @@ function restful_date_time_format_element_validate($element, &$form_state) {
...
+function restful_restful_resource_alter(ResourceInterface &$resource) {
+  $plugin_definition = $resource->getPluginDefinition();
+  // Check for the rate limit configuration.
+  if (!empty($plugin_definition['rateLimit'])) {
+    $resource = new RateLimitedResource($resource);
+  }
+  // Check for the rate limit configuration.
+  if (!empty($plugin_definition['renderCache']) && !empty($plugin_definition['renderCache']['render'])) {
```
amitaibu: should be enough just `!empty($plugin_definition['renderCache']['render']`

restful.module
```diff
@@ -561,3 +564,18 @@ function restful_date_time_format_element_validate($element, &$form_state) {
     )));
   }
 }
+
+/**
+ * Implements hook_restful_resource_alter().
+ */
+function restful_restful_resource_alter(ResourceInterface &$resource) {
```
amitaibu, line 573: If we already decided to go with a hook (instead of the decorators doing their own registration themself), I think that maybe we should provide a hook - where those plugins return the class, and keep the alter as a way for implementing modules to remove/ swap some classes?

src/Authentication/AuthenticationManager.php
```diff
@@ -103,7 +104,7 @@ public function getAccount(RequestInterface $request, $cache = TRUE) {
 
     if (!$account) {
 
-      if ($this->plugins->count() && !$this->getIsOptional()) {
+      if (RestfulManager::isRestfulPath($request) && $this->plugins->count() && !$this->getIsOptional()) {
```
amitaibu, line 107: `RestfulManager::isRestfulPath($request)` -- how come this was invoked if this wasn't a restful path?

do we care if it's not a restful path? I for example already had a cases where I've manually invoked the authentication manager for non restful path.

src/Plugin/resource/CachedResource.php
```diff
@@ -191,14 +206,24 @@ public function getControllers() {
    * {@inheritdoc}
    */
   public function index($path) {
-    return $this->subject->index($path);
+    return $this->getDataProvider()->index();
```
amitaibu: If I understand correctly `$this->subject` was holding up until now the "main" resource. I think that it would make it more readable to have a `$this->getPrimaryResource()->index()` - so it's clear the CacheResource doesn't care about how the data was retrieved -- it just needs to know who's responsible for it.

Naming is hard, but few suggestions:
$this->getMatchedResource()
$this->getMainResource()
$this->getPrimaryResource() -- (my favorite)

src/Plugin/resource/DataProvider/CachedDataProvider.php
```diff
@@ -261,10 +257,33 @@ public static function invalidateEntityCache($cid) {
...
+    $options = $this->getOptions();
+    $cache_info = $options['renderCache'];
+
+    if (!$cache_info['simple_invalidate']) {
+      // Simple invalidation is disabled. This means it is up to the
+      // implementing module to take care of the invalidation.
+      return;
+    }
+    // If the $cid is not '*' then remove the asterisk since it can mess with
+    // dynamically built wildcards.
```
amitaibu, line 281: is there a link to add to give more info?

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -128,6 +129,9 @@ public function __construct(RequestInterface $request, ResourceFieldCollectionIn
    * {@inheritdoc}
    */
   public function getContext($identifier) {
+    if (is_array($identifier)) {
```
amitaibu, line 132: when do we get an array?

src/Plugin/resource/CachedResource.php
amitaibu: Once we also have this `$this->getPrimaryResource()` I think we can remove all the other proxy methods, so it's always clear it's the primary resource responsible for the data. (example of such methods):

``` php
  /**
   * {@inheritdoc}
   */
  public function getContext($identifier) {
    return $this->subject->getContext($identifier);
  }
```

src/Plugin/resource/RateLimitedResource.php
```diff
@@ -196,4 +202,67 @@ public function getVersion() {
     return $this->subject->getVersion();
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function versionedUrl($path = '', $options = array(), $version_string = TRUE) {
```
amitaibu, line 180: I'm not a big fan of this proxy methods -- see previous comments. While it saves a few key strokes, it makes it unclear who's responsible for what.

src/Plugin/resource/ResourceInterface.php
```diff
@@ -253,4 +271,27 @@ public function versionedUrl($path = '', $options = array(), $version_string = T
    */
   public function access();
 
+  /**
+   * Return the controller from a given path.
```
amitaibu: controller from => controller of

src/Resource/ResourcePluginCollection.php
```diff
@@ -18,4 +20,43 @@ class ResourcePluginCollection extends DefaultLazyPluginCollection {
...
+  /**
+   * Overwrites LazyPluginCollection::get().
+   */
+  public function &get($instance_id) {
+    /** @var \Drupal\restful\Plugin\resource\ResourceInterface $resource */
+    $resource = parent::get($instance_id);
+    $resource->setConfiguration(array(
+      'request' => $this->request,
+    ));
+    // Allow altering the resource, this way we can read the resource's
```
amitaibu, line 56: maybe move this to the PHPdocs.

src/RestfulManager.php
```diff
@@ -168,4 +168,16 @@ public function process() {
...
+   * Checks if the passed in request belongs to RESTful.
+   *
+   * @param RequestInterface $request
+   *   The path to check.
+   *
+   * @return bool
+   *   TRUE if the path belongs to RESTful.
+   */
+  public static function isRestfulPath(RequestInterface $request) {
+    return strpos($request->getPath(), variable_get('restful_hook_menu_base_path', 'api')) === 0;
```
amitaibu: I'm not sure this check is correct, as resources may define their own hook_menu.

amitaibu:
@mateu-aguilo-bosch beautiful work. I've added a few comments.

restful.module
e0ipso, line 572: :+1:

restful.module
e0ipso: :+1:

restful.module
e0ipso, line 573: > I think that maybe we should provide a hook - where those plugins return the class, and keep the alter as a way for implementing modules to remove/ swap some classes?

I'm not sure I follow. This alter hook is for both implementing modules that provide decorators and restful (as a module that provides decorators). Keep in mind that the composition order is important. `(render cache (rate limit (articles)))` is different from `(rate limit (render cache (articles)))`

src/Authentication/AuthenticationManager.php
e0ipso, line 107: > how come this was invoked if this wasn't a restful path?

When we are building the menu router we get all the plugins. A part of the plugin initialization is setting the account provided by the `AuthenticationManager`.

> do we care if it's not a restful path?

Yes, otherwise the request object (among others) is not prepared and it breaks.

> I for example already had a cases where I've manually invoked the authentication manager for non restful path.

What was the use case?

src/Plugin/resource/CachedResource.php
e0ipso: > I think that it would make it more readable to have a $this->getPrimaryResource()->index()

:+1: I'm cool with that. Anyways `subject` is the nomenclature for the decorator pattern. I will change this to `$this->primaryResource->index()` since I'd rather keep the access to the decorated resource protected.

> I think we can remove all the other proxy methods

We need to keep the proxy methods, since this class implements `ResourceInterface` and should be used as any other resource. What we could do is create a `DecoratedResourceBase` abstract class that has all of the proxy methods. All resource decorators would extend from that one.

src/Plugin/resource/DataProvider/CachedDataProvider.php
e0ipso, line 281: This was copied an pasted from old code. This refers to standard `cache_clear` https://api.drupal.org/api/drupal/includes%21cache.inc/function/DrupalCacheInterface%3A%3Aclear/7

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso, line 132: If we are requesting things like `https://example.org/api/articles/1,2,3`.

src/Plugin/resource/RateLimitedResource.php
e0ipso, line 180: Then we probably do need the `DecoratedResourceBase` abstract class.

src/Plugin/resource/ResourceInterface.php
e0ipso: :+1:

src/Plugin/resource/RateLimitedResource.php
e0ipso, line 180: :+1:

src/Resource/ResourcePluginCollection.php
e0ipso, line 56: :+1:

src/RestfulManager.php
e0ipso: You are absolutely right. We need to check the menu router to see if the page callback is our front controller. That means that we need to centralize our front controller in `RestfulManager`.

e0ipso:
@amitaibu any impressions on the original benchmarks? Maybe worth revisiting the idea of caching resources by default…

restful.module
amitaibu, line 573: > Keep in mind that the composition order is important.

So how do we impose this order easily, without other implementing module needing to deal with `hook_module_implenmets_alter()`? (if no easier way we can keep it, just wondering if you have an idea)

src/Authentication/AuthenticationManager.php
amitaibu, line 107: > What was the use case?

I wanted the user to be switched to the one authenticated by RESTful upon `hook_file_download_access()` -- so on private files an anonymous user can send `https://example.com/system/files/some-private-file.jpg?access_token=12345`

``` php
/**
 * Implements hook_file_download_access().
 *
 * Get the user account from the RESTful authentication.
 */
function shoov_restful_file_download_access($file_item, $entity_type, $entity) {
  $auth_plugins = array_keys(restful_get_authentication_plugins());

  $auth_manager = new RestfulAuthenticationManager();

  // Register all authentication plugins.
  foreach ($auth_plugins as $auth_plugin_name) {
    $auth_manager->addAuthenticationProvider($auth_handler);
    $auth_manager->setIsOptional(TRUE);
  }

  $request = restful_parse_request();
  $auth_manager->getAccount($request);
}
```

src/Plugin/resource/CachedResource.php
amitaibu: > All resource decorators would extend from that one.

:+1:

src/Plugin/resource/DataProvider/DataProviderEntity.php
amitaibu, line 132: Thanks, Worth adding this as a comment.

amitaibu:
> Maybe worth revisiting the idea of caching resources by default…

Remember that the reason we disabled them, was that upon development it was quite confusing - as the one a developer changed for example the `publidFields()` it wasn't reflected. So it would be bad DX, but good for performance ;)

restful.module
e0ipso, line 573: I believe that is a very tricky thing to do. You can play with module weights, but that may have other unintended consequences. I don't know how else this could be achieved.

amitaibu, line 573: Ok, let's keep as is.

e0ipso:
I believe that this is now ready.

e0ipso:
@amitaibu I'm going to merge this and start solving merge conflicts. Please, feel free to create any follow ups we need.

amitaibu:
Looks good. Damn you like abstractions ;)

e0ipso:
hahaha. I do, but this time it was you who asked for `ResourceDecoratorBase`.

PR #481, opened 2015-04-17
2.x Rate limits

e0ipso:
This PR builds on top of https://github.com/RESTful-Drupal/restful/pull/477. This is the [delta](https://github.com/mateu-aguilo-bosch/restful/compare/2.x-render-cache...2.x-rate-limits?expand=1).

This will add Rate Limit support for _any_ (not just entities) resource that has the `rateLimit` property in the annotation definition.

Also, there are some changes that will allow nesting of decorators. Resource decorators are just wrappers that add functionality to a _real_ resource (like for instance rate limits or caching). Resource decorators can be treated like any other decorator, since they implement `ResourceInterface`.


amitaibu:
@mateu-aguilo-bosch can you please re-roll for easier review?

e0ipso:
I'll re-roll tomorrow morning. You can use the _delta_ link to see the changes in this PR isolated.

e0ipso:
done.

restful.module
```diff
@@ -573,11 +573,11 @@ function restful_date_time_format_element_validate($element, &$form_state) {
...
   // Check for the rate limit configuration.
-  if (!empty($plugin_definition['rateLimit'])) {
-    $resource = new RateLimitedResource($resource);
-  }
-  // Check for the rate limit configuration.
   if (!empty($plugin_definition['renderCache']['render'])) {
     $resource = new CachedResource($resource);
   }
+  // Check for the rate limit configuration.
+  if (!empty($plugin_definition['rateLimit'])) {
```
amitaibu, line 580: So rate limit needs to be the last one. IT means that implementing modules will need to alter the alter. We don't have to tackle this now, but worth thinking about a nicer DX

src/Plugin/ResourcePluginManager.php
```diff
@@ -32,12 +44,24 @@ public function __construct(\Traversable $namespaces, \DrupalCacheInterface $cac
...
-  public static function create($bin = 'cache') {
-    return new static(Module::getNamespaces(), _cache_get_object($bin));
+  public static function create($bin = 'cache', RequestInterface $request = NULL) {
+    return new static(Module::getNamespaces(), _cache_get_object($bin), $request);
+  }
+
+  /**
+   * Overrides PluginManagerBase::createInstance().
+   */
+  public function createInstance($plugin_id, array $configuration = array()) {
```
amitaibu, line 63: why do we need to override the base? Worth adding a comment.

src/Plugin/rate_limit/RateLimit.php
```diff
@@ -38,7 +39,12 @@
    */
   public function __construct(array $configuration, $plugin_id, $plugin_definition) {
     parent::__construct($configuration, $plugin_id, $plugin_definition);
-    $this->period = $configuration['period'];
+    try {
+      $this->period = new \DateInterval($configuration['period']);
+    }
+    catch (\Exception $e) {
+      throw new ServerConfigurationException(sprintf('Invalid rate limit period: %s.', $configuration['period']));
```
amitaibu: Maybe: `Invalid rate limit period: %s. Should be a valid format of \DateInterval.`

src/Plugin/resource/DataProvider/DataProviderResource.php
```diff
@@ -63,7 +63,7 @@ public function __construct(RequestInterface $request, ResourceFieldCollectionIn
    * {@inheritdoc}
    */
   public static function init(RequestInterface $request, $resource_name, array $version) {
-    $plugin_manager = ResourcePluginManager::create();
+    $plugin_manager = ResourcePluginManager::create('cache', $request);
```
amitaibu, line 66: what does the `cache` mean here?

src/Plugin/resource/DataProvider/RateLimitedDataProvider.php
```diff
@@ -0,0 +1,183 @@
...
+ * Contains Drupal\restful\Plugin\resource\DataProvider\RateLimitedDataProvider.
+ */
+
+namespace Drupal\restful\Plugin\resource\DataProvider;
+
+use Drupal\restful\Exception\ServerConfigurationException;
+use Drupal\restful\Plugin\resource\Field\ResourceFieldInterface;
+use Drupal\restful\RateLimit\RateLimitManagerInterface;
+
+class RateLimitedDataProvider implements DataProviderInterface {
```
amitaibu: The DataProvider name is confusing here, as infact the RateLimit isn't really a data provider - it just a middleware. (Is the cache resource also a middleware?) Maybe we  actually need :

single resource 
single data provider
multiple middlewares ?

src/RateLimit/Entity/RateLimitController.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+/**
+ * @file
+ * Contains Drupal\restful\RateLimit\Entity\RateLimitController.
+ */
+
+namespace Drupal\restful\RateLimit\Entity;
+
+class RateLimitController extends \EntityAPIController {}
```
amitaibu, line 10: why extend the controller?

restful.module
e0ipso, line 580: Actually Cache needs to be the first one (the one more deep in the composition), since cache will shortcut and return when there is a cache hit.

src/Plugin/resource/DataProvider/RateLimitedDataProvider.php
amitaibu: @mateu-aguilo-bosch does this make sense to you? Want to discuss it on skype?

e0ipso: I'm renaming it to be:

``` php
class RateLimitedMiddleware implements DataProviderInterface {
```

src/RateLimit/Entity/RateLimitController.php
e0ipso, line 10: Having your own typed objects is usually a good idea. This means that you have a way to check if that controller is a rate limit controller. You also give yourself some room to add new methods to it without having to change the API and breaking compat with the major version.

This carries from 1.x.

src/Plugin/resource/DataProvider/DataProviderResource.php
e0ipso, line 66: This is part of the API that exposes `Plug`. This is the bin where the discovered plugins are cached. `'cache'` is the default value.

src/Plugin/rate_limit/RateLimit.php
e0ipso: :+1:

src/Plugin/resource/DataProvider/RateLimitedDataProvider.php
e0ipso: :+1:

src/Plugin/ResourcePluginManager.php
e0ipso, line 63: We are overriding to set the `$request` when the plugin object is instantiated.

e0ipso, line 63: :+1:

src/Plugin/resource/DataProvider/RateLimitedDataProvider.php
e0ipso: You were right, this file was doing nothing :-o

I removed it.

e0ipso:
@amitaibu this should be pretty close to merge now.

amitaibu:
> I'm renaming it to be:
> 
> class RateLimitedMiddleware implements DataProviderInterface {

Couldn't find that class. Anyway, I wonder if indeed it shouldn't implement MiddlewareInterface, so we can make sure there's only a single DataProvider in each request?

src/Plugin/resource/DataProvider/RateLimitedDataProvider.php
amitaibu: oh, ok :)

amitaibu: but still see my comment about the MiddlewareInterface

e0ipso:
> Couldn't find that class.

Yeah, after commenting that I added an extra comment saying that I removed the class entirely.

src/Plugin/resource/DataProvider/RateLimitedDataProvider.php
e0ipso: A data provider is a class that provides full crud on a given backend type (EMW, db_query, filesystem, plugin list, …). Therefore we need a _DataProvider_ for every backend type.

Additionally we have decorators. Those decorators are the candidates to be called _middle ware_. The decorators are _DataProviders_ that have an additional feature added to it (caching, …). The decorators can apply (add these features) to any _DataProvider_, providing the flexibility of horizontal inheritance.

There are different kinds of decorators:
- Resource decorators: add functionality at the resource level. For instance: «increment the rate limit for ever processed request».
- Data provider decorators: add functionality at the data provider level. For instance: «do not make a query for something we did before».
- Resource field decorator: adds functionality to the object representing a single field. For instance: «add Field API capabilities to a generic public property».

As you can see, by using a common name in Computer Science as _decorator_ we can apply it freely throughout the code base. If we go by middle ware, we would need to specify the type of middleware and document how to create one. By calling it a decorator, developers can assume the it is a [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) and know how to implement it.

Was this what you had in mind?

PR #487, opened 2015-04-22
only access_token is allowed when authenticating a user

xuxucode:
at the moment, users can be authenticated using both access_token and refresh_token, the refresh_token should be disallowed


e0ipso:
Closing in favor of #489.

Issue #483, opened 2015-04-17
Files uploaded through PUT do not make it into handler intact


mpetrovic:
I'm trying to replace a file on the server with the file given by the user, and using the PUT method to do so. 

I'm setting the Content-Type header to the mimetype of the file, and the file itself as the data. By the time it gets to my Restful plugin, the data is gone. There's a few bytes of headers in the request array, and then nothing. 

Looking at restful_parse_request, it looks like its treating everything that isn't a GET or POST like it's assuming all content will either be in JSON or query string format. 

**Proposal**
Check the Content-Type header for the request when the method is not GET or POST. If it's application/x-www-form-urlencoded, then run it through parse_str. If it's application/json, run it through drupal_json_decode. If it's mulltipart/form-data, parse it manually. if it's anything else, pass the request data untouched to the handler for it to handle.

rbran100:
@mateu-aguilo-bosch does @mpetrovic solution sound good to you?

amitaibu:
Hi OpenScholar folks! :)
I'm not sure I follow. What I would expect to send (and I'm not sure it's going to work, as I've never tested it when I wrote the File resource), is something like this:

amitaibu:
![postman](https://cloud.githubusercontent.com/assets/125707/7235657/11d4cebc-e798-11e4-9217-4b383c0494f8.jpg)

mpetrovic:
I've been trying with lots of things. The closest I've gotten is with raw. It cannot be form-data, since php://input just doesn't work. I haven't tried x-www-form-urlencoded in my own stuff, so I don't know how that handles files.

amitaibu:
@mpetrovic wouldn't another solution be:

1) Create a new file (as would happen if you used a form)
2) Replace the references in the node(s) referencing the new file
3) Delete the old file

mpetrovic:
That is way more work then swapping the file out on disk. More error-prone too. All I'm trying to do is replace the file on disk with a new one.

amitaibu:
Are you using the default file upload (`/api/file-upload`). If so then it currently support only [create](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/RestfulFilesUpload.php#L16)

> That is way more work then swapping the file out on disk. More error-prone too

You are probably right, but I believe it's the same it's handled when uploading a file via form.

mpetrovic:
[...]

PR #488, opened 2015-04-22
Allow files to be uploaded directly through PUT method. 

mpetrovic:
From #483.

If the Content-Type header is not a known value, we assume it's a direct file upload. The file is copied to a temporary location, made into a Drupal File entity, and then put into the request array to be passed to the handler. At this point, it's up to the handler to move the file into a permanent location or process it.


e0ipso:
@mpetrovic thank you for contributing! Any idea why tests are failing?

mpetrovic:
@mateu-aguilo-bosch 
The tests are failing because they say they set the Content-Type header to 'application/json' data and then actually sending an HTTP encoded string. They either need to set the header to 'application/x-www-form-encoded' or send actual JSON data.

mpetrovic:
@mateu-aguilo-bosch 
There are still 4 tests failing. They are failing because the $_POST array is empty, even though php://input has data and the Content-Type is 'application/x-form-www-encoded'. I can't think of why this would be though.

mpetrovic:
Closing in favor of a cleaner PR.

rbran100:
New PR #568

PR #489, opened 2015-04-22
generate access token and refresh token for correct users

xuxucode:
the resulting uid of both access token and refresh token is 0, when trying to refresh token (e.g. make a call to http://www.example.com/api/refresh-token/89LcfdhOWptEEcjcoDpuLGZVKUX_Agpp_IL2VwohN1Q).

### DIFF
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -30,6 +30,7 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
     $query = new EntityFieldQuery();
     $result = $query
       ->entityCondition('entity_type', 'restful_token_auth')
+      ->entityCondition('bundle', 'access_token')
       ->propertyCondition('token', $token)
       ->range(0, 1)
       ->execute();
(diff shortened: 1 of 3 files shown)


e0ipso:
This looks great!

Would you mind adding a test that passes for authentication denied using the refresh token?

xuxucode:
I want to add some test, but:

`Fatal error: Call to a member function get() on a non-object in restful/modules/restful_token_auth/tests/RestfulTokenAuthenticationTestCase.test on line 42
FATAL RestfulTokenAuthenticationTestCase: test runner returned a non-zero error code (255).`

I don't know what to do

xuxucode:
restful_get_restful_handler('access _token') returns null, since restful_get_restful_plugins() always returns empty array() in the test

amitaibu:
> restful_get_restful_handler('access _token') returns null, since restful_get_restful_plugins() always returns empty array() in the test

Can you please add the test case to the PR, so we can have a look at the failing test.

xuxucode:
@amitaibu I added the test, please help run it

amitaibu:
Thanks, I'll give it a look.

amitaibu:
@edwardaa [this](https://gist.github.com/amitaibu/1bed7e60ff8f73cc9132) should fix it - it was a caching issue in the test itself, where the user account was cached, thus the token was never checked.

xuxucode:
Thanks, the test in the gist is passed with a fresh Drupal installation.

amitaibu:
@edwardaa can you please copy/ paste it into your branch, so we'll see if also Travis likes it.

xuxucode:
the build passed :-)

amitaibu:
> the build passed :-)

Well, what do you know ;) 

Thanks, merged!

e0ipso:
This has been ported to 7.x-2.x

PR #494, opened 2015-04-24
2.x Remove angular example

e0ipso:
:_(

Sad to see it go, but it needs to happen.


amitaibu:
I think we can move this to  a `restful_angular_example` repo under RESTful-drupal, no?

e0ipso:
That's a good idea. I don't think there's a need for a d.o project, though. Right?

amitaibu:
yeah, just on github would be enough

e0ipso:
Done. I even maintained the git history.

https://github.com/RESTful-Drupal/restful_angular_example

amitaibu:
:+1:

PR #495, opened 2015-04-24
Prevent preceding comma in Vary header.

chasingmaxwell:
This just prevents a preceding comma when the Vary header is empty originally.


amitaibu:
Sorry for the delay, but thanks!

PR #496, opened 2015-04-24
Removing typecasting on the argument of viewMultiple

makangus:
RestfulBase->process() passes the path into the controller as an argument. Typecasting the argument as an array will result in an error when a controller is defined as the following.

```
    return array(
      '^.*$' => array(
        \RestfulInterface::GET => 'viewMultiple',
      ),
    );
  }
```

Error:

```
Recoverable fatal error: Argument 1 passed to RestfulDataProviderSample::viewMultiple() must be of the type array, string given
```

This PR is not ready as RestfulDataProviderEFQ, RestfulDataProviderDbQuery and RestfulBase are still implementing RestfulDataProviderInterface, but I wanted to throw something up to see if there is something I am missing.

### DIFF
--- a/plugins/restful/RestfulDataProviderInterface.php
+++ b/plugins/restful/RestfulDataProviderInterface.php
@@ -24,13 +24,13 @@ interface RestfulDataProviderInterface {
   /**
    * View a collection of items.
    *
-   * @param array $ids
+   * @param mixed $ids
    *   An array of items to view.
    *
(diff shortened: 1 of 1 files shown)


e0ipso:
After talking to @makangus we decided to close this one out.

The example in the RestfulEntityBase is very misleading. We have taken a different approach in 2.x.

PR #497, opened 2015-04-26
Optional temp files

### LINKED ISSUE #462: RestfulFilesUpload makes files permanent by default
I'm trying to figure out how best to allow file uploads using the restful module. It seems like the pattern used by the `restful_angular_example` uploads the files then associates the `fid` with the entity that's saved later. 

One thing that's a little problematic about this approach is that every file uploaded is [automatically converted from a temporary file to a permanent one](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/RestfulFilesUpload.php#L73-L75):

``` php
// Change the file status from temporary to permanent.
$file->status = FILE_STATUS_PERMANENT;
file_save($file);
```

That seems a little strange to me because it seems very likely that you'll end up with orphaned files. The nice thing about leaving them as temporary files until they're referenced by another content type, is that they'll automatically be removed.

It's easy to demonstrate the how this can become a problem if you drop two or three images onto `restful_angular_example`'s form. You click `Submit` and only one image is attached to the entity. The other files won't be removed, but won't be used anywhere.

drewish:
Created this to fix #462. 

I need some help figuring out how to test this. I'm not sure if there's an easy way to change a plugin's options during a test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,4 +43,4 @@ before_script:
   - drush runserver 127.0.0.1:8080 &
   - until netstat -an 2>/dev/null | grep '8080.*LISTEN'; do true; done
 
-script: drush test-run RESTful --uri=http://127.0.0.1:8080
+script: drush test-run RestfulCreateEntityTestCase --uri=http://127.0.0.1:8080
--- a/plugins/restful/RestfulFilesUpload.php
+++ b/plugins/restful/RestfulFilesUpload.php
@@ -26,6 +26,8 @@ class RestfulFilesUpload extends \RestfulEntityBase {
    * - "validators": By default no validation is done on the file extensions or
    *   file size.
    * - "scheme": By default the default scheme (e.g. public, private) is used.
+   * - "make_permanent": When true the file's status will be set to permanent
+   *   and a file useage for restful module will be added.
    */
   public function __construct(array $plugin, \RestfulAuthenticationManager $auth_manager = NULL, \DrupalCacheInterface $cache_controller = NULL, $language = NULL) {
     parent::__construct($plugin, $auth_manager, $cache_controller, $language);
@@ -41,6 +43,7 @@ class RestfulFilesUpload extends \RestfulEntityBase {
       ),
       'scheme' => file_default_scheme(),
       'replace' => FILE_EXISTS_RENAME,
+      'make_permanent' => TRUE,
     );
 
     $this->setPluginKey('options', drupal_array_merge_deep($default_values, $options));
@@ -59,6 +62,7 @@ class RestfulFilesUpload extends \RestfulEntityBase {
       throw new \RestfulBadRequestException('No files sent with the request.');
     }
 
+    $options = $this->getPluginKey('options');
     $ids = array();
 
     foreach ($_FILES as $file_info) {
@@ -70,12 +74,14 @@ class RestfulFilesUpload extends \RestfulEntityBase {
 
       $file = $this->fileSaveUpload($name);
 
-      // Change the file status from temporary to permanent.
-      $file->status = FILE_STATUS_PERMANENT;
-      file_save($file);
+      if ($options['make_permanent']) {
+        // Change the file status from temporary to permanent.
+        $file->status = FILE_STATUS_PERMANENT;
+        file_save($file);
 
-      // Required to be able to reference this file.
-      file_usage_add($file, 'restful', 'files', $file->fid);
+        // Required to be able to reference this file.
+        file_usage_add($file, 'restful', 'files', $file->fid);
+      }
 
       $ids[] = $file->fid;
     }
--- a/tests/RestfulCreateEntityTestCase.test
+++ b/tests/RestfulCreateEntityTestCase.test
@@ -181,7 +181,6 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
   public function testFileUploadOptions() {
     variable_set('restful_file_upload', TRUE);
     variable_set('restful_file_upload_allow_anonymous_user', TRUE);
-    $handler = restful_get_restful_handler('files', 1, 0);
 
     // Upload a permanent file.
     $return = $this->uploadFile();
@@ -199,6 +198,12 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
 
 
     // Upload a temporary file
+    //drupal_static_reset('restful_get_restful_handler');
+    cache_clear_all();
+print_r(array_keys(restful_get_restful_plugins()));
+    $handler = restful_get_restful_handler('files', 1, 0);
+    $this->assertNotNull($handler);
+var_dump($handler);
     $options = $handler->getPluginKey('options');
     $options['make_permanent'] = FALSE;
     $handler->setPluginKey('options');
--- a/tests/RestfulCreateEntityTestCase.test
+++ b/tests/RestfulCreateEntityTestCase.test
@@ -178,31 +178,49 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
 
   }
 
+  public function testFileUploadSettings() {
+    variable_set('restful_file_upload', TRUE);
+    variable_set('restful_file_upload_allow_anonymous_user', TRUE);
+    // Test access for anonymous user (allowed).
+    $return = $this->uploadFile();
+    $this->assertEqual($return['code'], 200, 'File uploaded.');
+
+    $return_array = drupal_json_decode($return['body']);
+    $file = file_load($return_array['data'][0]['id']);
+
+    // Now that we have a successfully uploaded file, make sure it's the same
+    // file that was uploaded.
+    $original = hash_file('md5', $this->imagePath);
+    $uploaded = hash_file('md5', file_create_url($file->uri));
+    $this->assertEqual($original, $uploaded, 'Original and uploaded file are identical.');
+
+    $this->assertEqual($file->status, FILE_STATUS_PERMANENT, 'File status is permanent');
+
+    $usage = file_usage_list($file);
+    $this->assertEqual(reset($usage['restful']['files']), 1, 'File should be used by restful');
+  }
+
   /**
    * Test access for file upload.
    */
   public function testFileUploadAccess() {
     variable_set('restful_file_upload', TRUE);
-    variable_set('restful_file_upload_allow_anonymous_user', TRUE);
+
     // Test access for anonymous user (allowed).
+    variable_set('restful_file_upload_allow_anonymous_user', TRUE);
     $return = $this->uploadFile();
     $this->assertEqual($return['code'], 200, 'File upload is allowed to anonymous users.');
-    variable_set('restful_file_upload_allow_anonymous_user', FALSE);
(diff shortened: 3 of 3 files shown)


plugins/restful/RestfulFilesUpload.php
```diff
@@ -26,6 +26,8 @@ public static function controllersInfo() {
    * - "validators": By default no validation is done on the file extensions or
    *   file size.
    * - "scheme": By default the default scheme (e.g. public, private) is used.
+   * - "make_permanent": When true the file's status will be set to permanent
```
amitaibu, line 29: maybe this can be called `status` and the default value be `FILE_STATUS_PERMANENT`?

tests/RestfulCreateEntityTestCase.test
```diff
@@ -178,31 +178,49 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
...
+
     // Now that we have a successfully uploaded file, make sure it's the same
     // file that was uploaded.
     $original = hash_file('md5', $this->imagePath);
-    $return_array = drupal_json_decode($return['body']);
-    $file = file_load($return_array['data'][0]['id']);
     $uploaded = hash_file('md5', file_create_url($file->uri));
     $this->assertEqual($original, $uploaded, 'Original and uploaded file are identical.');
+
+    $this->assertEqual($file->status, FILE_STATUS_PERMANENT, 'File status is permanent');
```
amitaibu: missing dot in the end of the sentence

tests/RestfulCreateEntityTestCase.test
```diff
@@ -178,31 +178,49 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
...
+
+    $usage = file_usage_list($file);
+    $this->assertEqual(reset($usage['restful']['files']), 1, 'File should be used by restful');
```
amitaibu: Maybe better: `File is registered by RESTful.`

amitaibu:
> way to change a plugin's options during a test

`$this->setPluginKey();`

plugins/restful/RestfulFilesUpload.php
drewish, line 29: well it's not just the status, it's also the file reference which will prevent it from being deleted.

tests/RestfulCreateEntityTestCase.test
drewish: I went with `File usage is registered by RESTful.`

drewish:
I was a little confused by the `$this` but searching for `setPluginKey` turned up this pattern:

```
    $handler = restful_get_restful_handler('test_articles', 1, 1);
    $handler->setPluginKey('url_params', array(
      'filter' => TRUE,
      'sort' => FALSE,
      'range' => TRUE,
    ));
```

I'll give that a try.

tests/RestfulCreateEntityTestCase.test
```diff
@@ -178,31 +178,73 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
...
+    $this->assertEqual($file->status, FILE_STATUS_PERMANENT, 'File status is permanent.');
+    $usage = file_usage_list($file);
+    $this->assertEqual(reset($usage['restful']['files']), 1, 'File usage is registered by RESTful.');
+
+
+    // Upload a temporary file
+    //drupal_static_reset('restful_get_restful_handler');
+    cache_clear_all();
+print_r(array_keys(restful_get_restful_plugins()));
+    $handler = restful_get_restful_handler('files', 1, 0);
```
drewish, line 204: This is returning NULL but it's not obvious to me why. Locally I'm seeing `files` but Travis is not:

```
Array
(
    [0] => per_role_content__1_0:anonymous user
    [1] => per_role_content__1_0:authenticated user
    [2] => per_role_content__1_0:administrator
    [3] => csrf_token
    [4] => discovery__1_0
    [5] => login_cookie__1_0
    [6] => users__1_0
    [7] => node_user__1_0
    [8] => watchdog
    [9] => articles__1_0
    [10] => articles__1_1
    [11] => articles__1_2
    [12] => articles__1_4
    [13] => articles__1_5
    [14] => articles__1_6
    [15] => articles__1_7
    [16] => articles__2_0
    [17] => tags__1_0
    [18] => variable__1_0
    [19] => db_query_test__1_0
    [20] => entity_tests__1_0
    [21] => main__1_0
    [22] => main__1_1
    [23] => main__1_2
    [24] => main__1_3
    [25] => main__1_4
    [26] => main__1_5
    [27] => tests__1_0
    [28] => test_articles__1_0
    [29] => test_articles__1_1
    [30] => test_articles__1_2
    [31] => test_articles__1_3
    [32] => test_articles__1_4
    [33] => restful_test_translatable_entity__1_0
    [34] => test_tags__1_0
)
```

amitaibu, line 204: You might need to set the [variable](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/file/files_upload/1.0/files_upload__1_0.inc#L3) and clear menu cache

plugins/restful/RestfulFilesUpload.php
amitaibu, line 29: >  it's also the file reference which will prevent it from being deleted.

IMO, the file reference is an implementation detail to the fact you want the file to be permanent or temporary. By using the named constant `FILE_STATUS_PERMANENT` I think it makes it more readable and more inline with core's lingo.

.travis.yml
```diff
@@ -43,4 +43,4 @@ before_script:
   - drush runserver 127.0.0.1:8080 &
   - until netstat -an 2>/dev/null | grep '8080.*LISTEN'; do true; done
 
-script: drush test-run RESTful --uri=http://127.0.0.1:8080
+script: drush test-run RestfulCreateEntityTestCase --uri=http://127.0.0.1:8080
```
amitaibu, line 46: that's temporary, right? :)

drewish, line 46: Yeah the only way I can get the tests to run is on Travis and it was taking way to long to get to the failure.

plugins/restful/RestfulFilesUpload.php
drewish, line 29: It's been a while since I helped with the File API but I'm fairly certain they're more orthogonal. The usage is for reference counting to avoid deleting files in use. file_delete checks the references before deleting. So technically you could just skip setting the permanent status and add a usage, that would prevent it from being deleted but it would end up with a message in the logs every time cron ran. 

Adding the usage will prevent the file from being deleted if you're putting the fid into image or file fields, then remove it.

tests/RestfulCreateEntityTestCase.test
drewish, line 204: Doing that on line 182 then cleaning all caches on 202. Wondering if the handler options aren't shared between the server and test instances.

plugins/restful/RestfulFilesUpload.php
amitaibu, line 29: >  I'm fairly certain they're more orthogonal.

Maybe In that case we need to keys:

```
'status' => FILE_STATUS_PERMANENT,
// The name of the module to register the file under, or FALSE
'set_usage' => 'restful'
```

e0ipso, line 29: I like @amitaibu's suggestion. Of course, we will need to make `'set_usage'` to default to `'restful'`. That way we can omit it from the plugin definition.

tests/RestfulCreateEntityTestCase.test
drewish, line 204: @amitaibu any other ideas? i'd love to get this merged but your testing framework makes it tricky.

plugins/restful/RestfulFilesUpload.php
drewish, line 29: A static value like that would suffice for a just creating the reference to prevent deletion attempts. 

The only place core displays those values is when logging the failed deletion: https://github.com/drupal/drupal/blob/7.x/modules/system/system.module#L3047
But I think that at least shows the intention of the values.

If you look at the some of the places in core where they're set you'll see all three parameters (module, type, id) are used to indicate which module is using the file and how it's using it:
https://github.com/drupal/drupal/blob/7.x/modules/file/file.field.inc#L245 :

``` php
file_usage_add($file, 'file', $entity_type, $id);
```

https://github.com/drupal/drupal/blob/7.x/modules/image/image.module#L458 :

``` php
file_usage_add($file_new, 'image', 'default_image', $field['id']);
```

https://github.com/drupal/drupal/blob/7.x/modules/user/user.module#L491 :

``` php
file_usage_add($picture, 'user', 'user', $account->uid);
```

The idea there being that you could build links to the usages.

So I guess my point is really that the caller should probably be responsible for registering a usage. If you're putting those files into image or file fields, they'll be registered there automatically. If you're storing it elsewhere you should manage your own usage so the files can be deleted. 

That could be a big change in behavior so perhaps the change should only happen in the 2.x branch. The `set_usage` option could be there just for those who want to keep the 1.x behavior.

tests/RestfulCreateEntityTestCase.test
amitaibu, line 204: > any other ideas?

I'll have a look on my local

> your testing framework makes it tricky.

What do you mean? AFAIK it's regular simpletest -- what do you suggest to improve?

drewish, line 204: I could never get them to run locally and I saw you were using Travis so I tried to follow the steps it was using locally but no luck. The only way I've been able to run them was via Travis.

drewish, line 204: And it seems like Travis runs two processes for the tests.

amitaibu:
@drewish haven't forgotten about it, just taking some time -- DrupalCon preparations... :)

drewish:
I'm on vacation for a week so definitely not worried ;)

amitaibu:
@HelenaEksler can you please try to help and fix the test

tests/RestfulCreateEntityTestCase.test
```diff
@@ -178,31 +178,73 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
...
+    $this->assertNotNull($handler);
+var_dump($handler);
+    $options = $handler->getPluginKey('options');
+    $options['make_permanent'] = FALSE;
+    $handler->setPluginKey('options');
```
HelenaEksler, line 209: => `$handler->setPluginKey('options', $options);`

HelenaEksler:
What I've found out that if you change variable `restful_file_upload` value in test even with clearing the cache - plugin is still not available. Tried this process not in test - situation is the same, even after clearing cache right after changing the variable value (the same in both directions TRUE->FALSE and FALSE->TRUE).

e0ipso:
@HelenaEksler is that a testing problem, or a patch problem?

HelenaEksler:
@mateu-aguilo-bosch This is the testing problem. Once you've changed the variable - you can't change it back in the current flow. Not related to the patch.

e0ipso:

Thanks for everything!

drewish:
Sounds good. We decided to use a different module.

PR #500, opened 2015-04-28
499: Add Node Reference resource support.

iStryker:


e0ipso:
@iStryker I don't have any environment to test this. Is this working in the following scenarios:
- [ ] Listing entities with node reference fields.
- [ ] Viewing an entity without a resource key.
- [ ] Viewing an entity with the resource key and `full_view => FALSE`.
- [ ] Viewing an entity with the resource key and embedding the result of the target resource.
- [ ] Posting an entity with the info of the embedded resource to create both entities.

Thanks!

iStryker:
I have used it as a resource for a public field and it works.  I will try to test the other parameters in the next day or so.

e0ipso:
@iStryker any progress on this?

iStryker:
Code is now in production and has no problem.  I will try and test it with
the list you mention this week.

e0ipso:
That is awesome! Thanks for the response. I'll be monitoring this issue to hear from your tests.

e0ipso:
This has been merged.

Thanks for the contribution!

Issue #492, opened 2015-04-23
Null Empty values


iStryker:
Currently empty public field values are not getting outputted to the api.  After speaking with @nathanv, he said it is best practice for all of them to return NULL.    Should we be using this best practice?

e0ipso:
I don't hold a big preference in this matter. Maybe @nathanv can share some documentation to point to best practises in this case.

Thanks for pointing this out!

vespoli:
I think it is probably a best practice because as a consumer of the service, it is preferable to not have to always check for the existence of a field before using/checking it's value.

e0ipso:
Good!

Care to provide a PR? :-)

krisgraham:
I just created a PR for this. checkPropertyAccess() returns false for fields with no data (in addition to those which the user doesn't have permission to access), so now it just populates the value with the default NULL for either case.

iStryker:
Tested this with Entity Reference field and it works!

prudloff-insite:
In cas anyone stumbles on this, here is the PR mentioned by @krisgraham: https://github.com/RESTful-Drupal/restful/pull/505

PR #505, opened 2015-05-05
Return null for empty or unavailable fields - addresses issue #492

krisgraham:


amitaibu:
Not sure this approach is right, as if you don't have access to a field, you shouldn't know about it.

krisgraham:
@amitaibu That makes sense, but means you'll probably need a separate if() before the access check to specifically check for empty content. If that sounds alright, I'll create another PR for it.

amitaibu:
If the user has access and the content is empty - we can show the property.

e0ipso:
@krisgraham are you still working on this? May I provide direction?

Thanks.

e0ipso:

Thanks for everything!

PR #507, opened 2015-05-05
WIP: Check the property access after setting the values

ordavidil:


amitaibu:
Thanks @ordavidil I've added `WIP` to the title, as we'll still need to write tests for this.

amitaibu:
The test should mimic the case where `entity_metadata_no_hook_node_access()` is retuning FALSE. Trying to figure out a simple scenario..

amitaibu:
closed in favor of #539

PR #510, opened 2015-05-12
2.x multilingual tests

e0ipso:
This PR depends on #491.

PR #511, opened 2015-05-12
2.x json simple test

e0ipso:
This PR depends on #510.


e0ipso:
@amitaibu this test is also green.

Issue #515, opened 2015-05-15
Caches (that arent set) are deleted umpty times


mojzis:
I keep hoping i am not right and i miss a little detail somewhere ... 
We have an entity thats being updated rather often and has nothing to do with restful. We are dealing with plenty requests to `delete from restful_cache`, referencing that entity.
From the code it looks like on every entity update, a cache clear is attemtped for **every endpoint plugin** there is, resulting in loads of delete statements where there is nothing to delete. 
Some of the gotchas I didnt understand :
- in the [entity update hook](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/restful.entity.inc#L40), should we do it for _every entity_ without a possibility to skip some ? (and twice for users)
- later in [RestfulManager::invalidateEntityCache](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/includes/RestfulManager.php#L288) its going through all plugins disregarding which entity has just been updated
- and in [RestfulBase->cacheInvalidate](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/RestfulBase.php#L1394) we invalidate even though `$cache_info['render']` [is set to false](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/includes/RestfulManager.php#L132)

Please prove me wrong :) Otherwise lets discuss how to improve this ... 
Thanks :)

PR #517, opened 2015-05-18
don't delete cache when its not set #515

mojzis:


e0ipso:
Well, that makes a lot of sense.

Thanks for the contribution.

Issue #521, opened 2015-05-27
Add watchdog message upon \REstfulException


amitaibu:
Currently our exceptions are inside try/catch so they never reach the logs

PR #522, opened 2015-05-27
Add watchdog entry if excpetion was thrown

amitaibu:
#521

### DIFF
--- a/restful.module
+++ b/restful.module
@@ -528,8 +528,16 @@ function restful_menu_process_callback($resource_name, $version = NULL) {
       drupal_add_http_header($header_name, $header_value);
     }
 
-    $severity = $e->getCode() < 500 ? WATCHDOG_NOTICE : WATCHDOG_ERROR;
-    watchdog('restful', $e->getMessage(), array(), $severity);
+    if ($e->getCode() < 500) {
+      // Even though it's an exception, it's in fact not a server error - it
+      // might be just access denied, or a bad request, so we just want to log
+      // it, but without marking it as an actual exception.
+      watchdog('restful', $e->getMessage());
+    }
+    else {
+      // This one should really be considered as an error.
+      watchdog_exception('restful', $e);
+    }
   }
   catch (Exception $e) {
     $result = array(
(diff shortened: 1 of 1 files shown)


restful.module
```diff
@@ -527,6 +527,9 @@ function restful_menu_process_callback($resource_name, $version = NULL) {
     foreach ($e->getHeaders() as $header_name => $header_value) {
       drupal_add_http_header($header_name, $header_value);
     }
+
+    $severity = $e->getCode() < 500 ? WATCHDOG_NOTICE : WATCHDOG_ERROR;
+    watchdog('restful', $e->getMessage(), array(), $severity);
```
e0ipso: We should probably use `watchdog_exception`.

amitaibu: Didn't know about `watchdog_exception` :) I think we should use it only for the below case and when status code >= 500.

restful.module
```diff
@@ -527,6 +527,17 @@ function restful_menu_process_callback($resource_name, $version = NULL) {
     foreach ($e->getHeaders() as $header_name => $header_value) {
       drupal_add_http_header($header_name, $header_value);
     }
+
+    if ($e->getCode() < 500) {
```
e0ipso, line 531: What if we simplify this as:

``` php
// Even though it's an exception, it's in fact not a server error - it
// might be just access denied, or a bad request, so we just want to log
// it, but without marking it as an actual exception.
// If the code is 5XX then it is an actual server exception.
$severity = $e->getCode() < 500 ? WATCHDOG_NOTICE : WATCHDOG_ERROR;
watchdog_exception('restful', $e, NULL, array(), $severity);
```

amitaibu, line 531: the "problem" with watchdog_exception() is that it adds the line number where exception was thrown. However for exceptions < 500, we don't really need that info, since it's not really a server error, but just the way RESTful is taking advantage of exceptions to short circuit the flow.

e0ipso, line 531: Works for me.

e0ipso:
This was ported to 7.x-2.x in #726.

PR #524, opened 2015-05-29
2.x Fix authentication tests

e0ipso:
This will bring back the authentication tests.

### DIFF
--- a/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -280,7 +280,7 @@ class ResourceFieldEntity implements ResourceFieldEntityInterface {
       return FALSE;
     }
 
-    $access = $property_wrapper->access($op, $account);
+    $access = $interpreter->getWrapper()->value() !== FALSE && $property_wrapper->access($op, $account);
     return $access !== FALSE;
   }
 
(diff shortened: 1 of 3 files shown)


e0ipso:
@amitaibu tests are green.

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -280,7 +280,7 @@ public function access($op, DataInterpreterInterface $interpreter) {
       return FALSE;
     }
 
-    $access = $property_wrapper->access($op, $account);
+    $access = $interpreter->getWrapper()->value() !== FALSE && $property_wrapper->access($op, $account);
```
amitaibu, line 288: `$interpreter->getWrapper()->value() !== FALSE` - Why is this needed?

amitaibu:
added a small question - rest looks good.

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 288: If `$interpreter->getWrapper()->value() === FALSE` it means that the entity could not be loaded, thus checking properties on it will result in errors.

Ex: this happens when the embedded author is the anonymous user. Doing `user_load(0)` returns `FALSE`.

amitaibu, line 288: Ok, I think it's worth adding this as a comment.

e0ipso, line 288: Fixed in 9bdcd44.

e0ipso:
@amitaibu please merge.

PR #527, opened 2015-05-29
Add documentation for adding a sort property using a URL parameter

edhollinghurst:
I think this would be useful to include.


amitaibu:
Thanks!

PR #530, opened 2015-06-01
in entity update / delete hooks don’t invalidate the cache for user

mojzis:
... and other entities #515

As it is, user is invalidated twice - in the entity hook as well as in the user hook. This is an attempt to invalidate the user only once.
Also, a site may contain many entity types, where restful and its cache (and cache invalidation) is not going to be used at all. Probably the fastest is to allow the developer to set those in a variable, instead of going to the plugin definitions.

### DIFF
--- a/restful.entity.inc
+++ b/restful.entity.inc
@@ -35,18 +35,40 @@ function restful_entity_info() {
  * Implements hook_entity_update().
  */
 function restful_entity_update($entity, $type) {
-  list($entity_id) = entity_extract_ids($type, $entity);
-  $cid = 'paet:' . $type . '::ei:' . $entity_id;
-  \RestfulManager::invalidateEntityCache($cid);
+  if(restful_entity_good_for_invalidation($type)) {
+    list($entity_id) = entity_extract_ids($type, $entity);
+    $cid = 'paet:' . $type . '::ei:' . $entity_id;
+    \RestfulManager::invalidateEntityCache($cid);
+  }
 }
 
 /**
  * Implements hook_entity_delete().
  */
 function restful_entity_delete($entity, $type) {
-  list($entity_id) = entity_extract_ids($type, $entity);
-  $cid = 'paet:' . $type . '::ei:' . $entity_id;
-  \RestfulManager::invalidateEntityCache($cid);
+  if(restful_entity_good_for_invalidation($type)) {
+    list($entity_id) = entity_extract_ids($type, $entity);
+    $cid = 'paet:' . $type . '::ei:' . $entity_id;
+    \RestfulManager::invalidateEntityCache($cid);
+  }
+}
+
+/**
+ * Helper function, used in the entity hooks to determine whether entities are supposed to be cached or not.
+ * 
+ * @param $type entity type - string
+ * @return bool 
+ */
+
+function restful_entity_good_for_invalidation($type) {
+  $skip_types = array(
+    // Saves to the configuration entity dont necesarrily need to be cached & invalidated.   
+      'rate_limit',
+    // User is being delt with separately.
+    'user'
+  );
+  $skip_types = array_merge($skip_types, variable_get('restful_entities_to_skip',array()));
+  return ! in_array($type, $skip_types);
 }
 
 /**
--- a/restful.entity.inc
+++ b/restful.entity.inc
@@ -35,7 +35,7 @@ function restful_entity_info() {
  * Implements hook_entity_update().
  */
 function restful_entity_update($entity, $type) {
-  if(restful_entity_good_for_invalidation($type)) {
+  if (restful_entity_determine_invalidation_by_type($type)) {
     list($entity_id) = entity_extract_ids($type, $entity);
     $cid = 'paet:' . $type . '::ei:' . $entity_id;
     \RestfulManager::invalidateEntityCache($cid);
@@ -46,7 +46,7 @@ function restful_entity_update($entity, $type) {
  * Implements hook_entity_delete().
  */
 function restful_entity_delete($entity, $type) {
-  if(restful_entity_good_for_invalidation($type)) {
+  if (restful_entity_determine_invalidation_by_type($type)) {
     list($entity_id) = entity_extract_ids($type, $entity);
     $cid = 'paet:' . $type . '::ei:' . $entity_id;
     \RestfulManager::invalidateEntityCache($cid);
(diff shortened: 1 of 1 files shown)


amitaibu:
@HelenaEksler please review

restful.entity.inc
```diff
@@ -35,18 +35,40 @@ function restful_entity_info() {
...
+ * Helper function, used in the entity hooks to determine whether entities are supposed to be cached or not.
+ * 
+ * @param $type entity type - string
+ * @return bool 
+ */
+
+function restful_entity_good_for_invalidation($type) {
+  $skip_types = array(
+    // Saves to the configuration entity dont necesarrily need to be cached & invalidated.   
+      'rate_limit',
```
HelenaEksler: Please, fix indentation.

restful.entity.inc
```diff
@@ -35,18 +35,40 @@ function restful_entity_info() {
...
+/**
+ * Helper function, used in the entity hooks to determine whether entities are supposed to be cached or not.
+ * 
+ * @param $type entity type - string
+ * @return bool 
+ */
+
+function restful_entity_good_for_invalidation($type) {
+  $skip_types = array(
+    // Saves to the configuration entity dont necesarrily need to be cached & invalidated.
```
HelenaEksler: `dont` => `doesn't`
`necesarrily` => `necessarily`

restful.entity.inc
```diff
@@ -35,18 +35,40 @@ function restful_entity_info() {
...
+ * 
+ * @param $type entity type - string
+ * @return bool 
+ */
+
+function restful_entity_good_for_invalidation($type) {
+  $skip_types = array(
+    // Saves to the configuration entity dont necesarrily need to be cached & invalidated.   
+      'rate_limit',
+    // User is being delt with separately.
```
HelenaEksler: `delt` => `dealt`

restful.entity.inc
```diff
@@ -35,18 +35,40 @@ function restful_entity_info() {
...
+    'user'
+  );
+  $skip_types = array_merge($skip_types, variable_get('restful_entities_to_skip',array()));
+  return ! in_array($type, $skip_types);
```
HelenaEksler: `! in_array($type, $skip_types);` => `!in_array($type, $skip_types);`

restful.entity.inc
```diff
@@ -35,18 +35,40 @@ function restful_entity_info() {
...
+    list($entity_id) = entity_extract_ids($type, $entity);
+    $cid = 'paet:' . $type . '::ei:' . $entity_id;
+    \RestfulManager::invalidateEntityCache($cid);
+  }
+}
+
+/**
+ * Helper function, used in the entity hooks to determine whether entities are supposed to be cached or not.
+ * 
+ * @param $type entity type - string
```
HelenaEksler: Variable description on the next line.

restful.entity.inc
```diff
@@ -35,18 +35,40 @@ function restful_entity_info() {
...
+ * @return bool 
+ */
+
+function restful_entity_good_for_invalidation($type) {
```
HelenaEksler: Please, change the name of the function:
`restful_entity_determine_invalidation_by_type($entity_type)`

mojzis:
@HelenaEksler thanks ! changes applied :)

restful.entity.inc
mojzis: it was actually meant as plural, probably needs rewording to be more understandable ?

restful.entity.inc
```diff
@@ -35,18 +35,47 @@ function restful_entity_info() {
...
+/**
+ * Helper function, used in the entity hooks.
+ * 
+ * To determine whether entities are supposed to be cached or not.
+ * 
+ * @param $type entity type - string
+ *   Entity type to be tested.
+ * @return bool 
+ */
+
```
HelenaEksler, line 65: No need in this line break.

restful.entity.inc
```diff
@@ -35,18 +35,47 @@ function restful_entity_info() {
...
+    \RestfulManager::invalidateEntityCache($cid);
+  }
+}
+
+/**
+ * Helper function, used in the entity hooks.
+ * 
+ * To determine whether entities are supposed to be cached or not.
+ * 
+ * @param $type entity type - string
```
HelenaEksler: You can keep one description. 

``` php
* @param $entity_type
*   Entity type to be tested.
```

restful.entity.inc
```diff
@@ -35,18 +35,47 @@ function restful_entity_info() {
...
+ *   Entity type to be tested.
+ * @return bool 
+ */
+
+function restful_entity_determine_invalidation_by_type($type) {
```
HelenaEksler: It's worth to change the parameter name to `$entity_type`.

restful.entity.inc
mojzis, line 65: @HelenaEksler sorry, not sure what you meant there ?

mojzis:
should i squash it into 1 ?

restful.entity.inc
```diff
@@ -35,18 +35,47 @@ function restful_entity_info() {
...
-  \RestfulManager::invalidateEntityCache($cid);
+  if (restful_entity_determine_invalidation_by_type($type)) {
+    list($entity_id) = entity_extract_ids($type, $entity);
+    $cid = 'paet:' . $type . '::ei:' . $entity_id;
+    \RestfulManager::invalidateEntityCache($cid);
+  }
+}
+
+/**
+ * Helper function, used in the entity hooks.
```
amitaibu, line 57: Change to: `Determine if an entity's cache should be invalidated.`

restful.entity.inc
```diff
@@ -35,18 +35,47 @@ function restful_entity_info() {
...
+ * 
+ * To determine whether entities are supposed to be cached or not.
+ * 
+ * @param string $entity_type
+ *   Entity type to be tested.
+ * @return bool
```
amitaibu, line 63: add a line break before, and doc to the return value, i.e.:

```
 * @param string $entity_type
 *   Entity type to be tested.
 *
 * @return bool
 *   True if the entity's cache should be invalidated.
```

restful.entity.inc
```diff
@@ -35,18 +35,47 @@ function restful_entity_info() {
...
+  $skip_types = array(
+    // Saves to the configuration entity don't necessarily need to be
+    // cached & invalidated.   
+    'rate_limit',
+    // User is being dealt with separately.
+    'user'
+  );
+  $skip_types = array_merge(
+    $skip_types, 
+    variable_get('restful_entities_to_skip', array())
```
amitaibu, line 76: Do we really need this whole function -- why don't we just check the `user` entity type?

amitaibu:
btw, no need to squash, we always `merge squash` and credit the author.

restful.entity.inc
mojzis, line 76: The idea was that people could add their entities that don't need to go through the invalidation loop. ie stuff that has nothing to do with their rest API. I thought the variable would be faster then going through the plugins.

e0ipso, line 76: I feel that if this `'restful_entities_to_skip'` is not documented, it will not be used. We should probably add a line to the documentation.

After that this PR is ready to be merged.

PR #535, opened 2015-06-16
Allow overriding entity cache parameters

Kazanir:
We're putting together a site where the language negotiation is done via an HTTP `Accept-Language:` header, but this resulted in cache conflicts since there were no parameters in the URL on which the differing language values could be cached. Unfortunately, with the way `viewEntity()` is structured we had to copy/paste the entire function just to change the cache tagging -- it seems better to allow this to be overridden at the plugin layer so I did a small re-factoring to support this.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -350,7 +350,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
       $values[$public_field_name] = $value;
     }
 
-    $this->setRenderedCache($values, array($this->entityCacheTags());
+    $this->setRenderedCache($values, $this->entityCacheTags());
     return $values;
   }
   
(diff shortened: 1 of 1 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -353,11 +350,25 @@ public function viewEntity($id) {
...
+   * The array of parameters by which entities should be cached.
+   *
+   * @param mixed
+   *   The entity ID of the entity to be cached.
+   * 
+   * @return array
+   *   An array of parameter keys and values which should be added
+   *   to the cache key for each entity.
+   */
+  public function entityCacheTags($entity_id) {
```
amitaibu, line 367: lets rename this to `getEntityCacheTags`.

Kazanir:
I think changing the invalidation to account for this is probably the best idea -- if no one tells me that's too annoying to try then I will go look at how to accomplish it. :)

e0ipso:
@Kazanir thanks for giving this a shot. Let me know if you hit a dead end.

e0ipso:
I made some changes and merged this.

Thanks for contributing.

e0ipso:
This is no longer necessary in 7.x-2.x since we can override the cache fragments (https://www.youtube.com/watch?v=TL-y1M_1gao&list=PLZOQ_ZMpYrZv8_c7jd_CkO_93-DnyVFY5&index=23). Removing tag.

PR #539, opened 2015-06-21
Checking property settings only after setting the value

amitaibu:
#507

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -675,13 +675,15 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
         continue;
       }
 
+      $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
+      $wrapper->{$property_name}->set($field_value);
+
+      // We check the property access only after setting the values, as the
+      // access callback's response might change according to the field value.
       if (!$this->checkPropertyAccess('edit', $public_field_name, $wrapper->{$property_name}, $wrapper)) {
         throw new \RestfulBadRequestException(format_string('Property @name cannot be set.', array('@name' => $public_field_name)));
       }
 
-      $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
-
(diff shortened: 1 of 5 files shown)


amitaibu:
Lets see what Travis thinks about this... I was able to see the tests fails on my local without the necessary code change, and green after.

amitaibu:
Ready for review.

plugins/restful/RestfulEntityBase.php
```diff
@@ -675,13 +675,15 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
         continue;
       }
 
+      $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
+      $wrapper->{$property_name}->set($field_value);
+
+      // We check the property access only after setting the values, as the
+      // access callback's response might change according to the field value.
```
e0ipso, line 682: :+1:

tests/RestfulCreatePrivateNodeTestCase.test
```diff
@@ -0,0 +1,66 @@
...
+    $node1 = $this->drupalCreateNode($settings);
+
+    // Assert user has access to the node.
+    $this->assertTrue(node_access('view', $node1, $user1), 'Author has access to view node.');
+
+    // Assert another user doesn't have access to the node.
+    $this->assertFalse(node_access('view', $node1, $user2), 'Authenticated user, but not author does not have access to view the node.');
+
+    $handler = restful_get_restful_handler('test_articles', 1, 2);
+    $handler->setAccount($user1);
```
e0ipso, line 55: I'm not sure I'm understanding the use case here, but shouldn't this be `$user2`?

amitaibu, line 55: no, the issue is that the user that really has access to node1, couldn't set a reference to it, because `restful_node_access_test_node_access_records()` got an empty reference and returned false.

`$user2` is there just to "prove" that the node is indeed "private"

e0ipso, line 55: Oooh. I see.

e0ipso:
This has been ported to 7.x-2.x in #727. Removing tag.

PR #540, opened 2015-07-05
Bug fixes to RestfulDataProviderDbQuery

onumossn:
As the view function was not calling getQuery, it would not include properties from joins that someone may do. If changes were made, it would be exactly the same as viewMultiple other than one accepts arrays, so I just called viewMultiple.

As a value was required to be not falsy for process_callbacks to be called, it would not allow processing falsy values. For example, certain tables in Drupal appear to use "1" and "0" for true and false values. Someone may want to convert those "1" and "0" to true and false before returning the data, but "0" is falsy and process_callback could not be called.

### DIFF
--- a/plugins/restful/RestfulDataProviderDbQuery.php
+++ b/plugins/restful/RestfulDataProviderDbQuery.php
@@ -360,8 +360,8 @@ abstract class RestfulDataProviderDbQuery extends \RestfulBase implements \Restf
   /**
    * {@inheritdoc}
    */
-  public function view($id) {
-    return $this->viewMultiple(array($id));
+  public function view($ids) {
+    return $this->viewMultiple(explode(',', $ids);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


e0ipso:
Thanks @mohammed-nauage-gsa can you see why tests are failing?

onumossn:
Fixed the tests. I am sorry about that. I had a hard time trying to setup testing locally. I am pretty new to php and drupal. Do things look ok now?

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -581,7 +554,7 @@ public function mapDbRowToPublicFields($row) {
       }
 
       // Execute the process callbacks.
-      if ($value && $info['process_callbacks']) {
+      if ($info['process_callbacks']) {
```
amitaibu: maybe we should check `!is_null()`? I don't think we should process NULL values

onumossn:
I think its debatable, but I think that makes sense in most cases that I can imagine, so I updated it with isset since !is_null === isset.

Also, I noticed viewMultiple doesn't ever seem to actually be used beyond how I used it in the this file. What do you think of removing it or should it stay to make sure a defined coding/naming pattern is kept?

onumossn:
Are there any other issues that need to be resolved with this pull requests?

e0ipso:
@mohammed-nauage-gsa sorry for the delay. I have been working intensely in the `2.x` branch.

e0ipso:
Merged.

Thanks again!

PR #542, opened 2015-07-09
2.x Update plug

e0ipso:
@amitaibu you will not believe how long it took to get to this point. `Plug` only depends on `Registry Autoload` now, so there is no need for ~~Composer Manager~~ or ~~XAutoload~~.

Since Travis has changed a bit I don't expects tests to be green right away.


e0ipso:
@amitaibu YAY!

After much playing with Travis-CI this is green. Please, merge this to fix 7.x-2.x (which is broken without this).

.travis.yml
```diff
@@ -60,33 +61,23 @@ script:
...
   - patch -p1 $TRAVIS_BUILD_DIR/../drupal/sites/all/modules/entity/modules/callbacks.inc  < 2086225-entity-access-check-node-create-3.patch
 
   # Enable the RESTful modules
-  - drush en --yes simpletest
-  - drush --yes pm-enable restful restful_token_auth
+  - drush --yes pm-enable simpletest restful restful_token_auth
   # Run the tests
-  - drush test-run RESTful --uri=http://localhost
+  - cd $TRAVIS_BUILD_DIR/../drupal
+  - php ./scripts/run-tests.sh --php $(which php) --verbose --color --concurrency 4 --url http://localhost RESTful | tee /tmp/simpletest-result.txt
```
amitaibu, line 81: this part can benefit from a comment or two :) - why not use `drush test-run`?  Otherwise, feel free to merge!

e0ipso, line 81: The problem is that drush discontinued `drush test-run` in modern drush (the current stable release).

e0ipso:
Merged.

PR #548, opened 2015-07-17
2.x Entity create test

e0ipso:
This will bring back the _Entity create test_.

Fixing this was specially challenging due to the fact that it's not possible to simulate a scenario that `is_uploaded_file` can swallow, so I had to add a variable to avoid mocking the whole class for the tests.


restful.module
```diff
@@ -93,7 +93,9 @@ function restful_menu() {
     ->getResourceManager()
     ->getPlugins();
   foreach ($plugins->getIterator() as $plugin) {
-    /** @var ResourceInterface $plugin */
+    if (!$plugin instanceof ResourceInterface) {
+      continue;
```
amitaibu, line 99: worth a comment

src/Plugin/resource/CsrfToken.php
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Class CsrfToken
+ * @package Drupal\restful\Plugin\resource
+ *
+ * @Resource(
+ *   name = "csrf_token:1.0",
+ *   resource = "csrf_token",
+ *   label = "CSRF Token",
+ *   description = "Resource that provides CSRF Tokens.",
```
amitaibu: maybe `Resource that provides CSRF Tokens when using cookie authentication.`

src/Plugin/resource/CsrfToken.php
```diff
@@ -0,0 +1,109 @@
...
+        'callback' => '\Drupal\restful\Plugin\resource\CsrfToken::getCsrfToken',
+      ),
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function dataProviderFactory() {
+    return new DataProvider\DataProviderNull($this->getRequest(), $this->getFieldDefinitions(), $this->getAccount(), array());
```
amitaibu: If I understand correctly we use `DataProviderNull` because we calculate this on the fly. If so maybe worth adding a comment.

Second thing is that it's very verbose considering we just want a null implementation. Maybe we can change the signature of  `\Drupal\restful\Plugin\resource\DataProvider\DataProvider::__construct(RequestInterface $request = UNLL, ResourceFieldCollectionInterface $field_definitions = NULL, $account = NULL, array $options = NULL, $langcode = NULL);``

So here we could just `return new DataProvider\DataProviderNull();`

src/Plugin/resource/CsrfToken.php
```diff
@@ -0,0 +1,109 @@
...
+    $account = $this->getAccount();
+    return (bool) $account->uid && parent::access();
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function index($path) {
+    $values = array();
+    foreach ($this->publicFields() as $public_property => $info) {
```
amitaibu, line 83: Again ,probably out of the scope of this PR, but I believe this part is being duplicated. Could we have another service take care of this iteration and callback execution?

src/Plugin/resource/DataProvider/DataProviderFile.php
```diff
@@ -293,4 +294,33 @@ protected function checkUploadErrors($source, array $files) {
...
+   * Helper function that checks if a file was uploaded via a POST request.
+   *
+   * @param string $filename
+   *   The name of the file.
+   *
+   * @return bool
+   *   TRUE if the file is uploaded. FALSE otherwise.
+   */
+  protected static function isUploadedFile($filename) {
+    return variable_get('restful_insecure_uploaded_flag', FALSE) || is_uploaded_file($filename);
```
amitaibu: shouldn't it be `(variable_get('restful_insecure_uploaded_flag', FALSE) && !empty($_FILES)) || ...`

amitaibu: actually. This is opening a can of worms :) ppl might start abusing `$_FILES`.

How about adding this hack on an extended class in the restful_test module?

tests/RestfulCreateEntityTestCase.test
```diff
@@ -94,7 +97,9 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
       'image_multiple' => array($images[0], $images[1]),
     );
 
-    $response = $handler->post('', $request);
+    $handler->setRequest(Request::create('', array(), RequestInterface::METHOD_POST, NULL, FALSE, NULL, array(), array(), array(), $request));
+    $handler->setPath('');
```
amitaibu, line 102: btw, is it being explicit, or is it required.?

tests/RestfulCreateEntityTestCase.test
```diff
@@ -186,35 +202,70 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
...
   }
 
   /**
    * Uploads a file issuing a POST HTTP request.
+   *
+   * @return \Drupal\restful\Plugin\resource\ResourceInterface
+   *   The file resource.
    */
-  protected function uploadFile() {
+  protected function fileResource() {
```
amitaibu, line 240: I'm sure the below ~10 lines took much more time then it seems... good job! :)

tests/RestfulCreateEntityTestCase.test
```diff
@@ -223,11 +274,21 @@ class RestfulCreateEntityTestCase extends RestfulCurlBaseTestCase {
...
 
-    return $this->httpRequest('api/file-upload', \RestfulInterface::POST, array(
-      'filename' => $value,
-    ), array(
-      'Content-Type' => 'multipart/form-data',
-    ), TRUE);
+    $tmp_name = drupal_tempnam('/tmp', 'restful_test_');
+    file_put_contents($tmp_name, file_get_contents($this->imagePath));
+    $headers = new \Drupal\restful\Http\HttpHeaderBag(array('Content-Type' => 'multipart/form-data'));
+    $files = array(
```
amitaibu, line 277: `// Mock the $_FILES global variable.`

amitaibu:
Added a few comments. I really like how I'm able to read through 2.x code and "simply understand it" :)

tests/RestfulCreateEntityTestCase.test
e0ipso, line 240: Aw yeah. There's some sweat over those lines.

tests/RestfulCreateEntityTestCase.test
e0ipso, line 102: It's required.

tests/RestfulCreateEntityTestCase.test
e0ipso, line 277: :+1:

src/Plugin/resource/DataProvider/DataProviderFile.php
e0ipso: That is a great idea! :+1:

src/Plugin/resource/CsrfToken.php
e0ipso: I refactored a bit. The base resource had an abstract `dataProviderFactory` method until now. That meant that any resource you wrote from scratch would need to declare its data provider. Instead of that I made 2 changes:
- `DataProviderNull` is the default data provider for resources written from scratch.
- There is a new protected method `dataProviderClassName` that can be overwritten on those resources. If you want a different data provider for your resource, just return the class in that method.

src/Plugin/resource/CsrfToken.php
e0ipso, line 83: There's a TODO in the code (for the parent class) to do just this.

e0ipso:
@amitaibu tests are green. Feel free to merge away.

PR #562, opened 2015-07-22
2.x Fix create node tests

e0ipso:
This PR depends on #561. Please, merge that one first.


amitaibu:
@mateu-aguilo-bosch needs a re-roll

e0ipso:
@amitaibu done.

restful.module
```diff
@@ -451,7 +451,7 @@ function _restful_get_data_from_menu_status(&$var) {
 
     case MENU_ACCESS_DENIED:
       $class_name = '\Drupal\restful\Exception\ForbiddenException';
-      $message = 'Access denied.';
+      $message = 'Access denied while checking menu access.';
```
amitaibu: I wouldn't change it. "menu access" is very Drupal. It's simply access denied ;)

src/Authentication/AuthenticationManager.php
```diff
@@ -109,7 +109,7 @@ public function getAccount(RequestInterface $request, $cache = TRUE) {
         drupal_page_is_cacheable(variable_get('restful_page_cache', FALSE));
 
         // User didn't authenticate against any provider, so we throw an error.
-        throw new UnauthorizedException('Bad credentials');
+        throw new UnauthorizedException('Anonymous user resolved for a resource that requires authentication.');
```
amitaibu: "Bad credentials" should remain there (possible with some more explanation) - since this is telling us credentials were sent, but not correct. It's not just trying to access a resource as anon.

amitaibu:
Minor comments, after that you can merge. Thanks.

restful.module
e0ipso: :+1:

src/Authentication/AuthenticationManager.php
e0ipso: :+1:

PR #564, opened 2015-07-23
2.x Cleanup old stuff

e0ipso:
Remove code from `7.x-1.x` that is present in `7.x-2.x`.


e0ipso:
That code was never run, it belonged to 7.x-1.x.

e0ipso:
If we ever need any of this back we can get it from git's history or 7.x-1.x branch.

PR #567, opened 2015-07-24
2.x Refactor rendering system

e0ipso:
The rendering (transforming the configuration into strings and arrays) has been moved from the DataProvider to the Formatter. That means that we can do the needed modifications when we need to.

There is the undesired side effect of the RenderCache not being useful on the DataProvider, so it will need to be refactored. I disabled it meanwhile.


amitaibu:
Sorry for loosing track of this. I'm out for vacation tomorrow, so lots of preparation...

e0ipso:
No worries! I needed this to be merged and took the liberty to do it. Please do not hesitate to open any follow ups.

amitaibu:
:+1:

PR #568, opened 2015-07-24
Allow files to be uploaded through the PUT method.

mpetrovic:
From #483.


e0ipso:
Thank you for pushing things forward. I feel that we are close!

I'm not comfortable merging this without proper testing. Would you be up for writing some Simpletests?

Thanks for your contribution!

mpetrovic:
@e0ipso 

I'm not that familiar with simple-tests. The test case I added isn't being run and I don't see why. Do you see anything?

PR #574, opened 2015-07-27
Better documentation at restful.admin.inc

theodorosploumis:
Provides a better documentation. In fact it maybe needs some more details.
Also changed the url to the DateInterval to the more appropriate **[dateinterval.createfromdatestring](php.net/manual/en/dateinterval.createfromdatestring.php)**.


e0ipso:
This was ported to 7.x-2.x in #729. Removing tag.

PR #580, opened 2015-08-05
Use the resource path

e0ipso:
@amitaibu the request path contains `v1.0/resource-name`, so we need to use the resource path to check that it's a listing.

I added a follow up to add testing for this case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,9 @@ script:
   - drush --yes pm-enable simpletest restful restful_token_auth
   # Run the tests
   - cd $TRAVIS_BUILD_DIR/../drupal
-  - php ./scripts/run-tests.sh --php $(which php) --verbose --color --concurrency 4 --url http://localhost RESTful | tee /tmp/simpletest-result.txt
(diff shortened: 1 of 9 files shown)


amitaibu:
I'd expect `$this->request->getPath()` to return the path without the versioning. So result from `getPath()` would be the same if version passed via URL or header.

amitaibu:
Also, you might have custom menu which doesn't correlate with the resource name.

e0ipso:
I tried that approach and found several problems:
1. The `Request` object should not know about the URL templates for the resources.
2. That would only work if we did not allow having custom URLs for the resources (see `'api/login'`).
3. Having the _path inside of a resource_ (like `/1234` for the articles resource) only makes sense in the context of a resource object, not the global `Request`.

That's why there is a separation between the request path and the resource path. Also, that is why we do `$handler->setPath('1234')` in the tests.

amitaibu:
ok, I thought the request path will change based on URL/ header of the versioning.

Test is failing, but seems legit :)

e0ipso:
It turns out we had a couple of broken tests that were firing a false positive :-/

Fixed now :-)

e0ipso:
@amitaibu this is ready for re-review.

e0ipso:
Merged. Added postmortem review tag.

PR #577, opened 2015-07-30
2.x JSON API Formatter tests

e0ipso:
This PR depends on #572. Please, merge that one first.

This will add testing to the JSON API formatter. This also includes several bugfixes!

PR #582, opened 2015-08-06
2.x Bring back DB Query providers

e0ipso:
This PR depends on #577. Pleas, merge that one first.


e0ipso:
@amitaibu I need db query to be merged, so adding to the postmortem list. I promise this is the last one.

PR #584, opened 2015-08-06
Add the option to use dash instead of underscore on the key name.

ordavidil:
We should let RESTful accept key names either with `-` or `_` .

### DIFF
--- a/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
+++ b/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
@@ -12,9 +12,11 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   public function applies(array $request = array(), $method = \RestfulInterface::GET) {
     $options = $this->getPluginKey('options');
     $key_name = !empty($options['param_name']) ? $options['param_name'] : 'access_token';
+    $dashed_key_name = str_replace('_', '-', $key_name);
 
-    // Access token may be on the request, or in the headers.
-    return !empty($request['__application'][$key_name]) || !empty($request[$key_name]);
+    // Access token may be on the request, or in the headers
+    // (may be a with dash instead of underscore).
+    return !empty($request['__application'][$key_name]) || !empty($request[$key_name]) || !empty($request['__application'][$dashed_key_name]) || !empty($request[$dashed_key_name]);
   }
 
   /**
@@ -23,7 +25,20 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   public function authenticate(array $request = array(), $method = \RestfulInterface::GET) {
     $options = $this->getPluginKey('options');
     $key_name = !empty($options['param_name']) ? $options['param_name'] : 'access_token';
-    $token = !empty($request['__application'][$key_name]) ? $request['__application'][$key_name] : $request[$key_name];
+    $dashed_key_name = str_replace('_', '-', $key_name);
+
+    if (!empty($request['__application'][$key_name])) {
+      $token = $request['__application'][$key_name];
+    }
+    elseif (!empty($request[$key_name])) {
+      $token = $request[$key_name];
+    }
+    elseif (!empty($request['__application'][$dashed_key_name])) {
+      $token = $request['__application'][$dashed_key_name];
+    }
+    else {
+      $token = $request[$dashed_key_name];
+    }
 
     // Check if there is a token that did not expire yet.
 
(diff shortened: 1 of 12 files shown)


e0ipso:
@ordavidil Thanks for contributing!

I'm not sure I understand the need of this. Wouldn't it be enough to create a different authentication provider that inherits from the access token and uses a different key?

ordavidil:
For me (and i'm sure it happens for other people as well), I always confuse between `access_token` and `access-token` and sometimes it takes too much time until I realize the problem is this mistype. It seems there is no security issue to let the user the option of which form to use in the app.

amitaibu:
@mateu-aguilo-bosch right now if you pass `access_token` or access_token`in the header it works fine. But if you pass in the $_GET, only`access_token` (which is the param name) will work.

@ordavidil the handling of the param should be moved to a centrealized place, so any plugin that has a param with underscore may use it.

e0ipso:
@amitaibu thank you for the clarification.

amitaibu:
@ordavidil is this ready?

ordavidil:
@amitaibu 
If the docs works for, then it is ready :-)

restful.module
```diff
@@ -820,3 +820,33 @@ function _restful_get_query_string_from_php_input() {
 
     return $query_string;
 }
+
+/**
+ * Extracting the token from a request by a key name,
```
amitaibu: Should be one liner

restful.module
```diff
@@ -820,3 +820,33 @@ function _restful_get_query_string_from_php_input() {
...
+ * either dashed or not.
+ *
+ * @param $options
```
amitaibu: `$param array ..`

restful.module
```diff
@@ -820,3 +820,33 @@ function _restful_get_query_string_from_php_input() {
...
+ *  The handler 'options' object.
+ * @param $request
```
amitaibu: `$param array ..`

restful.module
amitaibu: Also you can add below the oneliner more explanation, with a short example

restful.module
```diff
@@ -820,3 +820,32 @@ function _restful_get_query_string_from_php_input() {
...
+  if (!empty($request['__application'][$key_name])) {
+    return $request['__application'][$key_name];
+  }
+  elseif (!empty($request[$key_name])) {
+    return $request[$key_name];
+  }
+  elseif (!empty($request['__application'][$dashed_key_name])) {
+    return $request['__application'][$dashed_key_name];
+  }
+  else {
```
amitaibu: `else` not need here.

ordavidil:
@amitaibu 
You can  recheck.

modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php
```diff
@@ -12,9 +12,11 @@ class RestfulAuthenticationToken extends \RestfulAuthenticationBase {
   public function applies(array $request = array(), $method = \RestfulInterface::GET) {
     $options = $this->getPluginKey('options');
     $key_name = !empty($options['param_name']) ? $options['param_name'] : 'access_token';
+    $dashed_key_name = str_replace('_', '-', $key_name);
 
-    // Access token may be on the request, or in the headers.
-    return !empty($request['__application'][$key_name]) || !empty($request[$key_name]);
+    // Access token may be on the request, or in the headers
+    // (may be a with dash instead of underscore).
+    return !empty($request['__application'][$key_name]) || !empty($request[$key_name]) || !empty($request['__application'][$dashed_key_name]) || !empty($request[$dashed_key_name]);
```
amitaibu: Maybe better adapt `restful_extract_token_from_request()` to return FALSE if nothing found, to avoid this `!empty()` checks.

restful.module
```diff
@@ -820,3 +820,31 @@ function _restful_get_query_string_from_php_input() {
...
+ *
+ * @param $options
+ *  The handler 'options' object.
+ * @param $request
+ *  The current request.
+ *
+ * @return string
+ *  The token from the request.
+ */
+function restful_extract_token_from_request($options, $request) {
```
amitaibu: instead of `$options` pass a string with the param name

modules/restful_token_auth/restful_token_auth.module
```diff
@@ -131,3 +131,31 @@ function restful_token_auth_cron() {
...
+ *
+ * @param $param_name
+ *  The param name to check.
+ * @param $request
+ *  The current request.
+ *
+ * @return string
+ *  The token from the request.
+ */
+function restful_extract_token_from_request($param_name, $request) {
```
ordavidil: move as helper method

ordavidil: `extractTokenFromRequest`

amitaibu:
Looks good to me. @e0ipso - you ok with it?

e0ipso:
This looks good to me.

FYI, in Drupal 8 the token auth is going to be externalized to the new: https://www.drupal.org/project/simple_oauth

amitaibu:
Merged, thanks.

amitaibu:
@e0ipso dude, when did you have the time for simple_oauth?! :)

PR #585, opened 2015-08-06
Add mechanism to delete an access token.

lukewertz:
You may want to explicitly delete an access token in order to enforce a "logout" concept in an application that is consuming data from your API. This PR adds the ability for a user to delete their own access token.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,7 +245,7 @@ documentation:
       "required": false
     },
     "form_element": {
-      "type": "texfield",
+      "type": "textfield",
       "default_value": "",
       "placeholder": "",
       "size": 255,
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -1236,7 +1236,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
           ),
           // Information about the form element.
           'form_element' => array(
-            'type' => 'texfield',
+            'type' => 'textfield',
             'size' => 255,
           ),
         ),
--- a/tests/RestfulDiscoveryTestCase.test
+++ b/tests/RestfulDiscoveryTestCase.test
@@ -87,7 +87,7 @@ class RestfulDiscoveryTestCase extends RestfulCurlBaseTestCase {
         ),
         'form_element' => array (
           'allowed_values' => NULL,
-          'type' => 'texfield',
+          'type' => 'textfield',
           'default_value' => '',
           'placeholder' => '',
           'size' => 255,


e0ipso:
@lukewertz thanks for the PR! Sorry it took so long to respond, I am on vacation.

I agree with this, but I feel that we should remove the token using the entity controller.

See: https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/modules/restful_token_auth/includes/RestfulTokenAuthController.php#L41-L79

Maybe you could create a `deleteAuthToken($uid)` method with [this contents](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/modules/restful_token_auth/includes/RestfulTokenAuthController.php#L54-L64) and then refactor `generateRefreshToken` to also leverage this new function.

How does that sound?

Thanks again for contributing!

lukewertz:
@mateu-aguilo-bosch Thank you so much for the feedback & direction. (And thank you for replying while on vacation! That's seriously above-and-beyond.)

I'll refactor and update my PR. I hope to be able to do so this week.

Thank you very much!

e0ipso:
:blush:

lukewertz:
Back to you @mateu-aguilo-bosch for review.

Feedback is welcome. I'm happy to keep iterating, but I think this is closer to what you were wanting.

modules/restful_token_auth/plugins/restful/restful_token_auth/token_auth/access_token/1.0/access_token__1_0.inc
```diff
@@ -11,6 +11,7 @@ $plugin = array(
   'authentication_types' => array(
     'cookie',
     'basic_auth',
+    'token',
```
lukewertz: Adding token auth here to you can delete you access token while authenticated with your access token.

e0ipso:
This looks great!

Can I ask for testing? :blush:

lukewertz:
But of course.

lukewertz:
Do you think that [this test](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/modules/restful_token_auth/tests/RestfulTokenAuthenticationTestCase.test#L126-L135) should be removed/refactored?

e0ipso:
I mean a dedicated test for the DELETE request.

That one checks that the auth token is deleted after trying to authenticate with an expired token.

lukewertz:
Test added. I haven't dug into your tests to get a good sense of your testing strategy, so please feel free to provide (lots of) critical feedback. Thank you for your continued feedback in getting this written. I appreciate it.

e0ipso:
We also need to check that there is no token entity for that user after processing the DELETE request

lukewertz:
@mateu-aguilo-bosch Have you gotten a chance to take a look at my updated PR? Any feedback?

e0ipso:
This is looking good! Thanks @lukewertz!

e0ipso:
This is the CHANGELOG I want to write, but I'm not sure that we have permissions in place to control who can delete tokens. We should add those permissions (same kind of permissions on any content type) and we should enforce them in the resource.

Finally we'll need to make sure that unauthorized users get a 403 error if they are trying to delete a token they should not.

```
[FEATURE] (Token Auth) Allow user to delete token

Allow users to delete tokens so they can log-out from token
authentication. Deleting the auth token for a user should only be
allowed if the appropriate permissions allow.
```

e0ipso:
hey @lukewertz I'm sorry for that last minute request, I hope that didn't discourage you from writing the last bit for this PR.

In any case, your work is appreciated!

PR #588, opened 2015-08-17
[FEATURE] (Entity) Allow a different ID field

e0ipso:
This pull request allows you to specify a field to be used as the
primary ID. That means that all of the entity based resources need
to have the `id` field present. The `id` field is inherited from
the abstract base class ResourceEntity, but it can be swapped by
another one. If you don't want the `id` field present in your
resource input & output you can use the `'methods' => array()`
trick in the field definition. If you specify an `idField` in the
resource annotation, that field will be
used as the loadByIdField by default. This allows you to stop using
entity ids as your API ids and use UUIDs (maybe generated by the
uuid Drupal module).

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -10,7 +10,6 @@ namespace Drupal\restful\Plugin\resource\DataProvider;
 use Drupal\restful\Exception\ForbiddenException;
 use Drupal\restful\Exception\InternalServerErrorException;
 use Drupal\restful\Exception\ServerConfigurationException;
-use Drupal\restful\Http\HttpHeader;
 use Drupal\restful\Http\RequestInterface;
 use Drupal\restful\Plugin\resource\DataInterpreter\DataInterpreterEMW;
 use Drupal\restful\Plugin\resource\DataInterpreter\DataInterpreterInterface;
@@ -23,7 +22,6 @@ use Drupal\restful\Exception\UnprocessableEntityException;
 use Drupal\restful\Plugin\resource\Field\ResourceFieldInterface;
 use Drupal\restful\Plugin\resource\Resource;
 use Drupal\restful\Plugin\resource\ResourceInterface;
-use Drupal\restful\Resource\ResourceManager;
 
 class DataProviderEntity extends DataProvider implements DataProviderEntityInterface {
 
@@ -152,7 +150,20 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
       return array();
     }
 
-    return array_keys($result[$this->entityType]);
+    $entity_ids = array_keys($result[$this->entityType]);
+    if (empty($this->options['idField'])) {
+      return $entity_ids;
+    }
+
+    // Get the list of IDs.
+    $resource_field = $this->fieldDefinitions->get($this->options['idField']);
+    $ids = array();
+    foreach ($entity_ids as $entity_id) {
+      $interpreter = new DataInterpreterEMW($this->getAccount(), new \EntityDrupalWrapper($this->entityType, $entity_id));
+      $ids[] = $resource_field->value($interpreter);
+    }
+
+    return $ids;
   }
 
   /**
@@ -226,6 +237,8 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
     $wrapper->language($this->getLangCode());
     $interpreter = new DataInterpreterEMW($this->getAccount(), $wrapper);
     $resource_field_collection->setInterpreter($interpreter);
+    $id_field_name = empty($this->options['idField']) ? 'id' : $this->options['idField'];
+    $resource_field_collection->setIdField($this->fieldDefinitions->get($id_field_name));
 
     $input = $this->getRequest()->getParsedInput();
     $limit_fields = !empty($input['fields']) ? explode(',', $input['fields']) : array();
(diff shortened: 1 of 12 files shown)


amitaibu:
>  If you don't want the id field present in your
> resource input & output you can use the 'methods' => array()
> trick in the field definition

This feels like a bad hack. why are you forced to add it in the $public_fields? Currently for example you can just `unset ($public_fields['id'])` which feels more correct.

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -201,14 +203,34 @@ protected function propertyIdentifier(\EntityMetadataWrapper $property_wrapper)
...
       // ID.
-      $embedded_identifier = $property_wrapper->getIdentifier();
-    }
-    else {
-      // The property is a regular one, get the value out of it and use it as
-      // the embedded identifier.
-      $embedded_identifier = $this->fieldValue($property_wrapper);
+      $identifier = $property_wrapper->getIdentifier() ?: NULL;
+      $resource = $this->getResource();
+      // TODO: Make sure we still want to support full_view.
```
amitaibu, line 208: why not? :)

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -152,7 +150,20 @@ public function getIndexIds() {
...
+    $entity_ids = array_keys($result[$this->entityType]);
+    if (empty($this->options['idField'])) {
+      return $entity_ids;
+    }
+
+    // Get the list of IDs.
+    $resource_field = $this->fieldDefinitions->get($this->options['idField']);
+    $ids = array();
+    foreach ($entity_ids as $entity_id) {
+      $interpreter = new DataInterpreterEMW($this->getAccount(), new \EntityDrupalWrapper($this->entityType, $entity_id));
```
amitaibu, line 162: shouldn't `\EntityDrupalWrapper` be `\EntityMetadataWrapper`?

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -201,14 +203,34 @@ protected function propertyIdentifier(\EntityMetadataWrapper $property_wrapper)
...
+      /* @var ResourceInterface $resource */
+      $resource = $plugin_manager->createInstance($instance_id);
+      $plugin_definition = $resource->getPluginDefinition();
+      if (empty($plugin_definition['dataProvider']['idField'])) {
+        return $identifier;
+      }
+      try {
+        return $property_wrapper->{$plugin_definition['dataProvider']['idField']}->value();
+      }
+      catch (\EntityMetadataWrapperException $e) {
```
amitaibu, line 227: when do we get a `catch`?

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso, line 162: `entity_metadata_wrapper` yields `EntityDrupalWrapper` for entities. `\EntityMetadataWrapper` is not good enough in this case.

src/Plugin/resource/ResourceEntity.php
```diff
@@ -183,9 +186,21 @@ protected function processPublicFields(array $field_definitions) {
     // The fields that only contain a property need to be set to be
     // ResourceFieldEntity. Otherwise they will be considered regular
     // ResourceField.
-    return array_map(function ($field_definition) {
+    $field_definitions = array_map(function ($field_definition) {
       return $field_definition + array('class' => '\Drupal\restful\Plugin\resource\Field\ResourceFieldEntity');
     }, $field_definitions);
+
+    // If there is an alternate id field, use it instead of the entity id.
+    $plugin_definition = $this->getPluginDefinition();
```
amitaibu, line 194: i'd be happy to remove this part, and keep all declarations in publicFields() without auto-removing the ID. maybe one would like to expose the uuid, but keep the id in place.

src/Plugin/resource/DataProvider/DataProviderEntity.php
amitaibu, line 162: sorry, I meant why not:

``` php
$wrapper = entity_metadata_wrapper($this->entityType, $entity_id);
$interpreter = new DataInterpreterEMW($this->getAccount(), $wrapper);
```

e0ipso:
> This feels like a bad hack.

Hrmm… to me not doing it feels wrong. From the resource configuration point of view there should be no notion of Drupal specific fields. All filtering, sparse fieldsets, sorts, load by id field, … only treat with public fields. I don't see why this should be an exception, this is just a _glorified default loadByIdField_.

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso, line 162: I'm good with that :-)

amitaibu:
> Hrmm… to me not doing it feels wrong

I'm ok with that, my comment probably is about the "magic" happening here https://github.com/RESTful-Drupal/restful/pull/588/files#r37557288

amitaibu:
Apart of the above comment, rest looks good.

amitaibu:
btw, any tests to it somewhere?

e0ipso:
> btw, any tests to it somewhere?

Those will need to wait a bit :-S

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 208: You are right, it may not make sense for JSON API but it does for HAL and SimpleJson

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 227: If that property does not exist in the specified entity (or is not declared to Property API)

src/Plugin/resource/ResourceEntity.php
e0ipso, line 194: I'll create a follow up for that. Is that cool?

e0ipso, line 194: https://github.com/RESTful-Drupal/restful/issues/603

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 208: https://github.com/RESTful-Drupal/restful/issues/604

e0ipso:
Tests TODO in https://github.com/RESTful-Drupal/restful/issues/602

amitaibu:
You can merge this one, and maybe deal/ discuss my comment in a follow up, as it's the base of other PRs which are to hard to check for now.

PR #592, opened 2015-08-18
[BUGFIX] (JSON API) Fix nested includes

e0ipso:
Nested includes are now only included if requested. A included
compound document will be requested with dot notation, and RESTful
will ensure that full document linkage is provided.

PR #593, opened 2015-08-18
[FEATURE] Allow files to be references

e0ipso:
Allow files to be treated as entity references. Instead of getting
the file array, get the entity id or the full entity. To use this
behaviour use the 'class' property in the publicFields() definition.

### DIFF
--- a/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -274,14 +274,19 @@ class ResourceFieldEntity implements ResourceFieldEntityInterface {
       if (isset($resource['full_view']) && $resource['full_view'] === FALSE) {
         return $embedded_identifier;
       }
-      $request = Request::create('', array(), RequestInterface::METHOD_GET);
+      // We support dot notation for the sparse fieldsets. That means that
+      // clients can specify the fields to show based on the "fields" query
+      // string parameter.
+      $parsed_input = array(
+        'fields' => implode(',', $this->emdeddedSparseFieldsets()),
+      );
+      $request = Request::create('', $parsed_input, RequestInterface::METHOD_GET);
       // Remove the $_GET options for the sub-request.
-      $request->setParsedInput(array());
       // TODO: Get version automatically to avoid setting it in the plugin definition. Ideally we would fill this when processing the plugin definition defaults.
       $resource_data_provider = DataProviderResource::init($request, $resource['name'], array(
         $resource['majorVersion'],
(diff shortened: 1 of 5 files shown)


e0ipso:
This PR depends on #592. Please, merge that one first.

amitaibu:
@mateu-aguilo-bosch can you please update the different PRs with latest code - for easier review.

e0ipso:
@amitaibu done.

src/Plugin/resource/Field/ResourceFieldFileEntityReference.php
```diff
@@ -0,0 +1,41 @@
...
+   *
+   * @return \EntityMetadataWrapper
+   *   Either a \EntityStructureWrapper or a \EntityListWrapper.
+   *
+   * @throws ServerConfigurationException
+   */
+  protected function propertyWrapper(DataInterpreterInterface $interpreter) {
+    $property_wrapper = parent::propertyWrapper($interpreter);
+    // If the file_entity module is not installed, throw an exception.
+    if (!module_exists('file_entity')) {
```
amitaibu, line 30: why? You can use the wrapper on a file, without file_entity. AFAIK file_entity just makes that entity fieldable.

src/Plugin/resource/Field/ResourceFieldEntity.php
```diff
@@ -440,6 +445,27 @@ protected function formatterValue(\EntityMetadataWrapper $property_wrapper, \Ent
   }
 
   /**
+   * Gets the list of fields to show in the embedded resource.
+   *
+   * @return string[]
+   *   The list of fields.
+   */
+  protected function emdeddedSparseFieldsets() {
+    // TODO: Inject the request object here.
```
amitaibu, line 454: isn't it just doing `emdeddedSparseFieldsets(RequestInterface ..)`?

amitaibu:
minor comments, other comments done on another mega PR

src/Plugin/resource/Field/ResourceFieldFileEntityReference.php
e0ipso, line 30: Good point. I don't remember exactly why I put this (maybe entity access related?). Probably safe to remove.

src/Plugin/resource/Field/ResourceFieldEntity.php
e0ipso, line 454: The problem is that we don't have the request object in resource fields. Good news is that I (think that I) added support for that in a subsequent PR.

e0ipso:
Merged.

PR #595, opened 2015-08-19
[BUGFIX] Fix includes and fields for nested

e0ipso:
Pass the request down to the resource fields so we can pass relevant
query string parameters to compound documents. The parameters are
detected by stripping the public name + a dot from the beginning
of the parameter. So if 'related.images.uri' is passed to the top
request, 'images.uri' will be passed to the ResourceFieldResource
for 'related'. The resource field for 'images' will receive 'uri'
in the generated request object to compound the document.


e0ipso:
This PR depends on #593. Please, merge that one first.

e0ipso:
Tests TODO in https://github.com/RESTful-Drupal/restful/issues/605

e0ipso:
Merged.

PR #596, opened 2015-08-19
[BUGFIX] Make getImageUri static function

e0ipso:
So it can be referenced as process callbacks.


e0ipso:
This PR depends on #595. Please, merge that one first.

e0ipso:
Merged.

PR #597, opened 2015-08-19
[FEATURE] Adds base decorators for DataProviders

e0ipso:
When a resource wants to modify something of the data provider, they
can bind to a data provider decorator. That custom decorator can
have public methods overridden.


e0ipso:
This PR depends on #596. Please, merge that one first.

e0ipso:
Merged.

PR #599, opened 2015-08-21
[BUGFIX] Do not trigger access tag on sort

e0ipso:
When there is a sort on a field (and therefore a join to the field
table) we cannot add the access tag. Just like we are skipping it
when there is a field condition.


e0ipso:
This PR depends on #597.

e0ipso:
Merged.

PR #600, opened 2015-08-21
[FEATURE] Filter by the referencedIdProperty value

### LINKED ISSUE #598: 2.x Be smarter with filters and referencedIdProperty.
When outputting an entity reference like field, we can choose a property/field to use instead of the entity ID. We do that by using `referencedIdProperty`.

This is very handy for taxonomy terms, where you usually want the term name.

The problem is that when filtering the system does not do the reverse translation, so if you pass the term name instead of the tid you get 0 results.

This needs to be fixed.

e0ipso:
When outputting reference values you can use referencedPropertyId to
return a different field than the referenced entity's ID. With this
PR filters accept the same value that was returned. That allows you
for instance to filter based on taxonomy term name.

Fixes #598

### DIFF
--- a/modules/restful_example/src/Plugin/resource/node/article/v2/Articles__2_1.php
+++ b/modules/restful_example/src/Plugin/resource/node/article/v2/Articles__2_1.php
@@ -2,7 +2,7 @@
 
 /**
  * @file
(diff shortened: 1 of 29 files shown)


e0ipso:
This PR depends on #599.

e0ipso:
Merged.

PR #606, opened 2015-08-22
Add docs about accept formatter in Header.

theodorosploumis:
I think this is important to have it on docs since a lot of developers will look for a url formatter (see related issue https://github.com/RESTful-Drupal/restful/issues/549).


e0ipso:
Merged.

Thanks for contributing @theodorosploumis!

PR #607, opened 2015-08-24
[BUGFIX] Wrong query for propertyCondition in RestfulDataProviderEFQ.

ordavidil:
When `$property_name` is a _field_ and the operator is either `IN`, `BETWEEN` we pass the array values as is to the `fieldCondition` function.

Same login should happen when `$property_name` is a _property_, otherwise we get:

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax;
```

This PR is fixing that bug.


e0ipso:
Great! I fixed this same bug last week for 2.x

Can I ask for a test?

ordavidil:
You got it....

e0ipso:
This is great! Many thanks @ordavidil! I'll be merging this one soon.

e0ipso:
Merged.

PR #602, opened 2015-08-24
[FEATURE] Add tests to alternative reference id

e0ipso:
Suggestions:
- [x] Use the `uuid` module as the alternate ID.

Entity ID (idField):
- [x] Test that the self link and self ID are UUID.
- [x] Test that entities are loaded based on UUID.
- [x] Test that lists can be filtered by UUID.
- [x] Test that the compound resource ID is UUID in JSON API.

Reference fields. Include file, entityreference and taxonomy reference. (referencedPropertyId + idField)
- [x] Test that fields without a resource return the UUID when no resource is specified.
- [x] Test that fields with a resource connect with the embedded entity through the UUID if there is a matching idField.
- [x] Test that fields with a resource do not connect with the embedded entity through the UUID if there is not a matching idField.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ script:
   - drush --yes dl plug
   - drush --yes dl entity
   - drush --yes dl entityreference
+  - drush --yes dl uuid
   - drush --yes pm-enable plug
 
   # Patch Entity API.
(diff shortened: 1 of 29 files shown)


e0ipso:

e0ipso:
This PR depends on #602

e0ipso:
Merged.

PR #611, opened 2015-08-26
Update \RestfulEntityBase::checkEntityAccess() comment.

heylookalive:
Just overriden this method and noticed the comment was incorrect and missed out on the 'view' operation in the comment.


e0ipso:
Thanks for the contribution.

Merged!

PR #619, opened 2015-08-30
Check to see if paging links are present when data is passed to the f…

heylookalive:
…ormatter and add in if missing.

### DIFF
--- a/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
+++ b/plugins/formatter/hal_json/RestfulFormatterHalJson.class.php
@@ -85,14 +85,16 @@ class RestfulFormatterHalJson extends \RestfulFormatterBase implements \RestfulF
     }
 
     // Get self link.
-    $data['_links']['self'] = array(
-      'title' => 'Self',
-      'href' => $this->handler->versionedUrl($this->handler->getPath()),
-    );
+    if (!isset($data['_links']['self'])) {
+      $data['_links']['self'] = array(
+        'title' => 'Self',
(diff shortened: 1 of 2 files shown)


e0ipso:
What's the scenario where this is already present?

heylookalive:
Relating to cursor paging issue - https://github.com/RESTful-Drupal/restful/issues/594

With this change it means that I can implement cursor paging via a RESTful resource plugin without the need for creating new formatters.

The plugin can pass in the paging links via <code>additionalHateoas()</code>.

heylookalive:
Nudge

e0ipso:
I need to think about this a bit more. I don't like the idea of formatter logic leaking from the formatter into the resource (especially since we are already leaking from the resource to the formatter).

@amitaibu any strong feeling that may tip my undecided balance?

amitaibu:
I'll need to give it a deeper look, as I haven't done cursor paging yet myself.

heylookalive:
The approach I've tried to make is to have cursor paging be a setting which is agnostic of the formatter you're using, so you could have cursor paging on JSON or HAL JSON. It's possible you could have paging be it's own RESTful plugin type, but I can't think of any types other than offset or cursor paging.

As the changes are mostly within around the Provider layer, ComstackRestfulEntityBase overrides the addCidParams and queryListForPagination methods:
https://github.com/lohi-io/comstack_pm/blob/7.x-1.x/modules/comstack_pm_rest_api/plugins/restful/ComstackRestfulEntityBase.php#L28

I'm sure there'll be things that could be improved, the proposed pull request will at least facilitate this.

PR #616, opened 2015-08-29
[FEATURE] Allow nested filters

e0ipso:
Support JSON API relational filters specification in
https://gist.github.com/mateu-aguilo-bosch/1acbced683908ff631e3

PR #620, opened 2015-08-31
[FEATURE] Add metadata collection to data provider

e0ipso:
Data providers can add metadata reflecting the data gathering/setting
process. This information is made available to Formatters via the
resource so extra metadata can be added to the output.


e0ipso:
This PR depends on #616.

PR #625, opened 2015-09-03
Fix basePath so it works when lang code is in URL

jeff-h:
The basePath prepared by `url('', array('absolute' => TRUE))` doesn't have a trailing / like it should. Without it, the example angular code breaks on a site with language codes in the url e.g. domain.com/en/api/v1.5/articles. At present the code tries to go to e.g. domain.com/enapi/v1.5/articles

Not sure why you aren't using Drupal's `base_path()` but assuming you have your reasons :)


e0ipso:
This has also been merged in 2be98f1ee922ae1eee02073b37cab5bc680c0001

You are rocking it!

PR #627, opened 2015-09-04
Fix path to statically-served files

jeff-h:
When Drupal has language-codes in the URL, RESTful_angular_example.module calculates the path to statically-served files incorrectly. This patch should fix the problem.


e0ipso:
This has been merged in 2be98f1ee922ae1eee02073b37cab5bc680c0001

Thanks for your contribution!

Issue #628, opened 2015-09-04
Disbled plugin causing PHP errors on ResourceManager::getPlugins()


amitaibu:
#618

``` php
  $plugins = restful()
    ->getResourceManager()
    ->getPlugins();


  foreach ($plugins as $key => $plugin) {
    if (empty($plugin)) { / <-- plugin shouldn't be empty
      dpm($key);
    }
  }
```

amitaibu:
Update: Problem seems to be that `\Drupal\restful\Resource\ResourcePluginCollection::get` returns NULL when resource is empty. 

@e0ipso So I think we should change the signature of `\Drupal\restful\Resource\ResourceManager::getPlugins` to be `($only_enabled = TRUE)` by default, to prevent checking `!empty()` every time.

PR #629, opened 2015-09-04
ResourceManager::getPlugins should return enabled plugins by default

amitaibu:
#628

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderPlug.php
+++ b/src/Plugin/resource/DataProvider/DataProviderPlug.php
@@ -129,10 +129,10 @@ class DataProviderPlug extends DataProvider implements DataProviderInterface {
    * {@inheritdoc}
    */
   public function getIndexIds() {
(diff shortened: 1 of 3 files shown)


src/Plugin/resource/DataProvider/DataProviderPlug.php
```diff
@@ -129,10 +129,10 @@ public function remove($identifier) {
    * {@inheritdoc}
    */
   public function getIndexIds() {
-    // Return all of the instance IDs, even for disabled resources.
+    // Return all of the instance IDs.
     $plugins = restful()
       ->getResourceManager()
-      ->getPlugins(TRUE);
+      ->getPlugins();
```
amitaibu, line 135: @e0ipso I'm changing it here as in `applyFilters` we will get an error for a disabled plugin.

amitaibu:
ready for review.

oh, Now your @e0ipso also in Github! :)

Issue #633, opened 2015-09-08
Return appropriate headers for an OPTIONS call


e0ipso:
See #632.

PR #632, opened 2015-09-08
[FEATURE] Discover endpoints through OPTIONS calls

e0ipso:
Respond with information about the data using the OPTIONS calls. The
information is managed at the ResourceField level with the
FieldPublicInfo objects.

PR #637, opened 2015-09-09
[FEATURE] Provide access control headers in OPTIONS calls

e0ipso:
Add the Access-Control-Allow-\* headers to provide CORS support via
preflight OPTIONS requests. The origin domain can be specified
globally for all the API or individually for every endpoint.

### DIFF
--- a/restful.admin.inc
+++ b/restful.admin.inc
@@ -59,6 +59,20 @@ function restful_admin_settings($form_state) {
     ),
   );
 
+  $description  = 'When you make an XHR (AJAX) call to a resource under a ';
+  $description .= 'different host, most modern browsers will make an initial ';
+  $description .= 'OPTIONS request to the resource. The response to that can ';
+  $description .= 'contain the Access-Control-Allow-Origin, if that includes ';
+  $description .= 'the domain making the request the browser will allow the ';
+  $description .= 'cross-domain request. The contents of this field will be ';
+  $description .= 'used in that header.';
+  $form['restful_allowed_origin'] = array(
+    '#type' => 'textfield',
+    '#title' => t('CORS Preflight'),
+    '#description' => t($description),
+    '#default_value' => variable_get('restful_allowed_origin', ''),
+  );
+
   $form['restful_cache'] = array(
     '#type' => 'fieldset',
     '#title' => t('Cache'),
--- a/restful.module
+++ b/restful.module
@@ -339,7 +339,14 @@ function restful_delivery($var = NULL, $method = 'format') {
       ->getFormatterManager();
     $formatter_manager->setResource($resource);
     $plugin_definition = $resource->getPluginDefinition();
-    $formatter_name = isset($plugin_definition['formatter']) ? $plugin_definition['formatter'] : NULL;
+    if ($request->getMethod() == RequestInterface::METHOD_OPTIONS) {
+      // There is no guarantee that other formatters can process the
+      // auto-discovery output correctly.
+      $formatter_name = 'json';
+    }
+    else {
+      $formatter_name = isset($plugin_definition['formatter']) ? $plugin_definition['formatter'] : NULL;
+    }
     $output = call_user_func(array($formatter_manager, $method), $var, $formatter_name);
     $response->setContent($output);
   }
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -78,32 +78,19 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
       throw new InternalServerErrorException('The entity type was not provided.');
     }
     $this->entityType = $options['entityType'];
-    if (isset($options['bundles'])) {
+    if (!empty($options['bundles'])) {
       $this->bundles = $options['bundles'];
+    }
+    else {
+      // If no bundles are passed, then assume all the bundles of the entity
+      // type.
+      $entity_info = entity_get_info($this->entityType);
+      $this->bundles = array_keys($entity_info['bundles']);
     }
     if (isset($options['EFQClass'])) {
       $this->EFQClass = $options['EFQClass'];
     }
 
-    foreach ($this->fieldDefinitions as $key => $value) {
-      // Set the entity type and bundles on the resource fields.
-      if (!($value instanceof ResourceFieldEntityInterface)) {
-        continue;
-      }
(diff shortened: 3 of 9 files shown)


e0ipso:
Fixes #633.

e0ipso:
~~This PR depends on #632.~~

restful.admin.inc
```diff
@@ -59,6 +59,20 @@ function restful_admin_settings($form_state) {
...
+  $description  = 'When you make an XHR (AJAX) call to a resource under a ';
+  $description .= 'different host, most modern browsers will make an initial ';
+  $description .= 'OPTIONS request to the resource. The response to that can ';
+  $description .= 'contain the Access-Control-Allow-Origin, if that includes ';
+  $description .= 'the domain making the request the browser will allow the ';
+  $description .= 'cross-domain request. The contents of this field will be ';
+  $description .= 'used in that header.';
+  $form['restful_allowed_origin'] = array(
+    '#type' => 'textfield',
+    '#title' => t('CORS Preflight'),
```
amitaibu, line 71: you shouldn't pass variables to `t()`, only the text itself.

amitaibu, line 71: also the title is confusing. Lets call it `Allowed origin`. The cors preflight is a (geeky) implementation detail of the allowed origin.

restful.module
```diff
@@ -339,7 +339,14 @@ function restful_delivery($var = NULL, $method = 'format') {
       ->getFormatterManager();
     $formatter_manager->setResource($resource);
     $plugin_definition = $resource->getPluginDefinition();
-    $formatter_name = isset($plugin_definition['formatter']) ? $plugin_definition['formatter'] : NULL;
+    if ($request->getMethod() == RequestInterface::METHOD_OPTIONS) {
```
amitaibu, line 342: I'm not sure about this restriction - wouldn't you want to get the OPTIONS info in `json-api` format?

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -78,32 +78,19 @@ public function __construct(RequestInterface $request, ResourceFieldCollectionIn
...
     $this->entityType = $options['entityType'];
-    if (isset($options['bundles'])) {
+    if (!empty($options['bundles'])) {
       $this->bundles = $options['bundles'];
     }
+    else {
+      // If no bundles are passed, then assume all the bundles of the entity
+      // type.
+      $entity_info = entity_get_info($this->entityType);
+      $this->bundles = array_keys($entity_info['bundles']);
```
amitaibu, line 88: I think this isn't always populated (e.g. for `user` entity where entity == bundle). 

`!empty($entity_info['bundles']) ? array_keys($entity_info['bundles']) : $entity_info['type']`

src/Plugin/resource/Resource.php
```diff
@@ -567,4 +568,48 @@ protected function initAuthenticationManager() {
...
+   */
+  protected function preflight($path) {
+    $plugin_definition = $this->getPluginDefinition();
+    $allowed_origin = empty($plugin_definition['allowOrigin']) ? variable_get('restful_allowed_origin', NULL) : $plugin_definition['allow_origin'];
+    // Always add the allow origin if configured.
+    if ($allowed_origin) {
+      restful()
+        ->getResponse()
+        ->getHeaders()
+        ->add(HttpHeader::create('Access-Control-Allow-Origin', check_plain($allowed_origin)));
```
amitaibu, line 585: is `check_plain()` really needed?

src/Plugin/resource/Resource.php
```diff
@@ -567,4 +568,48 @@ protected function initAuthenticationManager() {
...
+    }
+
+    // Make sure the Access-Control-Allow-Methods is populated.
+    $allowed_methods = array();
+    foreach ($this->getControllers() as $pattern => $controllers) {
+      // Find the controllers for the provided path.
+      if ($pattern == $path || ($pattern && preg_match('/' . $pattern . '/', $path))) {
```
amitaibu, line 592: I'm a bit confused by this part. What is the pattern matching/ access checks?

restful.admin.inc
e0ipso, line 71: > The cors preflight is a (geeky) implementation detail of the allowed origin.

:+1:

> you shouldn't pass variables to t(), only the text itself.

I only put it in a variable to split it in different lines, to help with readability.

restful.module
e0ipso, line 342: Because formatters can do complex things based on other metada of the field. Like, the field normally is an Entity Reference, but when turned into an OPTIONS call it does not reference any entity. Moving that into relationships and trying to find an ID for the referenced entity doesn't make a lot of sense.

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso, line 88: :+1:

src/Plugin/resource/Resource.php
e0ipso, line 585: It's not harmful, I'm just trying to be careful about user input data. To your point, I don't have an attack vector in mind.

src/Plugin/resource/Resource.php
e0ipso, line 592: We can have different controllers per path. At this point in the code we are getting the appropriate controller for the given path.

Later on, we check if the controller has am access callback, if so we check the access.

restful.admin.inc
amitaibu, line 71: not allowed -- See [Translating Variables](https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/t/7) part

restful.module
amitaibu, line 342: ok

src/Plugin/resource/Resource.php
amitaibu, line 592: but wouldn't you expect the pattern match/access to happen pre-preflight? i.e. we shouldn't get to this point if there's no access.

e0ipso, line 592: Oh! We do have access to get here for the OPTIONS method. What this is trying to find out is what _other_ methods have access to this path, to populate `Access-Control-Allow-Methods` with the list of HTTP methods.

restful.admin.inc
e0ipso, line 71: works for me.

e0ipso:
Feedback was implemented and PR merged in d1b4c4fde0209f9d63ace13a851b08ac59779e79

Issue #638, opened 2015-09-09
Field Translation Support


ordavidil:
We should be able to `POST`, `PATCH` translated content to a translated field by it language.

e0ipso:
@ordavidil as you may have guessed, we are not doing any translated API at this moment. That means that the translation support is a bit brittle.

Please let me know how I can assist you in creating PRs to fix this.

Thanks for taking the time to report to the issue queue!

amitaibu:
@e0ipso yeah, @ordavidil already has a plan! ;)

PR #643, opened 2015-09-10
Field translate support

ordavidil:
#638


ordavidil:
@amitaibu @e0ipso 
Can you please make a quick review before I continue to the tests?

Plus, if you have any enlightenment for the case when `cardinality > 1` I'll be glad to get some help with that.

Thanks.

plugins/restful/RestfulEntityBase.php
```diff
@@ -665,6 +689,19 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
     $save = FALSE;
     $original_request = $request;
 
+    $entity_info = $this->getEntityInfo();
+    $default_language = language_default();
+
+    // Determine if the current bundle use 'title_field'.
+    $title_field = FALSE;
+    if (module_exists('title')) {
```
ordavidil: move to help method

`$title_field = $this->getTitleField();`

plugins/restful/RestfulEntityBase.php
```diff
@@ -324,16 +324,40 @@ public function viewEntity($id) {
         }
 
         if (empty($info['formatter'])) {
-          if ($sub_wrapper instanceof EntityListWrapper) {
-            // Multiple values.
-            foreach ($sub_wrapper as $item_wrapper) {
-              $value[] = $this->getValueFromProperty($wrapper, $item_wrapper, $info, $public_field_name);
+          $translatable = field_is_translatable($this->entityType, field_info_field($info['property']));
```
ordavidil: move to own method

plugins/restful/RestfulEntityBase.php
```diff
@@ -324,16 +324,40 @@ public function viewEntity($id) {
...
+          if ($translatable) {
+            $optional_languages = array_keys(language_list());
+
+            foreach ($optional_languages as $lang) {
+              $cloned_wrapper = clone $wrapper;
```
ordavidil: explain why

plugins/restful/RestfulEntityBase.php
```diff
@@ -324,16 +324,40 @@ public function viewEntity($id) {
...
+              }
+              else {
+                // Single value.
+                $value[$lang] = $this->getValueFromProperty($cloned_wrapper, $sub_wrapper, $info, $public_field_name);
+              }
             }
           }
           else {
-            // Single value.
-            $value = $this->getValueFromProperty($wrapper, $sub_wrapper, $info, $public_field_name);
```
ordavidil, line 335: Maybe we can always clone the wrapper - it will use a bit more memory, but code will be the same.

plugins/restful/RestfulEntityBase.php
```diff
@@ -690,7 +727,60 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
       }
 
       $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
-      $wrapper->{$property_name}->set($field_value);
+
+      $translatable = field_is_translatable($this->entityType, field_info_field($info['property']));
+      if ($translatable) {
+        // TODO: Add a logic when cardinality is higher than 1.
+        $single_value = count($field_value) == 1;
```
ordavidil: use `$field_info['cardinality']`

plugins/restful/RestfulEntityBase.php
```diff
@@ -690,7 +727,60 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
...
+
+        if (is_array($field_value)) {
+          $translatable_field_value = $field_value['value'];
+        }
+        else {
+          $translatable_field_value = $field_value;
+        }
+
+        // Translatable field values should passed as an object.
+        $values = json_decode($translatable_field_value);
```
ordavidil: `drupal_json_encode`

plugins/restful/RestfulEntityBase.php
```diff
@@ -690,7 +727,60 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
...
+        $optional_languages = array_keys(language_list());
```
ordavidil: `$enabled_languages`

plugins/restful/RestfulEntityBase.php
```diff
@@ -690,7 +727,60 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
...
+        foreach ($values as $current_language => $current_language_value) {
+
+          // Current language is not on option.
+          if (!in_array($current_language, $optional_languages)) {
+            throw new \RestfulBadRequestException(format_string('Language @language is not an option.', array('@language' => $current_language)));
```
ordavidil: is not enabled.

ordavidil: `The language`

plugins/restful/RestfulEntityBase.php
```diff
@@ -690,7 +727,60 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
       }
 
       $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
```
ordavidil: move into the foreach

plugins/restful/RestfulEntityBase.php
ordavidil, line 335: @amitaibu 
I found that we can give up the `clone`.

ordavidil:
@e0ipso 
As it seems, the checks which failed on travis are the **CSRF token** tests.
I couldn't reproduce this on my local machine, maybe you have some ideas ?
Thanks.

plugins/restful/RestfulEntityBase.php
```diff
@@ -324,15 +355,17 @@ public function viewEntity($id) {
         }
 
         if (empty($info['formatter'])) {
-          if ($sub_wrapper instanceof EntityListWrapper) {
-            // Multiple values.
-            foreach ($sub_wrapper as $item_wrapper) {
-              $value[] = $this->getValueFromProperty($wrapper, $item_wrapper, $info, $public_field_name);
+          if (field_is_translatable($this->entityType, field_info_field($info['property']))) {
```
e0ipso: Don't we need to check that 'property' is a Field API field before doing `field_info_field($info['property'])`.

plugins/restful/RestfulEntityBase.php
```diff
@@ -324,15 +355,17 @@ public function viewEntity($id) {
...
+            $enabled_languages = array_keys(language_list());
+
+            foreach ($enabled_languages as $language) {
```
e0ipso, line 379: If there is only a single language, we should probably stick to no language code nesting.

plugins/restful/RestfulEntityBase.php
```diff
@@ -324,15 +355,17 @@ public function viewEntity($id) {
...
+              $wrapper->language($language);
+              $sub_wrapper = $info['wrapper_method_on_entity'] ? $wrapper : $wrapper->{$property};
```
e0ipso, line 381: (Out of scope) TODO: Create a method to get the sub-wrapper.

I'm going to create a follow up issue.

e0ipso, line 381: https://github.com/RESTful-Drupal/restful/issues/654

plugins/restful/RestfulEntityBase.php
```diff
@@ -665,6 +735,15 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
     $save = FALSE;
     $original_request = $request;
 
+    $default_language = language_default();
+    $title_field = $this->getTitleField();
+
+    // Set the default language of a new instance if this bundle support
+    // field translation.
+    if ($this->getMethod() == \RestfulInterface::POST && entity_translation_node_supported_type($this->getBundle())) {
```
e0ipso: This helper class if for any kind of entity, will `entity_translation_node_supported_type` only work for nodes? Do we need a different helper function?

e0ipso: Do we want to re-use the `is write request` helper method (forgot the exact name in 1.x) instead of `$this->getMethod() == \RestfulInterface::POST`?

plugins/restful/RestfulEntityBase.php
```diff
@@ -689,8 +768,48 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
         continue;
       }
 
-      $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
-      $wrapper->{$property_name}->set($field_value);
+      if (field_is_translatable($this->entityType, field_info_field($info['property']))) {
```
e0ipso: Let's move this chunk to its own function to make the code more concise.

plugins/restful/RestfulEntityBase.php
```diff
@@ -689,8 +768,48 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
...
+        $values = $request[$public_field_name];
+
+        // Parse values in case it still not an array.
+        if (!is_array($values)) {
+          $values = drupal_json_decode($values);
```
e0ipso: Not sure what this is, can you explain?

e0ipso: Will we be in trouble if someone is using an XML output formatter?

plugins/restful/RestfulEntityBase.php
```diff
@@ -1256,7 +1375,16 @@ public function publicFieldsInfo() {
       return $public_fields;
     }
 
-    if (!empty($entity_info['entity keys']['label'])) {
+    $title_field = $this->getTitleField();
+
+    // Change the 'label' public field settings when a bundle use 'title'
+    // module.
+    if (!empty($title_field)) {
```
e0ipso: maybe:

``` php
if ($title_field = $this->getTitleField())
```

tests/RestfulViewEntityMultiLingualTestCase.test
```diff
@@ -147,8 +147,8 @@ class RestfulViewEntityMultiLingualTestCase extends RestfulCurlBaseTestCase {
     $response = $handler->get($id);
     $result = $response[0];
 
-    $this->assertEqual(trim(strip_tags($result['text_single'])), $text1, 'Entity view has correct result for "text_single" in language "' . $langcode . '".');
-    $this->assertEqual(trim(strip_tags($result['text_multiple'][0])), $text1, 'Entity view has correct result for the first value of "text_multiple" in language "' . $langcode . '".');
-    $this->assertEqual(trim(strip_tags($result['text_multiple'][1])), $text2, 'Entity view has correct result for the second value of "text_multiple" in language "' . $langcode . '".');
```
e0ipso: Please, do not remove the old test. That should still keep working, otherwise we may break someone's site when they upgrade their RESTful.

e0ipso:
@ordavidil added some feedback and some questions.

This is great work!! Thanks for contributing :-)

tests/RestfulViewEntityMultiLingualTestCase.test
ordavidil: @e0ipso
I understand, do you have any suggestions how to make this one to work?

Since the format has changed for translated fields, these tests will always fail because they don't relate to the `$langcode`.

e0ipso: > Since the format has changed for translated fields

We should still support not passing the `$langcode` –even for translated fields– and defaulting to the currently negotiated language.

amitaibu: > We should still support not passing the $langcode –even for translated fields

Correct.

However, the test should change as Or did, since the return value will be an array keyed by the langcode

plugins/restful/RestfulEntityBase.php
amitaibu: > Will we be in trouble if someone is using an XML output formatter?

Not really, we are inside `setPropertyValues`, so we actually listen only to the `simple json` format. 

As we discussed in Barcelona, supporting write in other methods will wait for after the 2.0 release

amitaibu: but I agree the drupal_json_decode() looks weird, and should probably be _not_ needed.

amitaibu:
``` php
    $request = array(
      'translated_full' => array(
        'en' => 'hello',
        'es' => 'hola',
      ),

      'translated_default' => 'hello'
    );
```

plugins/restful/RestfulEntityBase.php
ordavidil: By the documentation of `field_info_field`:

```
  * @return
 *   The field array, as returned by field_read_fields(), with an
 *   additional element 'bundles', whose value is an array of all the bundles
 *   this field belongs to keyed by entity type. NULL if the field was not
 *   found.
```

So I guess if this is not a Field API field it returns `NULL`,
hope i'm not talking nonsense...

amitaibu: @ordavidil following the suggestion from @e0ipso you can do:

``` php
$field_info = field_info_field($info['property'];

if ($field_info && field_is_translatable($this->entityType, $field_info)) {
```

plugins/restful/RestfulEntityBase.php
amitaibu, line 379: agreed

plugins/restful/RestfulEntityBase.php
ordavidil: > Do we want to re-use the `is write request` helper method (forgot the exact name in 1.x) instead of `$this->getMethod() == \RestfulInterface::POST`?

I think that we should set the instance default language only on creation.

ordavidil: > This helper class if for any kind of entity, will `entity_translation_node_supported_type` only work for nodes? Do we need a different helper function?

@e0ipso @amitaibu 
Is there any way to check translation support for other types?
Also, what if this is a custom entity ?

e0ipso:
Can I help you to move this forward @ordavidil?

ordavidil:
Hey @e0ipso,
sorry for postponing it for awhile...
1. I found that when saving an entity instance that its default language is `und` then the entity wrapper don't know how to handle languages, and more than that it doesn't give you any alert, hence I was thinking either to:
   1. Throw an exception with a message like "The current entity can't be translated until a default language will be set"
   2. Set the default language of the entity by using `language_default('language');`
2. I still haven't checked it on other entities than Node, I'll try to do it this week.

e0ipso:
Thanks for the recap @ordavidil. Is this issue exclusive of RESTful? Or it's more an Entity API issue?

ordavidil:
It's not exclusive of RESTful

e0ipso:
:+1:

ordavidil:
Hey @e0ipso,  I know, it took a while but basically I finished this :-)

The tests are failing because of `General error: 2006 MySQL server has gone away`, probably need to increase the _max_allowed_packets_, and on PHP 5.4 there are other tests which fails all the time.

Let me know how can we proceed from here,
Thanks.

ordavidil:
@e0ipso @amitaibu 
The test that I've added pass just fine:
selection_173

And I've also re-rolled the branch...

any suggestions ?
:-/

bricel:
@ordavidil - Can you re-roll this branch, I want to add it as a patch.

ordavidil:
Sure, I'll start working on it.

ordavidil:
@bricel,
re-rolled.

PR #649, opened 2015-09-11
[BUGFIX] Apply targetted nested filter to relationships

e0ipso:
Check the rendered value to see if the referenced entity passes a
targetted nested filter. Failing to do this leads to the relationships
to be included anyways for JSON API.

### DIFF
--- a/src/Plugin/formatter/FormatterJsonApi.php
+++ b/src/Plugin/formatter/FormatterJsonApi.php
@@ -126,6 +126,22 @@ class FormatterJsonApi extends Formatter implements FormatterInterface {
           $value = array($value);
           $ids = array($ids);
         }
+        // If some IDs were filtered out in the value while rendering due to the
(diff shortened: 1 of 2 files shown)


amitaibu:
Can you give an example (maybe even in the comments, because it's not very clear when ID can be missing)

e0ipso:
Imagine a _weird_ node that has a `field_cool_text: 'hello', 'hello', 'friend'` (3 values). That _weird_ node is referenced in an article.

And you do a filter like `articles?filter[weird.coolText][value]=hello&filter[weird.coolText][target]=weird` you would expect that you get the article as usual, but the _weird_ embedded only contains 2 field_cool_text in it.

This code fixes the JSON API formatter so it doesn't choke with the fact that we've nullified `'friend'`.

Now think that `field_cool_text` was an entity reference, so we get a NULL sometimes instead of a valid ID (to use with the embedded resource).

PR #651, opened 2015-09-16
[BUGFIX] Clean entity IDs

e0ipso:
Steps to reproduce the issue: (1) Load entity_test 4 so it gets
statically cached. (2) Load entity_test '4asdg'. You get the entity
4 instead of FALSE.

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -724,7 +724,10 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
   protected function isValidEntity($op, $entity_id) {
     $entity_type = $this->entityType;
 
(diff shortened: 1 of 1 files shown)


e0ipso:
See for the Entity API proper fix: https://www.drupal.org/node/2569333

amitaibu:
Ouch. Maybe worth adding it as patch dependency  -- sounds "safer" (and anyway this is a bit of an edge case, no?)

e0ipso:
> Ouch. Maybe worth adding it as patch dependency -- sounds "safer"

I'd rather fix it in our side, since I am not confident that I'm not breaking anything else in Entity API and the patch needs other people to work on.

> (and anyway this is a bit of an edge case, no?)

It is: only entities using the Entity API controller class are affected + there is some kind of UUID module active for that entity + you pass the UUID as an entity id to entity_load.

Issue #603, opened 2015-08-21
2.x Do not unset the original entity ID field


e0ipso:
If an alternate `idField` is specified, do not remove the existing `'id'`.

See https://github.com/RESTful-Drupal/restful/pull/588#discussion_r37557288

e0ipso:
PR https://github.com/RESTful-Drupal/restful/pull/655

PR #655, opened 2015-09-21
[BUGFIX] Do not unset the original 'id' field

e0ipso:
Do not remove the 'id' field even if there is an alternative id
field to load the entity by.


e0ipso:
This fixes #603.

PR #660, opened 2015-09-26
[FEATURE] Introduce the render cache with cache fragments

e0ipso:
Introduce the render cache at the prepare layer (right before encoding).
Besides, cache fragment entities are declared and created for smart
cache invalidation and fragmentation.


e0ipso:
This is ready to be reviewed.

I think this is a big leap forward, allowing resource results to be cached for a **LONG** TTL having smart cache invalidation based on cache fragments.

amitaibu:
wow, that's a big one - I'll need to wrap my head around it

e0ipso:
> wow, that's a big one - I'll need to wrap my head around it

want to screenshare for a quick rundown?

amitaibu:
Yeah, I'll go over the code, and then ping. 
btw, tests are failing

e0ipso:
I'm trying to find the time to get back to this. I have not forgotten about it.

e0ipso:
I'll be angry if tests don't come back green.

e0ipso:
:angry:

e0ipso:
Ha! @amitaibu tests are green.

amitaibu:
Just to update that I haven't forgot about this - I'm going to review this tomorrow.
(btw, tests are red)

e0ipso:
> (btw, tests are red)

nope

amitaibu:
sorry for the false alarm, I miss looked ;)

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
+  $query
+    ->entityCondition('entity_type', 'cache_fragment')
+    ->propertyCondition('type', 'entity_type')
+    ->propertyCondition('value', $type);
+  if (!$hashes = CacheFragmentController::lookUpHashes($query)) {
+    return array();
+  }
+
+  // Search for all the cache fragments with our entity id.
+  list($entity_id,,) = entity_extract_ids($type, $entity);
```
amitaibu: you can just  `list($entity_id)`

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
  */
 function restful_entity_delete($entity, $type) {
-  list($entity_id) = entity_extract_ids($type, $entity);
-  $cid = 'paet:' . $type . '::ei:' . $entity_id;
-  CacheDecoratedDataProvider::invalidateEntityCache($cid);
+  $resource_manager = restful()->getResourceManager();
+  foreach (_restful_entity_cache_hashes($entity, $type) as $hash) {
+    $handler = $resource_manager->getPlugin(CacheFragmentController::resourceIdFromHash($hash));
+    if (!$handler instanceof CacheDecoratedResourceInterface) {
+      return;
```
amitaibu, line 114: `// Resource isn't using the default cache decorator.`

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
+    }
+    if (!$handler->hasSimpleInvalidation()) {
+      return;
```
amitaibu, line 117: `// Resource takes care of its own invalidation.`

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
+    }
+    $handler
```
amitaibu, line 120: My biggest concern (which isn't huge), is how scalable it is. If we cache per user, and we have 10K users hitting the same resource -- it means we need to delete 10K on a single delete.

Contradicting my own concern - if we have so many caches, one would have to implement it's own invalidation logic.

agree?

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
 
 /**
  * Implements hook_user_update().
  */
 function restful_user_update(&$edit, $account, $category) {
-  // Due to the limitations for cid matching on clearing caches, we need to
-  // clear all the bin. We cannot do $cid = '%::uu' . $account->uid . '::pa';
-  CacheDecoratedDataProvider::invalidateEntityCache('*');
+  // Search for all the cache fragments with our entity id.
+  $query = new \EntityFieldQuery();
```
amitaibu, line 131: why do we need special case of user?

Also wouldn't it run twice - after `restful_entity_update`?

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
+      continue;
+    }
+    if (!$handler instanceof CacheDecoratedResourceInterface) {
+      return;
+    }
+    if (!$handler->hasSimpleInvalidation()) {
+      return;
+    }
+    // You can get away without the fragments for a clear.
+    $cache_object = new RenderCache(new ArrayCollection(), $hash, $handler->getCacheController());
```
amitaibu, line 154: I'm not clear on what we're doing here.

restful.entity.inc
```diff
@@ -29,42 +33,152 @@ function restful_entity_info() {
...
 
 /**
  * Implements hook_user_delete().
  */
 function restful_user_delete($account) {
-  // Due to the limitations for cid matching on clearing caches, we need to
-  // clear all the bin. We cannot do $cid = '%::uu' . $account->uid . '::pa';
-  CacheDecoratedDataProvider::invalidateEntityCache('*');
+  // Search for all the cache fragments with our entity id.
+  $query = new \EntityFieldQuery();
```
amitaibu, line 166: same questions as `restful_user_update()`

restful.info
```diff
@@ -34,7 +34,7 @@ files[] = tests/RestfulListTestCase.test
 ; files[] = tests/RestfulPassThroughTestCase.test
 files[] = tests/RestfulRateLimitTestCase.test
 files[] = tests/RestfulReferenceTestCase.test
-; files[] = tests/RestfulRenderCacheTestCase.test
+files[] = tests/RestfulRenderCacheTestCase.test
```
amitaibu, line 37: :sparkling_heart:

restful.install
```diff
@@ -64,6 +64,47 @@ function restful_schema() {
...
+    'description' => 'Cache fragment base table',
+    'fields' => array(
+      'tid' => array(
+        'description' => 'The cache fragment unique ID.',
+        'type' => 'serial',
+        'unsigned' => TRUE,
+        'not null' => TRUE,
+      ),
+      'type' => array(
+        'description' => 'The tag type. Example: entity_id.',
```
amitaibu: Maybe: `The type of the cache fragment, e.g "entity_id".`

restful.module
```diff
@@ -519,9 +520,11 @@ function restful_restful_resource_alter(ResourceInterface &$resource) {
 
   $plugin_definition = $resource->getPluginDefinition();
   // Check for the rate limit configuration.
-  if (!empty($plugin_definition['renderCache']['render'])) {
-    // NOTE: Disabled cache for now. It needs to be moved to the Render layer.
-    // $resource = new CacheDecoratedResource($resource);
+  if (
```
amitaibu, line 526: worth a a comment line.

src/Plugin/formatter/Formatter.php
```diff
@@ -81,4 +84,104 @@ protected static function isIterable($input) {
...
+   * @return bool
+   *   TRUE if the data can be cached.
+   */
+  protected function isCacheEnabled($data) {
+    // We are only caching field collections, but you could cache at different
+    // layers too.
+    if (!$data instanceof ResourceFieldCollectionInterface) {
+      return FALSE;
+    }
+    if (!$context = $data->getContext()) {
```
amitaibu, line 102: why is this needed?

amitaibu, line 102: sorry, I understand it.

src/Plugin/formatter/Formatter.php
```diff
@@ -81,4 +84,104 @@ protected static function isIterable($input) {
...
+   * Gets the cached computed value for the fields to be rendered.
+   *
+   * @param mixed $data
+   *   The data to be rendered.
+   * @param mixed $output
+   *   The rendered data to output.
+   */
+  protected function setCachedData($data, $output) {
+    if (!$render_cache = $this->createCacheController($data)) {
+      return;
```
amitaibu, line 134: just because getCachedData already does -- `return NULL`

src/Plugin/formatter/Formatter.php
```diff
@@ -81,4 +84,104 @@ protected static function isIterable($input) {
...
+   * Gets a cache controller based on the data to be rendered.
+   *
+   * @param mixed $data
+   *   The data to be rendered.
+   *
+   * @return RenderCacheInterface
+   *   The cache controller.
+   */
+  protected function createCacheController($data) {
+    $context = $data->getContext();
```
amitaibu, line 149: use `isCacheEnabled`?

Also, do we really need to do this extra check - wasn't it infact already tested in `isCacheEnabled`?

src/Plugin/formatter/Formatter.php
```diff
@@ -81,4 +84,104 @@ protected static function isIterable($input) {
...
+    if (!$cached_resource = $this->getResource()) {
+      return NULL;
+    }
+    return RenderCache::create($cache_fragments, $cached_resource->getCacheController());
+  }
+
+  /**
+   * Remove the unnecessary fields from the response.
+   *
+   * @param string[] $limit_fields
```
amitaibu, line 167: Is `string[]` a valid phpdocs?

src/Plugin/formatter/Formatter.php
```diff
@@ -81,4 +84,104 @@ protected static function isIterable($input) {
...
+   *   The list of fields.
+   * @param array $output
+   *   The prepared output.
+   *
+   * @return array
+   *   The filtered contents.
+   */
+  public function limitFields($limit_fields, array $output) {
```
amitaibu: `array $limit_fields`

src/Plugin/formatter/FormatterHalJson.php
```diff
@@ -186,6 +190,21 @@ protected function extractFieldValues($rows) {
       if (!$rows instanceof ResourceFieldCollectionInterface) {
         throw new InternalServerErrorException('Inconsistent output.');
       }
+
+      // This feels a bit awkward, but if the result is going to be cached, it
+      // pays off the extra effort of generating the whole resource entity. That
+      // way we can get a different field set with the previously cached entity.
+      // If the entity is not going to be cached, then avoid generating the
+      // field data altogether.
+      $limit_fields = $rows->getLimitFields();
```
amitaibu, line 199: why do even need this check - can't we remove this whole block - and simply wait for the caching [few](https://github.com/RESTful-Drupal/restful/pull/660/files#diff-bf10e1e10ec909f1a0ad72002bb6dd55R221) lines below?

src/Plugin/formatter/FormatterJson.php
```diff
@@ -98,13 +117,17 @@ protected function extractFieldValues($data) {
       }
       $output[$public_field_name] = $value;
     }
+    if ($this->isCacheEnabled($data)) {
```
amitaibu, line 120: this duplication is nasty - lets break it down, so formatter implementors don't have to deal with it.

src/Plugin/formatter/FormatterJsonApi.php
```diff
@@ -286,14 +231,80 @@ protected function getRequest() {
...
+      return $this->decompress($output, $included);
+    }
+    foreach ($output['__fields'] as $field_name => $field_contents) {
+      if (empty($field_contents['__embedded'])) {
+        $result['attributes'][$field_name] = $field_contents;
+      }
+      else {
+        // Handle single and multiple relationships.
+        $rel = array();
+        $single_item = $field_contents['__cardinality'] == 1;
```
amitaibu, line 282: Damn, I just got a vertigo :wink:

src/Plugin/formatter/FormatterJsonApi.php
```diff
@@ -309,6 +320,142 @@ protected function moveEmbedsToIncluded(array &$output, array $included) {
...
+   *   The prepared output.
+   *
+   * @return array
+   *   The filtered contents.
+   */
+  public function limitFields($limit_fields, array $output) {
+    if (!$limit_fields) {
+      return $output;
+    }
+    if (empty($output['__fields'])) {
```
amitaibu, line 341: can't we just pass `$output['__fields']` to the parent method, instead of duplicating this method?

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -231,19 +248,18 @@ public function view($identifier) {
     $id_field_name = empty($this->options['idField']) ? 'id' : $this->options['idField'];
     $resource_field_collection->setIdField($this->fieldDefinitions->get($id_field_name));
 
+    // Defer sparse fieldsets to the formatter. That way we can minimize cache
```
amitaibu, line 251: I don't understand what you mean here

src/RenderCache/Entity/CacheFragmentController.php
```diff
@@ -0,0 +1,143 @@
...
+   * Gets the hashes for an EFQ.
+   *
+   * @param \EntityFieldQuery $query
+   *   The EFQ.
+   *
+   * @return string[]
+   *   The hashes that meet the conditions.
+   */
+  public static function lookUpHashes(\EntityFieldQuery $query) {
+    $results = $query->execute();
```
amitaibu, line 88: Is there a real use case doing the execute() here, and not getting the results?

amitaibu:
It's official. You are crazy! :)

**Amazing** job

src/RenderCache/Entity/CacheFragmentController.php
e0ipso, line 88: This is (supposedly) only for developer experience. So they can tweak the query by overriding the controller. Not sure if that is ever gonna happen.

restful.entity.inc
e0ipso, line 120: **Agree**

Yeah. This is meant to be just for _simple invalidation_. More complex cases should be handled case by case. That's the reason behind being able to turn off _simple invalidation_.

restful.entity.inc
e0ipso, line 131: I asked myself the same question. I'm not sure why it was like this in `1.x`, but I just ported it in this case.

restful.entity.inc
e0ipso, line 154: Instead of searching for all the `restful_cache_fragment` entities that reference this cache and delete them + clearing cache, we create the `RenderCache` object (that contains the cache) and call the `->clear()` method on it.

src/Plugin/formatter/Formatter.php
e0ipso, line 149: > use isCacheEnabled?

We can't use it, since `isCacheEnabled` only applies to `ResourceFieldCollections`.

> Also, do we really need to do this extra check - wasn't it in fact already tested in isCacheEnabled?

Just making sure, in case we want to use this helper method in other contexts.

src/Plugin/formatter/Formatter.php
e0ipso, line 167: yes! Super useful as well for `PhpStorm` type hinting.

src/Plugin/formatter/FormatterHalJson.php
e0ipso, line 199: When caching is enabled and we get a cache MISS we want to generate output for the cache entry for the **whole** entity. That way we can use that cache record independently of the sparse fieldset.

On the other hand, if cache is not enabled we don't want to output for the whole entity, only the bits that we are going to need. For performance reasons.

That is why we bail just before `->render()`. `->render()` does the heavy lifting to get eh field info.

src/Plugin/formatter/FormatterJson.php
e0ipso, line 120: I tried (unsuccessfully) to find an elegant way to do this. The problem is that we don't have only one way to generate the output, since it will depend on runtime configuration.

e0ipso, line 120: Ideas?

src/Plugin/formatter/FormatterJsonApi.php
e0ipso, line 282: Yeah, there is a lot of _proprietary format_ business logic here. I did not know how to avoid this and still have cacheable entities (with all the includes that they may need in the cache record).

This denormalization + optional caching + re-normalization is (IMO) the ugliest part of RESTful. I can see that we can improve this in future sweeps, but I think it's a _good enough_ **first** iteration.

src/Plugin/formatter/FormatterJsonApi.php
```diff
@@ -286,14 +231,80 @@ protected function getRequest() {
...
    *
+   * @return array
+   *   The processed data.
+   *
    * @throws \Drupal\restful\Exception\InternalServerErrorException
    */
-  protected function moveEmbedsToIncluded(array &$output, array $included) {
+  protected function decompress(array $output, array &$included) {
+    static $depth = -1;
+    $depth++;
```
e0ipso, line 251: Not sure how I feel myself about this `$depth` variable. It's certainly useful for debugging purposes (I use it all the time for conditional breakpoints), but it doesn't drive any business logic.

src/Plugin/formatter/FormatterJsonApi.php
e0ipso, line 341: We'd need to do the `empty` check in the parent method in order to avoid PHP warnings.

src/Plugin/resource/DataProvider/DataProviderEntity.php
e0ipso, line 251: I tried to explain https://github.com/RESTful-Drupal/restful/pull/660/files#r41708622

Does it make sense that way?

e0ipso:
Thank you so much for the (in)sanity check.

I really think that we needed to improve performance, while making sure that we don't get cache poisoning and we have all the tools for smart invalidation. To me having _yet another_ TTL-based cache layer didn't make a lot of sense, since we already have Drupal's page cache, Varnish (and Akamai) doing that. I wanted the render cache entries to be able to be cached **forever** and never be cleared, except by the invalidation methods.

e0ipso:
## Next steps

### Making sure that we don't have infinite recursion https://github.com/RESTful-Drupal/restful/issues/676

When entity `A` references `B` and in turn `B` references `C, D and A`, we need to replace the embedded `A` by a placeholder that gets expanded in `->decompress`.

### Making sure we add cache fragments for embedded entities https://github.com/RESTful-Drupal/restful/issues/677

When entity `A` references `B` and in turn `B` references `C, D and A`. If the cache entry for `D` gets updated we need to clear the caches of `D`, `B` and `A`.

### Improve performance on uncached requests https://github.com/RESTful-Drupal/restful/issues/678

Right now we are fetching the embedded information for relationships without checking if the relationship was requested using the `include` query parameter.

restful.entity.inc
e0ipso: :+1:

restful.entity.inc
e0ipso, line 114: :+1: 

`// Resource isn't using a cache decorator.`

restful.entity.inc
e0ipso, line 117: :+1:

restful.install
e0ipso: :+1:

restful.module
e0ipso, line 526: :+1:

src/Plugin/formatter/Formatter.php
e0ipso, line 134: In general if you don't have a `@return` tag in your docblock, the recommendation is to return _void_ instead of _NULL_. Even if those are the same in PHP.

Returning different from `return;` in a function that does not return a value will make PhpStorm unhappy.

src/Plugin/formatter/Formatter.php
e0ipso: :+1:

e0ipso:
@amitaibu feedback has been applied.

restful.entity.inc
amitaibu, line 131: I think It's because of the granularity of the cache clear - https://github.com/RESTful-Drupal/restful/blob/f0f981a8552a8c8b429bce2991d06e551661adff/restful.entity.inc#L56-L58

restful.entity.inc
amitaibu, line 154: `$cache_object->clear();` is just a proxy to actually delete all the fragments, no? If so `// You can get away without the fragments for a clear.` isn't very clear

src/Plugin/formatter/Formatter.php
amitaibu, line 134: ok

src/Plugin/formatter/FormatterHalJson.php
amitaibu, line 199: :+1:

src/Plugin/formatter/FormatterJson.php
amitaibu, line 120: not very beautiful, but maybe have the main method broken to three parts? -- implementors will have to deal only with the middle part

src/Plugin/formatter/FormatterJsonApi.php
amitaibu, line 282: it's fine -- I just admitted my review of this part was poorly done ;)

src/Plugin/formatter/FormatterJsonApi.php
amitaibu, line 341: matter of taste - for me duplication of method is more extreme then `!empty()` - but I can go either way

src/Plugin/resource/DataProvider/DataProviderEntity.php
amitaibu, line 251: Sure, maybe just add a `@see` with a reference example?

src/RenderCache/Entity/CacheFragmentController.php
amitaibu, line 88: > Not sure if that is ever gonna happen.

Me neither, but it's ok to keep as is I guess.

amitaibu:
Minor comments + failing tests. After that it can be merged :)

restful.entity.inc
e0ipso, line 131: Then we don't need this anymore!

e0ipso:
Merging as per https://github.com/RESTful-Drupal/restful/pull/660#issuecomment-147382355

amitaibu:
Cool stuff!

PR #661, opened 2015-09-28
Fixed request and count calculation for HATEOAS ‘next’ pager

pcambra:
When listing nodes, the count and next pager is not calculated correctly when a filter is present, this PR fixes the request refresh and also takes care of unpublished content not being added to the final count.

Test is pending

### DIFF
--- a/tests/RestfulListTestCase.test
+++ b/tests/RestfulListTestCase.test
@@ -367,7 +367,7 @@ class RestfulListTestCase extends RestfulCurlBaseTestCase {
     $request['filter'] = array('integer_single' => '1');
     $result = $handler->get('', $request);
     $this->assertEqual($result[0]['id'], $nodes['abc'], 'Filter list by Single value field.');
-    
+
     // LIKE operator.
     $request['filter'] = array(
       'label' => array(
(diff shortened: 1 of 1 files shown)


e0ipso:
Thanks for that PR @pcambra!

It would be awesome if we had a test with unpublished nodes to cover that :-)

restful.module
```diff
@@ -388,6 +388,10 @@ function restful_get_restful_handler_for_path($path = NULL) {
   list($major_version, $minor_version) = \RestfulBase::getVersionFromRequest($path);
 
   $handlers[$path] = restful_get_restful_handler($resource, $major_version, $minor_version);
+  // Make sure we have a fresh request.
+  $request = restful_parse_request();
```
amitaibu: hi @pcambra :)
Why do we need this part?

pcambra: Because the request comes cached and without filters at the point of calculating the count of results, I'll try to demonstrate it in a test

pcambra:
Just split the count and unpublished in #664, working on a test for the filter request now

pcambra:
Not sure on how to proceed with this test. The issue is that the request for the count loses the filters when calling restful_get_restful_handler_for_path()

pcambra:
So here's a more clear view of what is happening, pushed a failing test.

When a resource is displayed within another, the count results seem inaccurate due to RestfulBase->getRequestForSubRequest stripping out filters among other things.

Any ideas on how to overcome this?

e0ipso:
@pcambra what we could do is not to strip the filters, but create a new `$request` array and set it to the sub-request.

Pseudocode:

``` php
$request = $parent_handler->getRequest();
$sub_request = $child_handler::cleanRequest($request);
$child_handler->setRequest($sub_request);
```

e0ipso:
FYI I re triggered the tests.

e0ipso:
I'm closing this one @pcambra. Feel free to request a re-open if you find the time/energy to fix the test.

pcambra:
I think we would still need to fix the subrequest issue properly, I'll give a try to another patch more in the lines of what we actually want :)

e0ipso:
I'm reopening the PR for bug report purposes then.

@amitaibu any recollection on how this worked in `7.x-1.x`? My mind is blurry with all the `2.x` changes.

PR #667, opened 2015-10-03
[BUGFIX] Add a more generic relationship ID method 

e0ipso:
Add a more generic reference ID method leveraging the metadata
in the referenced resource definition. To get the referenced ID when
there is a 'referencedIdProperty' we only need the entity type and
the bundle. Both of these are present in the destination resource
definition.

Additionally use a similar technique to join entity tables in our
custom EFQ, when the referenced entity ID is held in a property
(instead of a field).


e0ipso:
I know that the first diff is confusing but I just removed the special case handling, and refactored the code a little bit to make it more readable. Perhaps reading the resulting code is clearer than reading the diff.

e0ipso:
Merging and tagging for post-mortem review.

PR #672, opened 2015-10-06
Fixes #662, Records last-access time for users.

### LINKED ISSUE #662: Token authentication doesn't update Access or Login time
Drupal keeps track of when users last logged in or accessed the site as a logged in user. They are properties on the user entity.

When using token authentication, I think that both the access and login times should be updated, but they are currently not updated.

I think of this as a bug, but am unsure if this is by design for some reason.

lukewertz:
When a user's account is loaded, the user table should be updated to indicate that they have made an authenticated request. This mirrors the logic in _drupal_session_write().

### DIFF
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -123,6 +123,10 @@ class RestfulAuthenticationManager extends \ArrayObject {
     // This is necessary because the page cache system only looks at session
     // cookies, but not at HTTP Basic Auth headers.
     drupal_page_is_cacheable(!$account->uid && variable_get('restful_page_cache', FALSE));
+
+    // Record the access time of this request.
+    $this->setAccessTime($account);
+
     return $account;
   }
 
@@ -205,5 +209,20 @@ class RestfulAuthenticationManager extends \ArrayObject {
     return $this->originalUserSession;
   }
 
-
+  /**
+   * Set the user's last access time.
+   *
+   * @param object $user
+   *   A drupal user account.
+   *
+   * @see _drupal_session_write()
+   */
+  protected function setAccessTime($user) {
+    // This logic is pulled directly from _drupal_session_write().
+    if ($user->uid && REQUEST_TIME - $user->access > variable_get('session_write_interval', 180)) {
+      db_update('users')->fields(array(
+        'access' => REQUEST_TIME,
+      ))->condition('uid', $user->uid)->execute();
+    }
+  }
 }
--- a/plugins/authentication/RestfulAuthenticationManager.php
+++ b/plugins/authentication/RestfulAuthenticationManager.php
@@ -212,17 +212,17 @@ class RestfulAuthenticationManager extends \ArrayObject {
   /**
    * Set the user's last access time.
    *
-   * @param object $user
-   *   A drupal user account.
+   * @param object $account
+   *   A user account.
    *
    * @see _drupal_session_write()
    */
-  protected function setAccessTime($user) {
+  protected function setAccessTime($account) {
     // This logic is pulled directly from _drupal_session_write().
-    if ($user->uid && REQUEST_TIME - $user->access > variable_get('session_write_interval', 180)) {
+    if ($account->uid && REQUEST_TIME - $account->access > variable_get('session_write_interval', 180)) {
       db_update('users')->fields(array(
         'access' => REQUEST_TIME,
-      ))->condition('uid', $user->uid)->execute();
+      ))->condition('uid', $account->uid)->execute();
     }
   }
 }
--- a/tests/RestfulAuthenticationTestCase.test
+++ b/tests/RestfulAuthenticationTestCase.test
@@ -152,38 +152,29 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
 
     // Case 1. Ensure that access time is recorded for cookie auth.
     $user = $user1;
-    $cookie_provider = restful_get_authentication_handler('cookie');
 
-    // Make a request.
-    $handler->getAccount()->uid;
+    $user1_access_time_before = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user1->uid))->fetchObject();
+
+    // Perform request authentication.
+    $handler->getAccount();
 
     $user1_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user1->uid))->fetchObject();
     $this->assertEqual($user1_access_time->access, REQUEST_TIME, 'Cookie authenticated user access time is updated.');
 
+    $this->assertNotEqual($user1_access_time_before->access, $user1_access_time->access, 'Access time before and after request are equal.');
+
     // Case 2. Ensure that access time is recorded for basic auth.
     $user = $user2;
-    $basic_auth_provider = restful_get_authentication_handler('basic_auth');
 
     $_SERVER['PHP_AUTH_USER'] = $user2->name;
     $_SERVER['PHP_AUTH_PW'] = $user2->pass_raw;
     $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] = NULL;
     $handler = restful_get_restful_handler('articles');
 
-    // Make a request.
-    $handler->getAccount()->uid;
+    // Perform request authentication.
+    $handler->getAccount();
 
     $user2_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user2->uid))->fetchObject();
     $this->assertEqual($user2_access_time->access, REQUEST_TIME, 'Basic authenticated user access time is updated.');
-
-    // Case 3. Upon re-request, access time isn't updated if it is within the
-    // configured time limit.
-    // $user = $user1;
-
-    // Make a request.
-    // $handler->getAccount()->uid;
-
-    // $user1_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user1->uid))->fetchObject();
-    // $this->assertEqual($user1_access_time->access, REQUEST_TIME, 'Basic authenticated user access time is updated.');
-
   }
 }
--- a/tests/RestfulAuthenticationTestCase.test
+++ b/tests/RestfulAuthenticationTestCase.test
@@ -182,7 +182,7 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
 
     // To begin, we'll set the timestamp for user1 back a little bit.
     $num_updated = db_update('users')
-      ->fields(array('access' => 1000)
+      ->fields(array('access' => 1000))
       ->condition('uid', $user1->uid)
       ->execute();
 
--- a/tests/RestfulAuthenticationTestCase.test
+++ b/tests/RestfulAuthenticationTestCase.test
@@ -180,19 +180,26 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
     // Case 3. Ensure that the timestamp gets updated.
     $user = $user1;
 
+    // Get a timestamp that is in the past.
+    $the_past = REQUEST_TIME - variable_get('session_write_interval');
+
     // To begin, we'll set the timestamp for user1 back a little bit.
     $num_updated = db_update('users')
-      ->fields(array('access' => 1000))
+      ->fields(array('access' => $the_past))
       ->condition('uid', $user1->uid)
       ->execute();
 
     $user1_pre_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user1->uid))->fetchObject();
-    $this->assertEqual($user1_pre_access_time->access, 1000, 'Set user1 access time to a time in the past.');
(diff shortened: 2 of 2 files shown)


e0ipso:
I like this!

Would you be up to writing these 3 tests?
1. Authenticate a request with cookie; and see the access time being updated.
2. Authenticate (for a different user) a request with basic auth; and see the access time being updated.
3. Authenticate the same user right after that; and see the access time unchanged.

Can you see any other test case worth adding?

Thanks for helping RESTful being more robust.

e0ipso:
Adding `needs 2.x port` is a not to myself to port this improvement to the `2.x` branch. If you are up to it, feel free to take this task yourself.

lukewertz:
Since the time interval between updates is configurable, I'd recommend a fourth test which sets the time to a very small interval (a second or two?) and then a sleep and a re-request to ensure that access time is updated. I hate the idea of putting a `sleep` in a test, but it'll be relatively short.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -205,5 +209,20 @@ protected function getOriginalUserSession() {
...
-
+  /**
+   * Set the user's last access time.
+   *
+   * @param object $user
+   *   A drupal user account.
+   *
+   * @see _drupal_session_write()
+   */
+  protected function setAccessTime($user) {
```
amitaibu: $user => $account(so it never collides with the `global $user`

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -205,5 +209,20 @@ protected function getOriginalUserSession() {
...
+    // This logic is pulled directly from _drupal_session_write().
```
amitaibu, line 221: Lets move this to the PHPdocs itself.

plugins/authentication/RestfulAuthenticationManager.php
```diff
@@ -205,5 +209,20 @@ protected function getOriginalUserSession() {
     return $this->originalUserSession;
   }
 
-
+  /**
+   * Set the user's last access time.
+   *
+   * @param object $user
+   *   A drupal user account.
```
amitaibu: => A user object

(no need to mention drupal)

plugins/authentication/RestfulAuthenticationManager.php
e0ipso, line 221: I believe that the comment refers to the code below.

e0ipso:
> then a sleep and a re-request

I'd rather avoid sleeps. Tests are already slow enough ;-)

Why don't we set the variable to a negative value?

lukewertz:
> Since the time interval between updates is configurable, I'd recommend a fourth test which sets the time to a very small interval (a second or two?) and then a sleep and a re-request to ensure that access time is updated. I hate the idea of putting a sleep in a test, but it'll be relatively short.

This is proving to be difficult to test in simpletest. Since the access time is derived from the `REQUEST_TIME` constant, while simpletest is running the `REQUEST_TIME` isn't updated. I can check to see that time is being recorded, but I'm unsure of how to check whether the time is getting updated (beyond the initial set) or, for that matter, not updated appropriately.

I've added my preliminary tests so you can see where I'm headed with it and see if you have any initial feedback or suggestions fro circumventing the `REQUEST_TIME` undesirable behavior I described.

e0ipso:
Can't you play with the contents of the variable? Would that allow to get where you need to be?

plugins/authentication/RestfulAuthenticationManager.php
lukewertz, line 221: While I do reference the `_drupal_session_write()` in the docblock, I wanted to explicitly call out what was being pulled from the function in case more logic is added to this method in the future.

lukewertz:
That's a good idea. I'll hack at that for a bit.

lukewertz:
Well, it was a good idea, but:

http://php.net/manual/en/language.constants.php

> As the name suggests, that value cannot change during the execution of the script (except for magic constants, which aren't actually constants).

lukewertz:
Looks like I'm not alone in running into some odd issues with `REQUEST_TIME`: https://github.com/jhedstrom/drupalextension/issues/179#issuecomment-114519845

Given this, I think testing the fact that the value is stored is probably enough. Since the mechanism for updating the request time is pulled directly from Drupal, I'm not as concerned with that logic as I am with ensuring that I have it called at the correct place in the RESTful module. Simply validating that the values are stored (created) does that.

I propose we have two tests, then:
1) Test cookie authentication
2) Test basic authentication

Both tests should validate that the access time on the user is set when they make an authenticated request. I can add inline comments as to why there isn't more test coverage on it.

Any thoughts?

e0ipso:
> Well, it was a good idea, but:
> 
> http://php.net/manual/en/language.constants.php

I actually meant messing around with things like `variable_set('session_write_interval', -10)`.

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+    global $user;
+
+    $user1 = $this->drupalCreateUser();
+    $user2 = $this->drupalCreateUser();
+
+    $handler = restful_get_restful_handler('main', 1, 5);
+
+    // Case 1. Ensure that access time is recorded for cookie auth.
+    $user = $user1;
+    $cookie_provider = restful_get_authentication_handler('cookie');
```
e0ipso: This is not used anywhere.

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+
+    // Make a request.
+    $handler->getAccount()->uid;
+
+    $user1_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user1->uid))->fetchObject();
+    $this->assertEqual($user1_access_time->access, REQUEST_TIME, 'Cookie authenticated user access time is updated.');
+
+    // Case 2. Ensure that access time is recorded for basic auth.
+    $user = $user2;
+    $basic_auth_provider = restful_get_authentication_handler('basic_auth');
```
e0ipso: This is not used anywhere.

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+    $user1 = $this->drupalCreateUser();
+    $user2 = $this->drupalCreateUser();
+
+    $handler = restful_get_restful_handler('main', 1, 5);
+
+    // Case 1. Ensure that access time is recorded for cookie auth.
+    $user = $user1;
+    $cookie_provider = restful_get_authentication_handler('cookie');
+
+    // Make a request.
```
e0ipso: `// Perform request authentication`

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+    // Case 2. Ensure that access time is recorded for basic auth.
+    $user = $user2;
+    $basic_auth_provider = restful_get_authentication_handler('basic_auth');
+
+    $_SERVER['PHP_AUTH_USER'] = $user2->name;
+    $_SERVER['PHP_AUTH_PW'] = $user2->pass_raw;
+    $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] = NULL;
+    $handler = restful_get_restful_handler('articles');
+
+    // Make a request.
```
e0ipso: `// Perform request authentication`

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+    $user2 = $this->drupalCreateUser();
+
+    $handler = restful_get_restful_handler('main', 1, 5);
+
+    // Case 1. Ensure that access time is recorded for cookie auth.
+    $user = $user1;
+    $cookie_provider = restful_get_authentication_handler('cookie');
+
+    // Make a request.
+    $handler->getAccount()->uid;
```
e0ipso: Do we need the `->uid` part?

``` php
$handler->getAccount();
```

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+    $user = $user2;
+    $basic_auth_provider = restful_get_authentication_handler('basic_auth');
+
+    $_SERVER['PHP_AUTH_USER'] = $user2->name;
+    $_SERVER['PHP_AUTH_PW'] = $user2->pass_raw;
+    $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] = NULL;
+    $handler = restful_get_restful_handler('articles');
+
+    // Make a request.
+    $handler->getAccount()->uid;
```
e0ipso: Do we need the `->uid` part?

``` php
$handler->getAccount();
```

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+
+    // Case 1. Ensure that access time is recorded for cookie auth.
+    $user = $user1;
+    $cookie_provider = restful_get_authentication_handler('cookie');
+
+    // Make a request.
+    $handler->getAccount()->uid;
+
+    $user1_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user1->uid))->fetchObject();
+    $this->assertEqual($user1_access_time->access, REQUEST_TIME, 'Cookie authenticated user access time is updated.');
```
e0ipso, line 162: Can we record the access time before the update in L158 and do the following?

``` php
$this->assertNotEqual($user1_access_time_before, $user1_access_time->access);
```

tests/RestfulAuthenticationTestCase.test
```diff
@@ -138,4 +138,52 @@ class RestfulAuthenticationTestCase extends RestfulCurlBaseTestCase {
...
+    $_SERVER['REDIRECT_HTTP_AUTHORIZATION'] = NULL;
+    $handler = restful_get_restful_handler('articles');
+
+    // Make a request.
+    $handler->getAccount()->uid;
+
+    $user2_access_time = db_query('SELECT access FROM {users} WHERE uid = :d', array(':d' => $user2->uid))->fetchObject();
+    $this->assertEqual($user2_access_time->access, REQUEST_TIME, 'Basic authenticated user access time is updated.');
+
+    // Case 3. Upon re-request, access time isn't updated if it is within the
```
e0ipso: If setting negative values does not lead anywhere, let's remove the commented out code.

e0ipso:
Also, to get different request times in your tests, did you try something like:

``` php
$this->drupalGet('/api/v1.5/main/');
```

or

``` php
$this->httpRequest('/api/v1.5/main/', …);
```

?

For the second one you will need to make the test class to extend `RestfulCurlBaseTestCase` (if it's not doing it already).

lukewertz:
> I actually meant messing around with things like variable_set('session_write_interval', -10).

Sorry, I was too close to the forest to see the trees on that. I had already tried that at that point, but of course you had no way of knowing that. Your comment made me think that I might be able to hijack the `REQUEST_TIME`, but of course that didn't work.

I cleaned up the tests based on your feedback and reflecting the simpler testing approach.

I did try making a request with a different `REQUEST_TIME` through the two methods you mentioned (`drupalGet` and `httpRequest`) but couldn't get the requests to be made with a time other than the defined constant that is set when the tests start running. I also tried making requests directly through the `$handler->get()` and got the same result, of course.

Given all of this, I do think that the simpler testing approach is probably okay for this.

Thanks so much for your thorough feedback on this! I really appreciate it.

amitaibu:
Maybe after asserting the `access == REQUEST_TIME` you can db_update() the value to the past - so you'll able to verify that hitting the API will reset the value back to `REQUEST_TIME`

lukewertz:
Great idea, @amitaibu! Test coming just now.

lukewertz:
Okay! Added a test based on your feedback, @amitaibu. I set the last access time for user 1 to the configured-past (REQUEST_TIME - session_write_interval) which seems like the most reasonable way to test that.

All tests are passing now, too. Thanks so much @amitaibu and @e0ipso for your help in getting this written. I really appreciate it!

lukewertz:
Hi @e0ipso,

Is there anything left on this PR that needs to be done?

Thanks!

e0ipso:
@lukewertz thanks for the reminder.

This looks awesome, thanks! I'll be merging this asap.

e0ipso:
Thanks for the awesome PR!

Merged in 797dc601cd7ebe274f26e46260d4aa52527a86b0.

amitaibu:
:+1:

e0ipso:
This was ported to 7.x-2.x in https://github.com/RESTful-Drupal/restful/pull/732. Removing tag.

PR #673, opened 2015-10-07
[BUGFIX] Allow Patch/ Put NULL values

ordavidil:
This PR fixing a bug when writing `null` value to a property.


tests/RestfulUpdateEntityTestCase.test
```diff
@@ -59,6 +59,30 @@ class RestfulUpdateEntityTestCase extends DrupalWebTestCase {
     catch (Exception $e) {
       $this->pass('User cannot update using PUT method an entity with invalid property name.');
     }
+
+    $new_label = $this->randomName();
+    $new_body = $this->randomName();
+
+    // Wrapping the body text with a <p> tag and line break, to imitate the way
+    // drupal treats plain text in the body field.
+    $new_body = "<p>$new_body</p>\n";
```
amitaibu: that's weird :) - can we remove this?

tests/RestfulUpdateEntityTestCase.test
```diff
@@ -59,6 +59,30 @@ class RestfulUpdateEntityTestCase extends DrupalWebTestCase {
...
+
+    $request = array(
+      'label' => $new_label,
+      'body' => $new_body
+    );
+    // Setting new value for 'body'.
+    $result = $handler->put($id, $request);
+    // Making sure the new body has been set.
+    $this->assertEqual($result[0]['body'], $new_body);
```
amitaibu: you can `strip_tags($result[0]['body'])` instead of the `<p>` hack

tests/RestfulUpdateEntityTestCase.test
```diff
@@ -59,6 +59,30 @@ class RestfulUpdateEntityTestCase extends DrupalWebTestCase {
...
+
+    // Removing the body value.
```
amitaibu, line 75: Removing => Remove

plugins/restful/RestfulEntityBase.php
```diff
@@ -680,16 +680,19 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
       }
 
       $property_name = $info['property'];
-      if (!isset($request[$public_field_name])) {
-        // No property to set in the request.
+
+      if (!array_key_exists($public_field_name, $request)) {
```
amitaibu, line 684: this part definitely deserves a comment

plugins/restful/RestfulEntityBase.php
```diff
@@ -680,16 +680,19 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
...
         if ($null_missing_fields && $this->checkPropertyAccess('edit', $public_field_name, $wrapper->{$property_name}, $wrapper)) {
-          // We need to set the value to NULL.
-          $wrapper->{$property_name}->set(NULL);
+          $field_value = NULL;
+        }
+        else {
+          continue;
```
amitaibu, line 693: small comment

amitaibu:
Looks good, I've added minor comments. It was really surprising to realize we couldn't patch/put NULL values until now.

@e0ipso this probably needs porting to 2.x as-well

ordavidil:
@amitaibu 
Can you check again please?
Thanks.

amitaibu:
Merged, thanks!

amitaibu:
@e0ipso this probably needs 2.x porting

jeff-h:
Nice :) This should close my PR #680 except that this still, I think, needs to reverse the function description:

```
Determine if properties that are missing from the request array should
be treated as NULL, or should be skipped. **Defaults to FALSE, which will
set the fields to NULL.**
```

...is actually incorrect, I think. It would only set the fields to NULL if this bool is TRUE.

jeff-h:
This will also close my PR https://github.com/RESTful-Drupal/restful/pull/684

e0ipso:
This is ported to 7.x-2.x in #727. Removing the tag.

PR #674, opened 2015-10-09
[BUGFIX] Allow having and idField to use with JSON API

e0ipso:
JSON API requires an identifier for its display. Since we cannot
provide a default id field, we need to specify it in the annotation.

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderDbQuery.php
+++ b/src/Plugin/resource/DataProvider/DataProviderDbQuery.php
@@ -8,7 +8,6 @@
 namespace Drupal\restful\Plugin\resource\DataProvider;
 
 use Drupal\restful\Exception\BadRequestException;
-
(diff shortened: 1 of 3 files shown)


e0ipso:
I discovered this issue when dealing with this: https://github.com/RESTful-Drupal/restful/issues/675

src/Plugin/resource/DataProvider/DataProviderDbQuery.php
```diff
@@ -225,21 +226,6 @@ protected function mapDbRowToPublicFields($row) {
         // Allow passing the value in the request.
         continue;
       }
-//      // If there is a callback defined execute it instead of a direct mapping.
```
e0ipso, line 228: This was some commented out code.

e0ipso:
I going to merge this, since it's test covered and has been QA'ed in another place.

Please, feel free to create as many follow ups as necessary.

PR #681, opened 2015-10-14
7.x 2.x Undefined index notices during OPTIONS request (label, description, required)

rafatwork:
Receiving multiple undefined index notices while doing a OPTIONS request. Some examples:

PHP Notice:  Undefined index: label in /sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php on line 1181

PHP Notice:  Undefined index: description in /sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php on line 1182

PHP Notice:  Undefined index: required in /sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php on line 1197

### DIFF
--- a/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -1170,11 +1170,6 @@ class ResourceFieldEntity implements ResourceFieldEntityInterface {
         'label' => $field_instance['label'],
         'description' => $field_instance['description'],
       ));
-      $field_info = field_info_field($this->getProperty());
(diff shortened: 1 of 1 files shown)


e0ipso:
I believe that removing the code is not the correct fix. We should be adding the pertinent `empty(…)` checks instead.

rafatwork:
Agreed - perhaps it's not the right fix.

Although, perhaps you could elaborate why first the label & description are
added based on the field instance, to be overwritten afterwards by the
field info?

As to the required property: any examples where this property key is
available? I couldn't find this on various installations.

PR #682, opened 2015-10-14
[FEATURE] Clear caches on invalidation for embedded resources

e0ipso:
If a resource was cached in a denormalized response as an embedded
resource, and that resource needs to be invalidated; we also need to
clear all the resources containing it.

### DIFF
--- a/restful.entity.inc
+++ b/restful.entity.inc
@@ -86,9 +86,11 @@ function restful_entity_update($entity, $type) {
     if (!$handler->hasSimpleInvalidation()) {
       return;
     }
-    $handler
-      ->getCacheController()
-      ->clear($hash);
+    // You can get away without the fragments for a clear.
+    $cache_object = new RenderCache(new ArrayCollection(), $hash, $handler->getCacheController());
+    // Do a clear with the RenderCache object to also remove the cache fragment
+    // entities.
+    $cache_object->clear();
   }
 }
 
@@ -106,10 +108,11 @@ function restful_entity_delete($entity, $type) {
     if (!$handler->hasSimpleInvalidation()) {
       return;
     }
-    // Resource takes care of its own invalidation.
-    $handler
-      ->getCacheController()
-      ->clear($hash);
+    // You can get away without the fragments for a clear.
+    $cache_object = new RenderCache(new ArrayCollection(), $hash, $handler->getCacheController());
+    // Do a clear with the RenderCache object to also remove the cache fragment
+    // entities.
+    $cache_object->clear();
   }
 }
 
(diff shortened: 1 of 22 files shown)


amitaibu:
Needs review?

e0ipso:
Not just yet

e0ipso:
This is ready to be reviewed @amitaibu.

amitaibu:
Cool, I just need to take a deep breath before ;)

restful.entity.inc
```diff
@@ -96,9 +86,11 @@ function restful_entity_update($entity, $type) {
...
       return;
     }
-    $handler
-      ->getCacheController()
-      ->clear($hash);
+    // You can get away without the fragments for a clear.
+    $cache_object = new RenderCache(new ArrayCollection(), $hash, $handler->getCacheController());
+    // Do a clear with the RenderCache object to also remove the cache fragment
+    // entities.
+    $cache_object->clear();
```
e0ipso, line 93: I was not clearing the render cache properly before. If you do it this way, you clear the caches and delete the involved cache fragments.

src/Formatter/FormatterManager.php
```diff
@@ -212,9 +212,9 @@ public function getPlugin($instance_id) {
   protected function getPluginByName($name) {
     /* @var FormatterInterface $formatter */
     $formatter = $this->plugins->get($name);
-    $formatter->setConfiguration(array(
-      'resource' => $this->resource,
-    ));
+    if ($this->resource) {
+      $formatter->setResource($this->resource);
+    }
```
e0ipso, line 217: This was a problem for tests. The resource plugin was stored and could not be changed for the subsequent test. This was not an issue during regular processing.

src/Plugin/formatter/Formatter.php
```diff
@@ -162,26 +207,84 @@ protected function createCacheController($data) {
...
+   *
+   * @param bool|string[] $allowed_fields
+   *   The list of allowed fields in dot notation.
+   * @param string $prefix
+   *   The prefix used to select the fields and to remove from the front.
+   *
+   * @return bool|string[]
+   *   The new allowed fields for the nested sub-request.
+   */
+  protected static function unprefixInputOptions($allowed_fields, $prefix) {
```
e0ipso, line 275: Given some query param like `a,b,b.c,b.d,e,f,f.g,f.h,f.i`, this is used so we can pass only the request options `b.c,b.d` when making a sub-request to the resource for the `b` relationship.

amitaibu:
I will review on Thursday - marked in my calendar :)

e0ipso:
That's great!

Let me know if you want to go through it together.

PR #684, opened 2015-10-14
Don't throw exception when request contains fields with NULL

jeff-h:
At present the code throws `RestfulBadRequestException` if any of the request fields contain NULL. NULL should be a legitimate field value unless validation says otherwise.

This change fixes that, and also renames the variable `$original_request` to `$unprocessed_fields` which I felt better described its purpose.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -663,12 +663,12 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
 
     static::cleanRequest($request);
     $save = FALSE;
-    $original_request = $request;
+    $unprocessed_fields = $request;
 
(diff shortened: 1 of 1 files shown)


jeff-h:
Also, I'm thinking the tests should include some NULL-handling tests, but I'm unsure how to do that.

jeff-h:
I'm assuming the tests have failed for some other reason. My PR wasn't _that_ dangerous :D

e0ipso:
It would be better to use `array_key_exists`. Also, some tests will need to be adapted to the new behavior.

I agree that `NULL` should be a valid _existing_ value.

e0ipso:
Also, there are merge conflicts in this branch

amitaibu:
Didn't #673 fix this?

jeff-h:
@amitaibu — yeah, #673 has fixed this.

PR #686, opened 2015-10-15
[BUGFIX] Allow cached resources to have custom calls

e0ipso:
@amitaibu this is ready for review.

### DIFF
--- a/src/Plugin/resource/Decorators/ResourceDecoratorBase.php
+++ b/src/Plugin/resource/Decorators/ResourceDecoratorBase.php
@@ -311,6 +311,21 @@ abstract class ResourceDecoratorBase extends PluginBase implements ResourceDecor
 
   /**
    * If any method not declared, then defer it to the decorated field.
+   *
+   * This decorator class is proxying all the calls declared in the
+   * ResourceInterface to the underlying decorated resource. But it is not
+   * doing it for any of the methods of the parents of ResourceInterface.
+   *
+   * With this code, any method that is not declared in the class will try to
+   * make that method call it in the decorated resource.
+   *
+   * @param string $name
+   *   The name of the method that could not be found.
+   * @param array $arguments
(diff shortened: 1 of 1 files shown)


amitaibu:
Can you explain a bit please :)

e0ipso:
The cache decorator is proxying all the calls declared in the `ResourceInterface` to the decorated underlying resource. But it was not doing it for any of the methods of the parents of `ResourceInterface`.

With this code, any method that is not declared in the class will try to make that method call it in the decorated resource.

src/Plugin/resource/Decorators/ResourceDecoratorBase.php
```diff
@@ -309,4 +309,26 @@ public function getUrl(array $options = array(), $keep_query = TRUE, RequestInte
...
   }
 
+  /**
+   * If any method not declared, then defer it to the decorated field.
+   *
+   * This decorator class is proxying all the calls declared in the
+   * ResourceInterface to the underlying decorated resource. But it is not
+   * doing it for any of the methods of the parents of ResourceInterface.
+   *
+   * With this code, any method that is not declared in the class will try to
```
amitaibu, line 319: and if the method doesn't exist? Maybe we want to add a more descriptive  error message?

e0ipso, line 319: If the method does not exist we are falling back to PHP's native behavior, which is a fatal error.

Something like

``` php
$decorator->validMethodInDecorator();
```

and

``` php
$decorator->validMethodInDecorated();
```

will yield expected results.

Something like:

``` php
$decorator->invalidMethod()
```

will also result in the expected behavior, which is a fatal error.

amitaibu:
Had tight schedule with zendcon going on. Can you merge this, and I'll do a post merge review? Sorry :/

e0ipso:
zendcon?? So jealous!!

PR #692, opened 2015-10-20
[BUGFIX] Make sure we include all fields in includes

e0ipso:
By default if we are including a nested resource and not selecting
any particular field in it, then we should be adding all fields for
it. Additionally, fix a PHP warning accessing non-array properties.


e0ipso:
This PR builds on top of https://github.com/RESTful-Drupal/restful/pull/690

e0ipso:
@amitaibu extra tests added. This is ready for review.

PR #693, opened 2015-10-20
[BUGFIX] Centralize error handling

e0ipso:
So we can have a consistent way to add headers and outputs


e0ipso:
This PR builds on top of https://github.com/RESTful-Drupal/restful/pull/690.

PR #699, opened 2015-10-25
[BUGFIX] Max cap for the range to avoid perf problems

e0ipso:
Otherwise the consumers can request 100k elements in a single request.


e0ipso:
This PR needs a test.

e0ipso:
Test added.

PR #700, opened 2015-10-25
[FEATURE] All resource menuItem default to api/

e0ipso:
Make all the paths created from the resource to be nested under the
API's path prefix.

### DIFF
--- a/modules/restful_token_auth/restful_token_auth.module
+++ b/modules/restful_token_auth/restful_token_auth.module
@@ -46,8 +46,12 @@ function restful_token_auth_menu_alter(&$items) {
     return;
   }
   foreach ($access_token_plugin_definitions as $access_token_plugin_definition) {
-    if (!empty($items[$access_token_plugin_definition['menuItem']])) {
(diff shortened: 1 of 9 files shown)


e0ipso:
This changes a bit from `1.x`.

**menuItem:** Determines the URL where the resource is going to live. All paths specified using this definition property will be nested under the API's prefix. Ex: using `menuItem: "my-resource"` will result in `https://example.org/api/my-resource`.

amitaibu:
so this means that `file-upload` which is binary and non-restful will be under `/api` as-well?

e0ipso:
> so this means that file-upload which is binary and non-restful will be under /api as-well?

Yes. Actually, `file-upload` was already under `/api/`.

If we don't want that, we can modify this via regular `hook_menu_alter`.

e0ipso:
See: https://github.com/RESTful-Drupal/restful/pull/700/files#diff-269d980f2e7442e85e3c457c40cece2bL49

amitaibu:
Ok :)
Ready for review?

e0ipso:
Yes!

PR #702, opened 2015-10-25
[BUGFIX] Token resources need a special formatter

e0ipso:
Add a special formatter for token resources. Deprecate the use of
restful_unprepared_delivery in favor of a more appropriate solution.
Token endpoints need a special format, that means that they deserve
their special formater.


e0ipso:
Tests have been fixed.

PR #703, opened 2015-10-26
[Bugfix] Issue with minor_version=0

flashvnn:
When referenced resource with minor_version=0 it alway looking for lasted version because empty($resource['minor_version']) always  return true.


e0ipso:
This PR has been merged in 4df882832103a653b9a243127e8a2aa2fecb6222

Thanks a lot for contributing.

PR #707, opened 2015-10-27
[CLEANUP] Add testing for property access

e0ipso:
Fix and enable the testing for property access checks.


e0ipso:
This PR builds on top of 706.

e0ipso:
This should bring us **very** close to feature parity.

PR #709, opened 2015-10-29
[CLEANUP] Drop tests for unsupported features

e0ipso:
We don't need theses tests anymore.

### DIFF
--- a/restful.info
+++ b/restful.info
@@ -11,7 +11,6 @@ registry_autoload[] = PSR-4
 
 ; Tests
 files[] = tests/RestfulAuthenticationTestCase.test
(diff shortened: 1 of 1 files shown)


e0ipso:
@amitaibu I am dropping the tests for the following deprecated features:
- Pass through: now we have the `methods` array for public fields that specify under HTTP methods a field is available.
- Autocomplete: we can achieve autocomplete with regular filtering and sparse fieldsets.

amitaibu:
:+1: I won't miss those tests ;)

amitaibu:
btw, how do you reach autocomplete with the sparse fieldsets?

PR #710, opened 2015-10-29
[FEATURE] Add variables example

e0ipso:
This creates an example of how you can create a resource from scratch.


e0ipso:
One more step closer. :green_heart:

PR #711, opened 2015-10-29
Replace ui-select2 with ui-select.

kporras07:
This is a first attempt to replace deprecated ui-select2 with ui-select. 
If merged, this should fix https://github.com/RESTful-Drupal/restful/issues/286

It would be great if you could take a look.


e0ipso:
This has been merged in a921063132f6ddef6c3812b7dc6bdeed66ab3cfa.

Thanks a lot for your contribution!

PR #712, opened 2015-10-31
[FEATURE] Multiple bundles resources example

e0ipso:
Plus documentation.


tests/modules/restful_test/src/Plugin/resource/entity_test/EntityTests__1_0.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+class EntityTests__1_0 extends ResourceEntity implements ResourceInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function publicFields() {
+    $public_fields = parent::publicFields();
+
+    $public_fields['main_bundle'] = array(
```
amitaibu, line 42: I wonder if this is indeed a "best practice". It requires the client to know about the bundles, rather then thinking of it as a single "resource".

What if when a resource has multiple bundles, you silently skip over non existing fields (only for GET)?

e0ipso, line 42: > It requires the client to know about the bundles

Not sure I understand why. We could've called instead of `'main_bundle'` :arrow_right: `'batman!'`, …

> What if when a resource has multiple bundles, you silently skip over non existing fields (only for GET)?

I think that there are two big problems with that approach:
- This behaviour represents a drupalism where we bundle together things that don't belong together. Outside of Drupal it's hard to understand that _page_ and _article_ are the same thing (node) but a custom entity type _post_ is totally different. So grouping them together beyond the basic common properties is weird.
- Silently skipping fields altogether goes against [this](https://github.com/RESTful-Drupal/restful/pull/712/files#diff-243b408b86b557a6039177fd9849e028R4). And makes things like automated validation via Apiary or Swagger proxies close to impossible.

Thats why I was thinking that a multiple bundles resource should be a data provider that retrieves data about several bundles, and contains references to the resources that contain the presentation for those individual bundles.

I'd love to hear your thoughts on these points.

amitaibu, line 42: maybe I'm missing your point :)

What would be the output of a call to this resource?

e0ipso, line 42:
``` http
GET /entity_tests?include=batman!,tests_bundle HTTP 1.1

{
  "data": [{
    "type": "entity_tests",
    "id": 1,
    "relationships": {
      "batman!": {
        "data": {"type": "main", "id": 1}
      }
    }
  }, {
    "type": "entity_tests",
    "id": 2,
    "relationships": {
      "tests_bundle": {
        "data": {"type": "tests", "id": 2}
      }
    }
  }],
  "included": [
    {
      "type": "main",
      "id": 1,
      "attributes": {
        "label": "I belong to the main bundle.",
        "specific_main_field": true
      }
    }
    {
      "type": "tests",
      "id": 2,
      "attributes": {
        "label": "I belong to the tests bundle.",
        "this_is_different": "Indeed!"
      }
    }
  ]
}
```

amitaibu, line 42: I see, JSON API is indeed cleaning it up nicely :) 

but just to see that I understand, could you please show how would `simple json` formatter look.

e0ipso, line 42: For simple JSON it would be something like:

``` json
{
  "data": [
    {
      "id": 1,
      "batman!": {
        "label": "I belong to the main bundle.",
        "specific_main_field": true
      },
      "tests_bundle": null
    },
    {
      "id": 2,
      "batman!": null,
      "tests_bundle": {
        "label": "I belong to the tests bundle.",
        "this_is_different": "Indeed!"
      }
    }
  ],
  "links": {}
}
```

Both payloads have been manually generated so there may be some errors.

amitaibu, line 42: yeah, no problem, I just wanted to make sure I understand. Ok, it looks great. Sorry for the noise ;)

e0ipso, line 42: > I see, JSON API is indeed cleaning it up nicely :)
> 
> but just to see that I understand, could you please show how would simple json formatter look.

Awww, we almost finish each other's sentences…

amitaibu, line 42: :+1:

e0ipso, line 42: > yeah, no problem, I just wanted to make sure I understand.

Great! It's cool to have a sanity check.

GaxZE:
Thanks for directing me here @e0ipso - It's been a great deal of help.

I have two node types (blog and event) which I have defined as two separate resources. I then using this example, have tried to combined the two into a 'combined resource' but so far I'm not able to bring in any content from either resource. I'm simply getting two NULL values for each public_field in my combined resource.

``` php
  /**
   * {@inheritdoc}
   */
    $public_fields = parent::publicFields();

    $public_fields['blog'] = array(
      'property' => 'pid',
      'class' => '\Drupal\api\Plugin\resource\blog\v1\blog__1_0',
      'resource' => array(
        'name' => 'blog',
        'majorVersion' => 1,
        'minorVersion' => 0,
      ),
    );

    $public_fields['event'] = array(
      'property' => 'pid',
      'class' => '\Drupal\api\Plugin\resource\event\v1\event__1_0',
      'resource' => array(
        'name' => 'event',
        'majorVersion' => 1,
        'minorVersion' => 0,
      ),
    );

    return $public_fields;
  }
```

Outputs:

``` json
{
    "data": 
[
{
    "id": ​1,
    "label": "Placeholder blog",
    "self": "http://local/api/v1.0/entity_tests/1",
    "blog": null,
    "event": null
},
{
    "id": 2,
    "label": "Placeholder Event",
    "self": "http://local/api/v1.0/entity_tests/2",
    "blog": null,
    "event": null
},
```

e0ipso:
You probably want:

``` php
  /**
   * {@inheritdoc}
   */
    $public_fields = parent::publicFields();

    $public_fields['blog'] = array(
      'property' => 'nid',
      'class' => '\Drupal\restful\Plugin\resource\Field\ResourceFieldEntityReference',
      'resource' => array(
        'name' => 'blog',
        'majorVersion' => 1,
        'minorVersion' => 0,
      ),
    );

    $public_fields['event'] = array(
      'property' => 'nid',
      'class' => '\Drupal\restful\Plugin\resource\Field\ResourceFieldEntityReference',
      'resource' => array(
        'name' => 'event',
        'majorVersion' => 1,
        'minorVersion' => 0,
      ),
    );

    return $public_fields;
  }
```

The rationale behind this is: «I want to create a resource that lists nodes. Then for every content type, I create a public field that uses the `nid` to select the information from the specific resource (event, blog, …). Therefore I want to treat the `nid` field not as an integer but as an entity reference».

Let me know if that makes sense.

e0ipso:
@GaxZE if that fixes your problem, please take the time (part of the time you saved) to write some documentation in the wiki :-)

GaxZE:
Thanks @e0ipso that was the exact change required. I will have a go contributing to the wiki. This is a really sound module. :thumbsup:

GaxZE:
@e0ipso I haven't forgot to contribute to the wiki, I'm planning on writing a tutorial to document my experiences on this module. I've hit a few snags and wanted to know if restful module could turn the resource(content type in my case) into a property. This would save on a huge amount of `null` values being returned. Example:

### Before

``` json
{
  "blog": null,
  "event": null,
  "media": null,
  "page": {
    "language": "en",
    "nid": "8",
    "title": "Home ",
    "body": "<p>Home main content </p>",
    "hero-promo-image": [ ],
    "hero-promo-video": [ ],
    "media-image": [ ],
    "tags": [ "tag1" ]
  },
  "listing-page": null,
}
```

### After

``` json
{
  "language": "en",
  "content-type": "general-info",
  "nid": "8",
  "title": "Home ",
  "body": "<p>Home main content </p>",
  "hero-promo-image": [ ],
  "hero-promo-video": [ ],
  "media-image": [ ],
  "tags": [ "tag1" ]
}
```

Is this possible with restful?

e0ipso:
> Is this possible with restful?

No. And that is by design.

An entity returned from a resource should have a static schema (meaning not having dynamic if X then we have this set of fields, under this conditions Y happens). BTW: we are actually generating that schema for you after an OPTIONS call.

e0ipso:
If the consumer knows this is a resource after a page, they can query the page resource directly.

Additionally you can do `?fields=page` to remove all the fields that are not pages.

PR #717, opened 2015-11-02
[BUGFIX] Fix the resource plugin reference

e0ipso:
So the cache placeholder population can load the plugin.


e0ipso:

PR #720, opened 2015-11-03
Use proper 'sub_property' instead when getting public field value fro…

xuxucode:
For example: commerce_order of [Commerce Marketplace](https://www.drupal.org/sandbox/maciej.zgadzaj/1950386):

``` php
class RestfulCommerceOrder extends RestfulBaseEntity {

    $public_fields['seller'] = array(
      'property' => 'commerce_store',
      'sub_property' => 'creator',
      'resource' => array(
        'user' => 'users',
      ),
    );
```

should  use `creator` property rather than `commerce_store` property to get value from users resource.


e0ipso:
I believe in this case you are misusing the `'sub_property'` to navigate to the user resource. This is not the intended use for it. Instead you should have an intermediate resource for the commerce store, since that is an entity itself.

xuxucode:
thanks, I will try it.

PR #725, opened 2015-11-07
[BUGFIX] Port #489 to 7.x-2.x

e0ipso:
Generate access token and refresh token for correct users.


e0ipso:

PR #727, opened 2015-11-07
[BUGFIX] Check for property access after setting the value

e0ipso:
The result of the access check may depend on the new value being
set during write operations.

This is a port of #539 to 7.x-2.x.


e0ipso:
Tests are green. Merging.

PR #728, opened 2015-11-07
[FEATURE] Allow more operators per filters

e0ipso:
Adds more flexibility to query the exact elements needed.


e0ipso:

PR #730, opened 2015-11-07
[BUGFIX] Only count the published nodes

e0ipso:
Add the NODE_PUBLISHED condition to the count query to match the
list query. Otherwise pagination is off.


e0ipso:

PR #731, opened 2015-11-07
[BUGFIX] Refresh token cleanup

e0ipso:
Delete the access token associated with a refresh token when a new
refresh token is created. Also clean up the reference table.


e0ipso:
Tests are green. Merging.

PR #732, opened 2015-11-08
[FEATURE] Record last-access time for users

e0ipso:
When a user's account is loaded, the user table should be updated
to indicate that they have made an authenticated request. This
mirrors the logic in _drupal_session_write().


e0ipso:
Tests are green. Merging.

PR #735, opened 2015-11-10
Fixes a typo in checking permissions before deleting an entity.

lukewertz:
The `deleteEntity` method right now validates that the user has "update" permission, not "delete" permission. I have a use case where a user is allowed to update but not delete a node, thus this is returning a false-positive.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -594,7 +594,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
    * {@inheritdoc}
    */
   public function deleteEntity($entity_id) {
-    $this->isValidEntity('update', $entity_id);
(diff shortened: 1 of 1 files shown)


xuxucode:
same issue on 2.x.
A user can delete own articles even through `delete own article content` permission not permitted.

e0ipso:
Yikes! Sorry for taking so long to get to this, but I had to refactor the Travis CI integration. The pear channel for old versions of drush was taken down.

e0ipso:
I just kicked the tests again, but I'd like to add test coverage for this. Would you be up to it @lukewertz?

e0ipso:
Closed in favor of https://github.com/RESTful-Drupal/restful/pull/743.

e0ipso:
This was also fixed in 2.x. 915f0dfd92fb81ce767ad09bbe7084ec3af9ea83

Thank you gents!

PR #736, opened 2015-11-11
[BUGFIX] Make sure the description property is always available in the base RestfulException object.

hussainweb:
The `RestfulException` class contains a method `getDescription` that accesses a property called `description`. This property is not always available and in certain cases, logs errors like these:

```
Notice: Undefined property: RestfulException::$description in RestfulException->getDescription() (line 38 of /.../restful/exceptions/RestfulException.php)
```

On investigating, I found this method was introduced in 484a1744 with the description property added in child classes of `RestfulException`. If an application is extending that class and not defining the property, it results in the error above. While you may want to make it explicit that the description should be set in the child classes, it is not enforced which results in these notices. I think it is a good idea to at least keep it in the base class so that if a child class chooses not to use this, it doesn't have to.

On that note, why not just use the `message` property? I am just curious.

### DIFF
--- a/exceptions/RestfulBadRequestException.php
+++ b/exceptions/RestfulBadRequestException.php
@@ -15,9 +15,7 @@ class RestfulBadRequestException extends RestfulException {
   protected $code = 400;
 
   /**
-   * Defines the description.
-   *
-   * @var string
+   * {@inheritdoc}
    */
   protected $description = 'Bad Request.';
 
--- a/exceptions/RestfulException.php
+++ b/exceptions/RestfulException.php
@@ -29,6 +29,13 @@ class RestfulException extends Exception {
   protected $headers = array();
 
   /**
+   * Defines the description.
+   *
+   * @var string
+   */
+  protected $description;
+
+  /**
    * Gets the description of the exception.
    *
    * @return string


hussainweb:
I should add that this happened in my application because we directly throw `RestfulException` some of the times. I think that is supported because I see restful itself throwing this in certain places.

Also, the Travis build failed due to some other error (PEAR related). I can't restart it.

e0ipso:
That looks legit.

> we directly throw RestfulException some of the times. I think that is supported because I see restful itself throwing this in certain places

That is not forbidden, although it's discouraged. If RESTful is throwing those we'd need to follow up this issue and use the correct exception flavor.

Thanks for the contribution!

hussainweb:
Thank you for your comments @e0ipso. Do you see any other work left on this issue?

I thought we should replace all comments for `description` in sub-classes with `{@inheritdoc}`, but that could be a different issue. What do you think?

e0ipso:
> I thought we should replace all comments for description in sub-classes with {@inheritdoc}, but that could be a different issue. What do you think?

That seems good to me!

hussainweb:
@e0ipso, anything else pending here? I'd love to help if I can to speed things up and get this merged.

Question: How often do you push to repositories on d.o so that it can be packaged as dev?

e0ipso:
So... The pear service for drush was taken down, and now the Travis tests
are broken. I'm rebuilding the Travis script, and that may take a bit.

After that is in place I'll kick the tests here and merge when they come
back green.

I push every commit to d.o

hussainweb:
@e0ipso, thanks for the update. If there is anything I can do to help, please let me know.

e0ipso:
Closed in favor of https://github.com/RESTful-Drupal/restful/pull/739.

PR #739, opened 2015-11-12
[BUGFIX] Make sure the description property is always available in the base RestfulException object.

e0ipso:
From the original PR:
The `RestfulException` class contains a method `getDescription` that accesses a property called `description`. This property is not always available and in certain cases, logs errors like these:

```
Notice: Undefined property: RestfulException::$description in RestfulException->getDescription() (line 38 of /.../restful/exceptions/RestfulException.php)
```

On investigating, I found this method was introduced in 484a1744 with the description property added in child classes of `RestfulException`. If an application is extending that class and not defining the property, it results in the error above. While you may want to make it explicit that the description should be set in the child classes, it is not enforced which results in these notices. I think it is a good idea to at least keep it in the base class so that if a child class chooses not to use this, it doesn't have to.

On that note, why not just use the `message` property? I am just curious.

### DIFF
--- a/exceptions/RestfulBadRequestException.php
+++ b/exceptions/RestfulBadRequestException.php
@@ -15,9 +15,7 @@ class RestfulBadRequestException extends RestfulException {
   protected $code = 400;
 
   /**
-   * Defines the description.
-   *
-   * @var string
+   * {@inheritdoc}
    */
   protected $description = 'Bad Request.';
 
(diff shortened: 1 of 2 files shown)


e0ipso:
@hussainweb if you can port this PR to 7.x-2.x that'd be great!

e0ipso:
Merged!

Thanks for the contribution @hussainweb.

e0ipso:
e809091576a9d561f62baa627bb13258ba728a7f

hussainweb:
@e0ipso, I think this is not needed in 2.x. The RestfulException class there already has the description property.

e0ipso:
Oh!

Awesome then! Thanks.

PR #740, opened 2015-11-12
Modify comments on inherited properties to use inheritdoc

hussainweb:
This is in follow up to #736 and #739. This commit cleans up all inherited properties to use `{@inheritdoc}` instead of writing the comments each time. There also seems to be an oversight where one of the exception classes did not inherit from `RestfulException` and that is fixed here.


e0ipso:
This was fixed in 9919794fbcfedc44a84d916773498e021327e896.

Thanks again for all your work!

PR #741, opened 2015-11-12
[CLEANUP] Modify comments on inherited properties to use inheritdoc

hussainweb:
Similar to #740, but for the 7.x-2.x branch.


e0ipso:
Fixed in 360bbdfd598fdb31f790c7a7ea8e32a32809ca72.

Many thanks for the contribution!

PR #743, opened 2015-11-15
Fixes a typo in checking permissions before deleting an entity.

e0ipso:
From the original #735 by @lukewertz:

> The `deleteEntity` method right now validates that the user has "update" permission, not "delete" permission. I have a use case where a user is allowed to update but not delete a node, thus this is returning a false-positive.


e0ipso:
On second thoughts, we don't need an extra test for this.

e0ipso:
This has been merged. As usual, thanks for contributing!

PR #745, opened 2015-11-16
Port getQueryCount() method to 2.x

xuxucode:
I have some DataProviders and need to override count() method, but feel it complicated, the `getQueryCount()` method on 1.x is very convenient, so port it to 2.x.

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -178,20 +178,8 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
    * {@inheritdoc}
    */
   public function count() {
-    $query = $this->getEntityFieldQuery();
-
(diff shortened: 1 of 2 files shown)


e0ipso:
This looks good, although it seems that Travis CI is taking a vacation. Let's wait to see if the tests are triggered.

Thanks for contributing.

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -540,6 +528,29 @@ protected function getQueryForList() {
   }
 
   /**
+   * Prepare a query for RestfulEntityBase::count().
+   *
+   * @return \EntityFieldQuery
```
e0ipso, line 533: This returns a `\Drupal\restful\Util\EntityFieldQuery`.

e0ipso, line 533: I fixed that myself.

e0ipso:
This is merged. Thanks for contributing!

PR #746, opened 2015-11-18
Allow bundles FALSE

xuxucode:
`SQLSTATE[42S22]: Column not found: 1054 Unknown column 'comment.node_type' in 'where clause'` occurs on below `comments` resource:

``` php
/**
 * Class Comments__1_0
 * @package Drupal\restful_example\Plugin\resource\comment
 *
 * @Resource(
 *   name = "comments:1.0",
 *   resource = "comments",
 *   label = "Comments",
 *   description = "Export the comments with all authentication providers.",
 *   authenticationTypes = TRUE,
 *   authenticationOptional = TRUE,
 *   dataProvider = {
 *     "entityType": "comment",
 *     "bundles": {
 *       "comment_node_article"
 *     },
 *   },
 *   majorVersion = 1,
 *   minorVersion = 0
 * )
 */
class Comments__1_0 extends ResourceEntity implements ResourceInterface {}
```

Unlike other entities, the comment table doesn't have a column for bundle, If we allow no bundles using `bundles = FALSE`:

``` php
/**
* ...
 *   dataProvider = {
 *     "entityType": "comment",
 *     "bundles": FALSE,
 *   },
 * ...
 */
```

and bypass entity bundle condition for EntityFieldQuery, the comments resource would work.

### DIFF
--- a/modules/restful_example/src/Plugin/resource/DataProvider/DataProviderComment.php
+++ b/modules/restful_example/src/Plugin/resource/DataProvider/DataProviderComment.php
@@ -13,6 +13,27 @@ use Drupal\restful\Plugin\resource\DataProvider\DataProviderInterface;
 class DataProviderComment  extends DataProviderEntity implements DataProviderInterface {
 
   /**
+   * Overrides DataProviderEntity::setPropertyValues().
+   *
+   * Set nid and node type to a comment.
+   */
+  protected function setPropertyValues(\EntityDrupalWrapper $wrapper, $object, $replace = FALSE) {
+    $comment = $wrapper->value();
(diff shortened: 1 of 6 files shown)


e0ipso:
This looks great! It even has an example :clap:

Can we use that example in a test and make sure we can retrieve/create/update/delete a comment? That'd be awesome!

amitaibu:
Maybe we can simply ignore the bundle property if the entity type doesn't have one?

xuxucode:
Comment test is a little tricky, but now it works :-)

e0ipso:
Merged.

Thanks for the collaboration!

PR #747, opened 2015-11-18
[CLEANUP] Friendly error for incompatible BETWEEN ops

e0ipso:
When the BETWEEN operator is used, then 2 values should be provided
otherwise when piping the options along to the data provider there
will be an incomprehensible error. For MySql that means a syntax
error.


e0ipso:
This is green. :-)

PR #748, opened 2015-11-18
[BUGFIX] Remove cache fragments during shutdown

e0ipso:
When an entity is updated or deleted we need to clear caches and
remove the cache fragments. The problem is that depending on the
site configuration, that operation may be affected for a rolling
back transaction. Defer the clearing to the shutdown instead.


e0ipso:
Merging!

amitaibu:
was this merged? (I still see the merge pull request active)

e0ipso:
Yep. Squashing 2 commits messes with GH's detection system :-P

amitaibu:
So after this change, the request is returned to the client, and deleting happens after - right? Were there so many items, that it caused a significant delay?

PR #749, opened 2015-11-19
Fixes #685: Adds UID to an entity upon creation.

### LINKED ISSUE #685: BUG: Entities created by RESTful get the uid property too late.
Using content access, I have a node type (notes) that users are allowed to created/edit/delete _only their own_ nodes.

When RESTful's `createEntity()` method is called, the stub entity is created without the account information, which means that although the user is given access to create the entity, because the entity isn't (yet) owned by the user, no properties/fields can be set.

Allow me to demonstrate this with the following request:

```
POST /api/v1.0/lesson/264/note HTTP/1.1
Access-Token: eUWRUXwyMYIDhabSFApso6NoBHCh4tn88IOmmKeVyAI
Host: restful.local
Connection: close
User-Agent: Paw/2.2.5 (Macintosh; OS X/10.11.0) GCDHTTPRequest
Content-Length: 86

{
    "title": "Test Title",
    "body": "<p>An example note\r\n</p>",
    "lesson": 264
}
```

If I debug the output of `entity_access` ([here](http://www.drupalcontrib.org/api/drupal/contributions!entity!entity.module/function/entity_access/7)), I'll see the following response:

```
object(stdClass)[158]
  public 'type' => string 'notes' (length=5)
  public 'language' => string 'und' (length=3)
  public 'is_new' => boolean true
  public 'status' => int 1
  public 'promote' => int 0
  public 'sticky' => int 0
string 'create' (length=6)

object(stdClass)[158]
  public 'type' => string 'notes' (length=5)
  public 'language' => string 'und' (length=3)
  public 'is_new' => boolean true
  public 'status' => int 1
  public 'promote' => int 0
  public 'sticky' => int 0
  public 'title' => string 'Test Title' (length=15)
string 'update' (length=6)

{"type":"http:\/\/www.w3.org\/Protocols\/rfc2616\/rfc2616-sec10.html#sec10.4.1","title":"Property title cannot be set.","status":400,"detail":"Bad Request."}
```

When RESTful creates the entity [here](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/RestfulEntityBase.php#L633), there is no `uid` property on the element, therefore when the entity_access check is done for permission on the user account to `update` the in-memory entity (158 in my example), entity_access denies the `$op` because it cannot validate that the `$entity` is owned by the user making the request.

I would propose that when a stub entity is created, the `uid` property should be attached before any other access checks happen.

lukewertz:
When an entity is being created, it's possible to get a false-negative on the entity access check if the UID isn't on the entity and the user making the request isn't an admin user.

For example, a user role "Content Contributor" is allowed to create "Blog" nodes and is only allowed to edit their own (not _any_).

When a "Content Contributor" makes the following request:

```
POST api/v1.0/blog
{
    "title": "Test title",
    "body": "<p>An example body\r\n</p>",
}
```

They would receive the following back:

```
{
  "type": "http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.1",
  "title": "Property title cannot be set.",
  "status": 400,
  "detail": "Bad Request."
}
```

If you remove title, then you get the same message for body (or any other public field on the node). The reason for this is because the way that "create" happens in RESTful right now is that the entity is created without the UID (author) on the entity. So the entity is created but isn't attributed to the user making the request, so when the user making the request then tries to _edit_ the entity to set the a value, the permission system denies them access.

I consider this to be a critical bug.

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -634,7 +634,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
 
     // Set the uid as early as possible to avoid false-negatives on permission
     // checks.
-    if (!empty($this->getAccount()->uid) && $this->getAccount()->uid !== 0) {
+    if ('node' == $this->entityType && !empty($this->getAccount()->uid)) {
       $values['uid'] += $this->getAccount()->uid;
     }
 
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -594,7 +594,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
    * {@inheritdoc}
    */
   public function deleteEntity($entity_id) {
-    $this->isValidEntity('update', $entity_id);
+    $this->isValidEntity('delete', $entity_id);
 
     $wrapper = entity_metadata_wrapper($this->entityType, $entity_id);
     $wrapper->delete();
@@ -631,6 +631,12 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
     $entity_info = $this->getEntityInfo();
     $bundle_key = $entity_info['entity keys']['bundle'];
     $values = $bundle_key ? array($bundle_key => $this->bundle) : array();
+
+    // Set the uid as early as possible to avoid false-negatives on permission
+    // checks.
+    if (!empty($this->getAccount()->uid) && $this->getAccount()->uid !== 0) {
+      $values['uid'] += $this->getAccount()->uid;
+    }
 
     $entity = entity_create($this->entityType, $values);
 
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -632,12 +632,6 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
     $bundle_key = $entity_info['entity keys']['bundle'];
     $values = $bundle_key ? array($bundle_key => $this->bundle) : array();
 
-    // Set the uid as early as possible to avoid false-negatives on permission
-    // checks.
-    if ('node' == $this->entityType && !empty($this->getAccount()->uid)) {
-      $values['uid'] += $this->getAccount()->uid;
-    }
-
     $entity = entity_create($this->entityType, $values);
 
     if ($this->checkEntityAccess('create', $this->entityType, $entity) === FALSE) {
@@ -648,7 +642,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
 
     $wrapper = entity_metadata_wrapper($this->entityType, $entity);
 
-    $this->setPropertyValues($wrapper);
+    $this->setPropertyValues($wrapper, FALSE, 'create');
     return array($this->viewEntity($wrapper->getIdentifier()));
   }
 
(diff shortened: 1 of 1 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -632,6 +632,12 @@ public function createEntity() {
     $bundle_key = $entity_info['entity keys']['bundle'];
     $values = $bundle_key ? array($bundle_key => $this->bundle) : array();
 
+    // Set the uid as early as possible to avoid false-negatives on permission
+    // checks.
+    if (!empty($this->getAccount()->uid) && $this->getAccount()->uid !== 0) {
+      $values['uid'] += $this->getAccount()->uid;
```
e0ipso: Can we be sure that all entities have the `uid` key as their author property?

lukewertz: Ahh, good point. Certainly not. I was only thinking about node entities.

lukewertz: Given that, the fix for this might be to restructure the way entities are created. Perhaps we have to build the whole entity in memory before checking permissions on it. That seems pretty messy given field-level permissions, though.

lukewertz:
The proposed fix right now, allowing an operation type to be declared on `setPropertyValues` is going to take some refactoring. I'm happy to do it, but would like some input before I do to know if I'm headed in the right direction with this type of fix.

amitaibu:
This is weird, because we POST entities often, and never encountered this scenario. Can you reproduce on a clean installation?

lukewertz:
I'm not entirely sure when I'll get a chance to do a test on a completely clean install--it won't be this week. It wouldn't surprise me though if you didn't encounter this particular configuration in your testing. It would only happen if the user making the request had permission to create entities but they didn't have permission to "edit any" of the same entity. Unless you configured a test specifically for that condition, you wouldn't encounter it.

amitaibu:
> but they didn't have permission to "edit any" of the same entity

Yes, I did that exact same scenario just yesterday :)

amitaibu:
(and it works fine for me)

lukewertz:
Ahh, very interesting, @amitaibu. I'll build a fresh copy and test again.

The project I'm working on is using three different access controls: core's, Access Control, and Organic Group. It's quite possible that I have a mis-configuration in my system.

I also have many dozens of behat tests to check access controls throughout the system, which is how I caught this use case.

A clean install will tell me much.

Thanks!

xuxucode:
Did you apply this patch: [Prevent notice in entity_metadata_no_hook_node_access() when node is not saved](https://drupal.org/node/2086225#comment-8768373)?

e0ipso:
@lukewertz did you have a chance to check @edwardaa's suggestion?

lukewertz:
@e0ipso I'm working on testing it this morning (GMT-4). I should know today if it works or not.

Regardless, though, I'm wondering if we should look at fixing this issue within RESTful rather than in that patch since the patch has been open for so long without any indication that it is going to be merged, despite being ready to be merged.

e0ipso:
I am sure @fago is extremely busy with lots of different stuff. I'd love to have that merged, since it can benefit many other projects using Entity API.

> I'm wondering if we should look at fixing this issue within RESTful

That's a valid point, but I'm not 100% sure we can do that for all of the use cases. Let's look into that though!

> I'm working on testing it this morning (GMT-4). I should know today if it works or not.

That's fantastic news! Thanks a lot for your dedication. GTM-4… where are you based? Is that Quebec's timezone?

lukewertz:
@edwardaa & @e0ipso: Thank you both so much for your help and input. The patch that you linked to, @edwardaa, did the trick for me.

I'd still be open to doing the slight refactoring needed (similar to [this](https://github.com/RESTful-Drupal/restful/pull/749/files#diff-3c241a6c3b4e93a88a7da3e91c3e9358R664)) in any class that implements `setPropertyValues`. By allowing `setPropertyValues` to pass on the `$op`, it'd be pretty easy to fix this without relying on the patch. In my testing, this method worked well.

@e0ipso: GMT-4 aka, US Eastern time. I'm in Cincinnati, OH (I work remotely for Chicago-based [Palantir.net](https://palantir.net)).

e0ipso:
> Thank you both so much for your help and input. The patch that you linked to, @edwardaa, did the trick for me.

Closing per above's comment.

Issue #744, opened 2015-11-15
Anonymous POST does not respect node creation permission


obar:
I'm using RESTful 7.x-2.0, with a `ResourceNode` for a custom node type that allows anonymous users to create new content. With a POST to this resource to a new object, the response message is: "You do not have access to create a new resource."

When the permission "Bypass content access control" is enabled for anonymous users, the same POST succeeds and creates a new object, but naturally one cannot do this in production.

Since what I need from this API is incredibly simple, with only one resource that I want the ability to POST to, my work-around was to comment out the throwing of the following two exceptions in `restful/src/Plugin/resource/DataProvider/DataProviderEntity.php`:
     `throw new ForbiddenException('You do not have access to create a new resource.');`
     `throw new BadRequestException(format_string('Property @name cannot be set.', array('@name' => $public_field_name)));`

With the work-around, the permissions checks preceeding these exceptions still fail, but without the exception thrown the content is created and properties get set. Perhaps I'm missing some other permission here to make this work as expected?

e0ipso:
You should give the anonymous user the permission to create that particular content type. RESTful does not define special access permissions, but works with the content permissions (and field permissions).

e0ipso:
For instance set `create article content` for anon users.

e0ipso:
Check this test for more info: https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/tests/RestfulCreateNodeTestCase.test#L37

obar:
Sorry that wasn't clear in the first sentence, but anonymous users did have the "[content type]: create new content" permission set.

lukewertz:
If the anonymous user has permission to create content, but not "edit any" this this is the consistent with behavior discussed in bug https://github.com/RESTful-Drupal/restful/issues/685.

One way around this is to give Anonymous access to "edit any" but that seems like a really, really bad idea :)

I'm not sure that my fix for the bug would capture this use case of anonymous having access to update the entity once the shell entity is created. That's a different use case than what I'm trying to accomplish in  https://github.com/RESTful-Drupal/restful/pull/749

ademarco:
I can reproduce that, it seems to be due to `Drupal\restful\Plugin\resource\DataProvider\DataProviderEntity::setPropertyValues()` checking for `edit` operation when invoking `checkPropertyAccess()`:

``` php
      $entity_property_access = $this::checkPropertyAccess($resource_field, 'edit', $interpreter);
[...]

PR #754, opened 2015-11-23
Choose appropriate operation when checking property access #744.

ademarco:
Pull request for #744.

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -875,6 +875,7 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
     }
     $save = FALSE;
     $original_object = $object;
+    $op = $wrapper->getIdentifier() ? 'edit' : 'create';
     $interpreter = new DataInterpreterEMW($this->getAccount(), $wrapper);
     // Keeps a list of the fields that have been set.
     $processed_fields = array();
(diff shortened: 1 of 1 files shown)


e0ipso:
Is it possible to reproduce the bug in a test? That would be adding on top of your fantastic job analyzing this bug.

Maybe in `RestfulCreateNodeTestCase`.

ademarco:
Ok sure, I'll give it a try!

e0ipso:
Do you need any help on this? Do we finally need this, or the Entity API patch does the trick?

ademarco:
Sorry I got caught-up with some deadline at the moment, I'll have time to
test is properly next as from next week.

On Tue, Dec 1, 2015 at 1:31 AM, Mateu Aguiló Bosch <notifications@github.com

e0ipso:
That is fantastic news!

e0ipso:
@ademarco any news on this front?

ademarco:
Not yet, sorry, haven't got time as I thought I would, still I plan to
investigate it.

e0ipso:
@ademarco I'm unsure about the state of this PR. Do we need this? Should this be fixed by the Entity API patch?

laradevitt:
With this PR applied, because of [these lines in the ResourceFieldEntity class](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/resource/Field/ResourceFieldEntity.php#L422-L425)...

```
    if ($op == 'edit' && empty($info['setter callback'])) {
      // Property does not allow setting.
      return FALSE;
    }

```

... a record will be created even though there are fields for which there is no setter callback defined (the values will be empty). 

I have no idea if this would be preferred behaviour over not permitting the entity to be created at all.

omarlopesino:
@e0ipso i think to prevent this warning it's better correct it in restfu. Patch provided in https://www.drupal.org/node/2086225 issue tries to simulate 'create' when you call entity access in 'update' mode to prevent the error. But if you try check property access in create mode you won't have that error, and you will continue checking entity access.

And, if you try check edit mode for a property when you are creating a entity (a node by example), you are skipping create access, which user maybe don't have (imagine a user story when he have edit access but doesn't have create access). This is something that can occurs.

What do you think?

PR #756, opened 2015-11-23
[BUGFIX] Pass in the interpreter to the ResourceField::value

e0ipso:
Failing to do so results in a PHP error with the type hinting.


e0ipso:
Merged.

PR #770, opened 2015-12-01
[BUGFIX] Better support for forbidden entities

e0ipso:
When requesting a forbidden entity a 403 should arise. When requesting a forbidden entity as part of a list, it should be ignored and reported as denied in the metadata.


amitaibu:
Please don't merge yet .

e0ipso:
Cool.

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -822,11 +829,13 @@ protected function isValidEntity($op, $entity_id) {
         // Just return FALSE, without an exception, for example when a list of
         // entities is requested, and we don't want to fail all the list because
         // of a single item without access.
-
         // Add the inaccessible item to the metadata to fix the record count in
         // the formatter.
         $inaccessible_records = $this->getMetadata()->get('inaccessible_records');
-        $inaccessible_records[] = $entity_id;
+        $inaccessible_records[] = array(
```
amitaibu, line 835: I think this is a bit problematic. Not a security issue, but I think a bad practice (which is following core's bad practice):

In Drupal core when non-authorised user visits a private node they get a 403. IMO it should be 404 (which btw is the case in Drupal commerce's entities as far as I remember). 

While it's not a security per-se, there could be information disclosure - I could hand craft my url to see if content exists. So if `company/nike` is accessible by me, I could check if my competitor is also there by visiting `company/puma`. With RESTful since one can query by and field, I could search all the entities that have the word `puma` in them.

In short: 
- I think that when visiting a private entity - one should get a 404 exception (or make it configurable)
- In listing silently ignore the private entities, and don't add their info in the metadata.

@e0ipso thoughts? :)

e0ipso, line 835: I see your point. I believe that we need a setting controlling this behavior:
- [ ] Hide access denied (turns 403 into 404 when selecting entities with access denied)

> I think that when visiting a private entity - one should get a 404 exception (or make it configurable)

I like the idea of making it configurable. I believe that you can do something similar with Drupal pages. We would have a new `UnaccessibleRecordException` and `UnaccessibleRecordException::getCode()` will read from a variable.

> In listing silently ignore the private entities, and don't add their info in the metadata.

I'd make this configurable too, since in cases were the info disclosure is not a concern that is very helpful data. It tells me: «Hey, I know I told you these entities were there but now they are gone all of the sudden. They are gone b/c you don't have access to them.»

e0ipso:
@amitaibu thoughts?

src/Plugin/resource/DataProvider/DataProviderEntity.php
amitaibu, line 835: I like the ideas, of making both configurable. Would you agree the default setting should be the strict ones (404 + no metadata)?

e0ipso:
Merged.

Thanks!

Issue #772, opened 2015-12-02
Easy way to alter publicFieldsInfo by other modules


RoySegall:
Iv'e been working on a [module](https://github.com/RoySegall/apps_entity_restrictions) which define third party apps. When creating an application you can select which properties and which field it can access. I want to create a sub module which take care for the user about restrictions and i'm looking for an easy way to alter the public fields info for each end point.

The module would need to check the access for the for each property/field and set the access property to true or false. The easiest way I came up with was to add `drupal_alter` in [RestfulBase ::getPublicFields()](https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/plugins/restful/RestfulBase.php#L1110).

What do you think?

e0ipso:
You can add access calbacks to each of the public fields. Would that work?

RoySegall:
Sure but then I need to prevent from properties to be presented.

e0ipso:
Not sure what you mean, but you have `access callbacks` in RESTful for the field definitions. At least for 2.x (I don't remember about the old version).

RoySegall:
Sure, but I need the module to iterate over each plugin definition and apply it's logic business.

e0ipso:
Does that mean that you cannot put your `drupal_alter` in the access callbacks?

RoySegall:
I want to apply to `drupal_alter` to any endpoint.

RoySegall:
That mean - any end point defined in the system, entities and non-entities, will be handled by my sub-module.

PR #773, opened 2015-12-03
[WIP] Allow other modules to alter the plugin public fields info.

RoySegall:
#772

### DIFF
--- a/modules/restful_example/plugins/restful/node/articles/2.1/articles__2_1.inc
+++ b/modules/restful_example/plugins/restful/node/articles/2.1/articles__2_1.inc
@@ -1,0 +1,16 @@
+<?php
+
+$plugin = array(
+  'label' => t('Articles'),
+  'resource' => 'articles',
+  'name' => 'articles__2_1',
+  'entity_type' => 'node',
+  'bundle' => 'article',
+  'description' => t('Example for a plugin with altered public fields.'),
+  'class' => 'RestfulExampleArticlesResource__1_0',
+  'authentication_types' => TRUE,
+  'authentication_optional' => TRUE,
+  'major_version' => 1,
+  'minor_version' => 0,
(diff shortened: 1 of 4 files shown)


modules/restful_example/plugins/restful/node/articles/2.1/articles__2_1.inc
```diff
@@ -0,0 +1,16 @@
...
+  'name' => 'articles__2_1',
+  'entity_type' => 'node',
+  'bundle' => 'article',
+  'description' => t('Example for a plugin with altered public fields.'),
+  'class' => 'RestfulExampleArticlesResource__1_0',
+  'authentication_types' => TRUE,
+  'authentication_optional' => TRUE,
+  'major_version' => 1,
+  'minor_version' => 0,
+  'allow_public_alter' => FALSE,
```
RoySegall, line 15: => `allow_public_fields_alter`

e0ipso:
I am not a huge fan of this. I still fail to see the need for this.

Why can't custom access callbacks in one hand, and inheritance in another help with this?

RoySegall:
I'm trying to figure out a way for people to do as less they can by code. This mean that all the developer would need to do is to allow fields info alter and my module will handle the `access callbacks` property for them.

The module originally created to set up a backend similar to Facebook, and Twitter with the feel of third side applications, single sign on and reports for application usage: how many time the application succeeded with api access, how many times it's failed and in the far future I will add the option to notify people on this failures.

amitaibu:
@RoySegall I think that you can solve this with a few parent classes (extending from Base, NodeBase, etc') that will have your logic.

The `drupal_alter()` overhead will indeed be too big for the value.

RoySegall:
Closing this one. I extend and improve the API for module but I found out another problem. I'll open a follow up issue.

PR #779, opened 2015-12-07
Update README.md

jeff-h:
Make optional Entity Validator module version dependency more explicit.


e0ipso:
Thanks! This is merged.

PR #781, opened 2015-12-07
[FEATURE] Allow a generic field reference

e0ipso:
This includes adding support for arbitrary fields that return an ID
to be used as a reference as long as they include the corresponding
'resource' information.

### DIFF
--- a/tests/modules/restful_test/src/Plugin/resource/entity_test/main/v1/Main__1_8.php
+++ b/tests/modules/restful_test/src/Plugin/resource/entity_test/main/v1/Main__1_8.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\restful_test\Plugin\resource\entity_test\main\v1\Main__1_8.
+ */
+
+namespace Drupal\restful_test\Plugin\resource\entity_test\main\v1;
+
+use Drupal\restful\Plugin\resource\DataInterpreter\DataInterpreterInterface;
+
+/**
+ * Class Main__1_8.
+ *
+ * @package Drupal\restful_test\Plugin\resource
+ *
+ * @Resource(
+ *   name = "main:1.8",
+ *   resource = "main",
+ *   label = "Main",
+ *   description = "Export the entity test 'main' bundle.",
+ *   authenticationOptional = TRUE,
+ *   dataProvider = {
+ *     "entityType": "entity_test",
+ *     "bundles": {
+ *       "main"
+ *     },
+ *   },
+ *   majorVersion = 1,
+ *   minorVersion = 8
+ * )
+ */
+class Main__1_8 extends Main__1_0 {
+
+  /**
+   * Overrides ResourceEntity::publicFields().
+   */
+  protected function publicFields() {
+    $public_fields = parent::publicFields();
+
+    $class = get_class($this);
+    $public_fields['random_rel'] = array(
+      'callback' => $class . '::randomRelationship',
+      'class' => '\Drupal\restful\Plugin\resource\Field\ResourceFieldReference',
+      'resource' => array(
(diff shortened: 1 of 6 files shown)


amitaibu:
Can you please give an example how and when it's used

e0ipso:
@amitaibu let me write a test, that should clarify stuff.

amitaibu:
:+1:

e0ipso:
@amitaibu hopefully that makes more sense now.

tests/modules/restful_test/src/Plugin/resource/entity_test/main/v1/Main__1_8.php
```diff
@@ -0,0 +1,84 @@
...
+class Main__1_8 extends Main__1_0 {
+
+  /**
+   * Overrides ResourceEntity::publicFields().
+   */
+  protected function publicFields() {
+    $public_fields = parent::publicFields();
+
+    $class = get_class($this);
+    $public_fields['random_rel'] = array(
```
amitaibu, line 42: random is probably a misleading name. arbitrary? or `property_reference`?

Anyway, if I understand correctly this could be used for example on `$node->uid` to convert the response to a full blown user resource?

e0ipso, line 42: Yeah. My use case is: «Expose the redirect table, explode the path (node/1234) and add a reference to the TV Show node»

amitaibu, line 42: I think the $node->uid would be a better example - because the existing (advanced) example IMO is a bit too advanced.

e0ipso, line 42: The thing is that for `$node->uid` you _could_ use this method. But you are going to be better off with something already existing like:

``` php
$public_fields['author'] = array(
  'property' => 'uid',
  'class' => '\Drupal\restful\Plugin\resource\Field\ResourceFieldEntityReference',
  'resource' => array(
    'name' => 'users',
    'majorVersion' => 1,
    'minorVersion' => 0,
  ),
);
```

That is very similar, but it assumes that the result of the field is an ID of an entity so it will try to load the entity if needed (and keep track of the entity type, …).

amitaibu, line 42: :+1: in that case worth improving the docs to explain the use case.

e0ipso, line 42: https://github.com/RESTful-Drupal/restful/pull/781/files#diff-5d67d48619d69128ec21c42939014e14R13

amitaibu, line 42: yeah, when I saw that doc I thought it was like $node->uid (as it's a property, not an entity reference field). I think it's just needs mentioning that the resource could be a non-entity.

amitaibu, line 42: anyway, feature is cool :)

e0ipso, line 42: I'll push a clear documentation change, and then merge with your blessing.

amitaibu, line 42: you have my blessing. Always! :)

e0ipso, line 42: Docs have been expanded to capture this information.

e0ipso:
This was merged.

PR #783, opened 2015-12-09
Update DataProviderEntity.php

jeff-h:
A view op may have originated as part of a create operation. In that case, we don't throw an InaccessibleRecordException, as the create may succeed with or without the user having access to view the new item.

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -216,6 +216,7 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
     // The access calls use the request method. Fake the view to be a GET.
     $old_request = $this->getRequest();
     $this->getRequest()->setMethod(RequestInterface::METHOD_GET);
+    $this->getRequest()->setViaRouter(FALSE);
     $output = array($this->view($wrapper->getIdentifier()));
     // Put the original request back to a POST.
     $this->request = $old_request;
@@ -233,7 +234,7 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
       // This view may have been requested as part of a create operation. In
       // that case, we don't throw an InaccessibleRecordException, as the create
       // may succeed with or without the user having access to view the new item.
-      if (!$this->getResourcePath()) {
+      if (!$this->getRequest()->isViaRouter()) {
         return array();
       }
       throw new InaccessibleRecordException(sprintf('The current user cannot access entity "%s".', $entity_id));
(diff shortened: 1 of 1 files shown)


src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -230,6 +230,12 @@ public function view($identifier) {
     $entity_id = $this->getEntityIdByFieldId($identifier);
 
     if (!$this->isValidEntity('view', $entity_id)) {
+      // This view may have been requested as part of a create operation. In
+      // that case, we don't throw an InaccessibleRecordException, as the create
+      // may succeed with or without the user having access to view the new item.
+      if (!$this->getResourcePath()) {
```
e0ipso: We could actually check for the presence of a write method. Something like `$this->getRequest()->isWriteMethod()` (or whatever the actual method is).

jeff-h:
Unfortunately I can't use `$this->getRequest()->isWriteMethod()` because the method at this point is in fact a GET. It's due to a clever bit of code that converts the request temporarily to a GET:

https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/resource/DataProvider/DataProviderEntity.php#L216-L221

I am actually very uncertain about using `getResourcePath()` here. Would it be appropriate to use `$request->isViaRouter()`, setting this "fake" GET request's `viaRouter` to `false` earlier where the request is modified to be a GET?

e0ipso:
Using `viaRouter` is a good way to go. I think that `viaRouter` should populate automatically.

jeff-h:
Because the request in question is a clone of the original GET request, its `viaRouter` is `TRUE`. I've manually set it to `FALSE` in the clone and this solution is working for me.

jeff-h:
Hmm, is that test failure actually due to my change? Any way I can rerun a test, or can only you guys do that?

e0ipso:
I'm torn by this new feature. I would expect to get an error if I don't have access to view the entity. A 403 error would show that the entity was created as expected and there is no access to it.

amitaibu:
> I would expect to get an error if I don't have access to view the entity.

I agree, I think it shouldn't be handled by RESTful, but by the implementing module.

jeff-h:
> I would expect to get an error if I don't have access to view the entity. A 403 error would show that the entity was created as expected and there is no access to it.

I think you've convinced me. The problem was my immediate reaction to anything other than 200! I'll document this in the wiki so others realise more quickly that a 403 is fine in this specific edge-case.

I wonder if we should still add the following line at the point where the request is faked to be a `GET`. It's not needed right now, but might be more correct? Happy to revise the PR to suit.

``` php
    $this->getRequest()->setViaRouter(FALSE);
```

e0ipso:
Thanks for the documentation!

Closing.

PR #786, opened 2015-12-10
Update DataProviderDbQuery.php

jeff-h:
Data providers' `update()` method must return an array. The returned value from this method is ultimately passed via `$page_callback_result` to `drupal_deliver_page()` which expects it the value to be:

``` php
 * @param $page_callback_result
 *   The result of a page callback. Can be one of:
 *   - NULL: to indicate no content.
 *   - An integer menu status constant: to indicate an error condition.
 *   - A string of HTML content.
 *   - A renderable array of content.
```

Without this pull request, the value returned will be an object.

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderDbQuery.php
+++ b/src/Plugin/resource/DataProvider/DataProviderDbQuery.php
@@ -331,7 +331,7 @@ class DataProviderDbQuery extends DataProvider implements DataProviderDbQueryInt
     if (!drupal_write_record($this->getTableName(), $record, $id_columns)) {
       throw new ServiceUnavailableException('Record could not be updated to the database.');
     }
-    return $this->view($identifier);
+    return array($this->view($identifier));
   }
 
   /**


e0ipso:
Hmmm, isn't the result of `update()` going through the formatter to unravel it?

jeff-h:
The formatter is invoked inside `restful.module` :: `restful_delivery()`, but only after that function has made assumptions that it has received an array.

277: `function restful_delivery($var = NULL) {`
`$var` is received (ultimately) from `DataProviderDbQuery->update()` and without this PR it's an object

We crash and burn on the following line because we're attempting to access an object as an array:
284: `if (!empty($var['status'])) {`

It's not until line 318 that we involve the formatter:
318: `// Get the formatter for the current resource.`

There's a couple of precedents where `create()` and `update()` convert their results to an array, in `DataProviderVariable` from the example module:
- https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/modules/restful_example/src/Plugin/resource/variables/DataProviderVariable.php#L63
- https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/modules/restful_example/src/Plugin/resource/variables/DataProviderVariable.php#L132

e0ipso:
Oh! Now I get what you meant. Sorry I wasn't reading properly the other day.

This is merged now! Thanks!

PR #787, opened 2015-12-10
Return NULL if the embedded entity does not exist

xuxucode:
The embedded entity may be deleted, if in that case return NULL for the embedded entity, otherwise the present request won't work just due to a nonexistent embedded entity.

### DIFF
--- a/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -336,6 +336,10 @@ class ResourceFieldEntity implements ResourceFieldEntityInterface {
         // access to the property.
         return NULL;
       }
+      catch (UnprocessableEntityException $e) {
+        // If you access a nonexistent embedded entity.
+        return NULL;
+      }
       // Test if the $embedded_entity meets the filter or not.
       if (empty($parsed_input['filter'])) {
         return $embedded_entity;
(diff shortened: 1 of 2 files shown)


e0ipso:
I think that the underlying bug is embedding/requesting a deleted entity. How did you reproduce this bug?

xuxucode:
For example [Artilces__1_5.php](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/modules/restful_example/src/Plugin/resource/node/article/v1/Articles__1_5.php#L78-L88):

``` php
$public_fields['user'] = array(
  'property' => 'author',
  'resource' => array(
    'name' => 'users',
    'fullView' => TRUE,
    'majorVersion' => 1,
    'minorVersion' => 0,
  ),
);
```

If one of the articles belongs to a deleted user with uid 100, the request to `api/v1.5/articles` will result in

``` json
{
  "type": "http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html",
  "title": "The entity ID 100 does not exist.",
  "status": 422,
  "detail": "Unprocessable Entity; Validation errors."
}
```

expected result:

``` json
{
  "data": [
    ...
    {
      "id": 9,
      "label": "example",
      "self": "http://www.example.com/v1.5/articles/9",
      "user": null
    }
  ],
  "count": 10,
  "self": {
    "title": "Self",
    "href": "http://www.example.com/v1.5/articles"
  }
}
```

Note the user property in above result is set to **NULL**, since the user is deleted.

e0ipso:
Oh! It makes so much more sense now. Thanks for the extra explanation.

e0ipso:
This has been merged. Thanks!!!

PR #788, opened 2015-12-10
DataProvider: throw 422 instead of 503 when invalid urlParams are present

jeff-h:
Throw 422 not 503 when invalid urlParams are present.


jeff-h:
`503 "Service Unavailable"` seemed weird for this; it kind of implies that the server isn't responding properly and is going to come good shortly.

The general consensus as far as I could see is that 422 is the most appropriate for a request that had valid syntax but un-processable instructions.

e0ipso:
Thank you for this change!

PR #752, opened 2015-11-20
Log exception message in restful_menu_access_callback() #751

ademarco:

PR #794, opened 2015-12-13
[FEATURE] More semantic error codes on access check

e0ipso:
If a request is denied access via a RESTful exception do not discard the error code. Instead return a well formatted error response.

### DIFF
--- a/tests/RestfulRenderCacheTestCase.test
+++ b/tests/RestfulRenderCacheTestCase.test
@@ -9,8 +9,10 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Drupal\restful\Http\Request;
 use Drupal\restful\Http\RequestInterface;
 use Drupal\restful\RenderCache\RenderCache;
-use Drupal\restful\RenderCache\Entity\CacheFragmentController;
 
+/**
+ * Class RestfulRenderCacheTestCase.
+ */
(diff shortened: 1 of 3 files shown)


e0ipso:
This PR supersedes #752.

tests/RestfulRenderCacheTestCase.test
```diff
@@ -245,7 +247,7 @@ class RestfulRenderCacheTestCase extends \RestfulCurlBaseTestCase {
     // Try to access the cached resource.
     $response = $this->httpRequest($path);
     // The user should get a 403.
-    $this->assertEqual($response['code'], 403, 'Access denied for anonymous user.');
+    $this->assertEqual($response['code'], 401, 'Access denied for anonymous user.');
```
amitaibu, line 250: why 401? Also the comment above still says 403

e0ipso, line 250: Yikes. I'll fix the comment.

401 means authorization error, meaning that the requester meant to be identified with a Drupal user but the credentials were bad.

403 is more for, _I know who you are, and you cannot do that_.

amitaibu, line 250: sure, just the comment made me think it was wrong :)

e0ipso:
This was merged.

Thanks for the effort!

ademarco:
Great! Thanks!

PR #790, opened 2015-12-14
7.x-2.x - restful_cache_fragment grows at an enormous rate

e0ipso:
I've created a custom resource for a custom entity. This entity has a relationship to a node that represents editable content. When exposing this entity and it's node relationship via a custom resource, the restful_cache_fragment table is populated with a little over 10,000 entries (with a range of 100). While this provides great performance on subsequent requests; it makes flushing the cache or rebuilding the registry (drush rr) quite difficult without manually truncating the table. Is creating this many entries expected behavior?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,10 @@ install:
   # Make sure we don't fail when checking out projects
   - echo -e "Host github.com\n\tStrictHostKeyChecking no\n" >> ~/.ssh/config
   - sudo cat /etc/apt/sources.list
-  - echo "deb http://us-central1.gce.archive.ubuntu.com/ubuntu/ precise multiverse" | sudo tee -a /etc/apt/sources.list
-  - echo "deb-src http://us-central1.gce.archive.ubuntu.com/ubuntu/ precise multiverse" | sudo tee -a /etc/apt/sources.list
-  - echo "deb http://us-central1.gce.archive.ubuntu.com/ubuntu/ precise-updates multiverse" | sudo tee -a /etc/apt/sources.list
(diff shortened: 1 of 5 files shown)


e0ipso:
This does look like a performance concern. I will look into this.

e0ipso:
@ethanhinson can you please re-test with the code in this PR?

e0ipso:
That way we'll see if there is a performance improvement.

e0ipso:
This is an isolated test with 7500 cache fragment entities (it could get even worse).



The cache clear is sped up considerably and I don't think we're missing anything by not executing delete hooks. Besides we're making it configurable to play nice with the entity system if needed.

@amitaibu do you have any thoughts?

e0ipso:
I'm going to merge this. We can follow up later.

ethanhinson:
@e0ipso 

Quick tests on my local show _much_ improved cache flushes with 100k+ entries. Thanks for the refactor!

amitaibu:
Makes sense. I wonder if there's a real use case for not wanting the fast clear cache -- when do you think this hook_entity_delete() will be needed?

e0ipso:
> when do you think this hook_entity_delete() will be needed?

Never. Just allowing drupalisms via config. It would be like making the entity type fieldable, you can… but it's not recommended.

amitaibu:
Then IMO it's ok to keep the variable for that super rare use case, but maybe remove the UI option?

Another option which I prefer is just to simplify the code and have only the fast option. For the rare case - devs will override the class (which IMO will never really happen).

e0ipso:
Not firing the hooks for an entity is a weird thing. If we don't do it, we should definitely document it. I think that a checkbox in the config page is a very good mean of documentation.

amitaibu:
Fair enough

PR #796, opened 2015-12-14
[FEATURE] Performance improvements

e0ipso:
Add minor improvements to methods that are called a lot. A small
bit can add up when called tens of thousands of times.


e0ipso:
Some numbers that may mean nothing…

e0ipso:
Merged.

PR #799, opened 2015-12-16
Allow OPTIONS calls to access_token.

jeff-h:
Without this, a preflight `OPTIONS` call will return either a 
- `401 Unauthorized` if you request without auth (like a browser will do)
- `400 Bad Request` if you do provide auth


e0ipso:
Looks good!

e0ipso:
Merged! Thanks.

PR #802, opened 2015-12-17
[FEATURE] Detach render cache clear from cache flushes

e0ipso:
Render caches are robust enough to not depend on TTLs or other time
based processes.


e0ipso:
Merging.

PR #803, opened 2015-12-17
[BUGFIX] Avoid warning for invalid filters

e0ipso:
When an invalid filter is provided, throw the appropriate error and
avoid additional warnings.


e0ipso:
Merged

PR #805, opened 2015-12-19
Replace drupal_static(__FUNCTION__) with drupal_static(__METHOD__)

xuxucode:
Using `__FUNCTION__` can result in possible collisions in caches when method names overlap.


e0ipso:
Yup! :shipit:

e0ipso:
Merged.

PR #809, opened 2016-01-05
Update Users__1_0.php

pferlito:
Class name should be Users not Articles.


pferlito:
Realized this should be handled as a doc bug.

PR #812, opened 2016-01-07
Multiple auth types ignored if unsuccessful cookie sent

jeff-h:
If a resource specifies `authenticationTypes = TRUE` you would expect it to try each possible authentication provider until one of them successfully authenticates. However, if you supply an invalid cookie with the request (accidentally, in my case), as well as an `access_token`, the latter is never used.

The auth managers are attempted at https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Authentication/AuthenticationManager.php#L97-L103 in the following order: 
- basic auth
- cookie
- token

The cookie auth manager _always_ returns a `$account` even if it's for user 0. Therefore the token auth is never attempted.

You could argue that it's failing auth because I sent a cookie that fails auth, and you'd be right :D Personally, I would have thought it would keep trying, and the docs as they stand certainly imply this. If your stance is the current behaviour is correct, I should update the docs to make it clear that if you provide more than one means of authentication with the request, the auth providers will be checked in an arbitrary (albeit probably consistent) order, and the first auth data checked will end the authentication process whether successful or not.

### DIFF
--- a/src/Plugin/authentication/CookieAuthentication.php
+++ b/src/Plugin/authentication/CookieAuthentication.php
@@ -33,6 +33,10 @@ class CookieAuthentication extends Authentication {
 
     global $user;
     $account = user_load($user->uid);
+    
+    if ($account->uid === 0) {
+      return FALSE;
+    }
 
     if (!$request::isWriteMethod($request->getMethod()) || $request->getApplicationData('rest_call')) {
       // Request is done via API not CURL, or not a write operation, so we don't


e0ipso:
Even though I changed the code of this PR to make it more generic, I am still attributing this to @jeff-h for two reasons:
- He spotted the problem in the first place, and put up a PR.
- He deserves much more than a simple commit credit for all his work on the documentation.

https://github.com/RESTful-Drupal/restful/commit/54d285f083563faef6a56818454bdd560352b490

jeff-h:
Wow, thanks! You guys are awesome :)

jeff-h:
Hey... I'm back :D I've moved back into the server side of my project today and this issue has come back to bite me again... sorry if it's just because my mind has turned to RESTful mush since January :)

I'm trying to do a POST request, using token auth, but I keep getting the BadRequestException `No CSRF token passed in the HTTP header`.

I think it's because I'm doing a write op, whereas back in Jan I was only doing reads. I checked the `RestfulTokenAuthenticationTestCase` and indeed it doesn't test doing a write op with token auth, only reading.

Sadly my test writing skills aren't up to adding this test so I'm rather hoping you might get a chance to look into it at some point. I am expecting the test to fail with the current code, although surely someone's tried to do a POST or PUT using token auth before now?

e0ipso:
For write operations it's not enough to identify yourself, for security reasons you need to get a CSRF token. Did you try getting one? There should be documentation on how to do that.

jeff-h:
Actually, I was supplying an access_token and expecting that to be used. What happened though was when I logged into Drupal the cookie got sent along with the request and the cookie successfully authenticated my RESTful request, but the CSRF token was missing because I didn't actually want to use cookie auth.

It was all because my resource specified `authenticationTypes => TRUE` when I had no intention of using cookie auth. I've written up a note of caution about this at https://github.com/RESTful-Drupal/restful/wiki/1.2.-Defining-a-resource-plugin-(2.x)#a-word-of-caution

PR #816, opened 2016-01-11
Add parsing of Content Type and Length HTTP headers for PHP CGI/FPM.

mcdruid:
When PHP is running as CGI/FPM as opposed to apache mod_php, apache_request_headers() does not exist and HTTP headers have to be parsed from the $_SERVER superglobal.

Content-Type and Content-Length do not appear in $_SERVER with the HTTP_ prefix though, so a specific check is required for them.

This was addressed by symfony in https://github.com/symfony/symfony/issues/1234

### DIFF
--- a/src/Http/Request.php
+++ b/src/Http/Request.php
@@ -369,11 +369,12 @@ class Request implements RequestInterface {
       $headers = apache_request_headers();
     }
     else {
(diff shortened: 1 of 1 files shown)


e0ipso:
:+1:

e0ipso:
This was merged.

Issue #819, opened 2016-01-12
Token authentication in 2.x doesn't recognize http access token headers


lukewertz:
It seems that some browsers pass the access token key as a hyphen-separated string rather than an underscore-separated string. The 1.x branch had code to check for this: https://github.com/RESTful-Drupal/restful/blob/7.x-1.x/modules/restful_token_auth/plugins/authentication/RestfulAuthenticationToken.class.php#L22

PR coming momentarily to add similar logic to 2.x.

e0ipso:
Merged.

```
[FEATURE] Give access-token a new chance

Add support to access-token as a header name, instead of
access_token via code. Even if the key name can be changed in the
annotation, this is common enough to deserve some babysitting.
```

e0ipso:
FWIY you can change the name of the header name in the token resource configuration and set it to hyphenated if you need to.

lukewertz:
Thanks so much for your help with!

PR #820, opened 2016-01-12
Refs #819, Fixes hyphen/dash mismatch in access token.

lukewertz:
See #819.

### DIFF
--- a/modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
+++ b/modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
@@ -89,6 +89,13 @@ class TokenAuthentication extends Authentication {
 
     // Access token may be on the request, or in the headers.
     $input = $request->getParsedInput();
+
+    // If we don't have an $key_name on either the URL or the in the headers,
+    // then check again using a hyphen instead of an underscore.
+    if (empty($input[$key_name]) && empty($request->getHeaders()->get($key_name)->getValueString())) {
+      $key_name = str_replace('_', '-', $key_name);
+    }
+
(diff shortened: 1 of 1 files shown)


e0ipso:
Code looks legit. Can we have test coverage for that?

modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
```diff
@@ -89,6 +89,13 @@ protected function extractToken(RequestInterface $request) {
 
     // Access token may be on the request, or in the headers.
     $input = $request->getParsedInput();
+
+    // If we don't have an $key_name on either the URL or the in the headers,
```
amitaibu, line 93: an $key_name => a $key_name

modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
```diff
@@ -89,6 +89,13 @@ protected function extractToken(RequestInterface $request) {
...
+    // then check again using a hyphen instead of an underscore.
+    if (empty($input[$key_name]) && empty($request->getHeaders()->get($key_name)->getValueString())) {
```
e0ipso, line 95: We don't need the `empty()`.

``` php
if (empty($input[$key_name]) && $request->getHeaders()->get($key_name)->getValueString()) {
```

modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
amitaibu, line 93: maybe it's worth also adding in the comment the reason we do this (e.g. Apache's new versions not accepting headers with underscores)

e0ipso:
Merged

e0ipso:
I made some minor modifications before merging it.

PR #830, opened 2016-01-24
In-memory caching of authentication plugins.

emisaacson:
When there are a lot of sub-entities returned from a rest call
the getDefinitions method gets called often and each time
it hits the database to return the cached authentication plugins.
This could result in dozens or hundreds of redundant database
calls per request that look like this:

```
SELECT cid, data, created, expire, serialized FROM {cache} WHERE cid IN (:cids)

a:1:{s:5:":cids";a:1:{i:0;s:22:"authentication_plugins";}}
```

Adding a memory cache of the definitions ensures there are no
redundant database calls. This cut the TTFB of an in-the-wild
service by more than half.

### DIFF
--- a/src/Plugin/AuthenticationPluginManager.php
+++ b/src/Plugin/AuthenticationPluginManager.php
@@ -62,7 +62,7 @@ class AuthenticationPluginManager extends DefaultPluginManager {
    * {@inheritdoc}
    */
   public function getDefinitions() {
-    $definitions = &drupal_static(__FUNCTION__);
+    $definitions = &drupal_static(__METHOD__);
     if (!isset($definitions)) {
       $definitions = parent::getDefinitions();
     }
--- a/src/Plugin/AuthenticationPluginManager.php
+++ b/src/Plugin/AuthenticationPluginManager.php
@@ -57,4 +57,24 @@ class AuthenticationPluginManager extends DefaultPluginManager {
     return new static(Module::getNamespaces(), _cache_get_object($bin));
   }
 
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDefinitions() {
+    $definitions = &drupal_static(__FUNCTION__);
+    if (!isset($definitions)) {
+      $definitions = parent::getDefinitions();
+    }
+    return $definitions;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDefinition($plugin_id, $exception_on_invalid = TRUE) {
+    $definitions = $this->getDefinitions();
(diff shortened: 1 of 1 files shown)


src/Plugin/AuthenticationPluginManager.php
```diff
@@ -57,4 +57,24 @@ public static function create($bin = 'cache') {
     return new static(Module::getNamespaces(), _cache_get_object($bin));
   }
 
+
+  /**
+   * {@inheritdoc}
+   */
+  public function getDefinitions() {
+    $definitions = &drupal_static(__FUNCTION__);
```
e0ipso:
``` php
$definitions = &drupal_static(__METHOD__);
```

e0ipso:
`parent::getDefinitions` will only give you definitions for a certain type. Isn't this code redundant with `DefaultPluginManager::getDefinitions()`?

Cache backend: 

And `DiscoveryCachedTrait`

Cache in memory:

emisaacson:
@e0ipso It's almost redundant, but with 1 difference, which is that mine caches the definitions in memory for the duration of the request (across all AuthenticationPluginManager instances) while the existing code caches in per-object scope.

During creation of each Resource AuthenticationManager->addAllAuthenticationProviders gets called [here](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/resource/Resource.php#L630) which creates an AuthenticationPluginManager instance in its own scope, then calls out multiple times to AuthenticationManager->addAuthenticationProvider, which each time creates a new AuthenticationPluginManager instance in its own scope, and this results in ((1 + _number of authentication providers_) \* _number of resources_) AuthenticationPluginManager instances created per request and an equal number of database calls. In our case, we have an endpoint that creates 200 resources on a single request so this causes 800 redundant database calls and it's very costly.

The code in this PR solves the issue for us but I could definitely see an argument instead for some refactoring to ensure that a reasonably small number of AuthenticationPluginManager instances are created and the already-created instance is passed in during creation of the resource. That would prevent needing drupal_static or changing AuthenticationPluginManager, it may be better or may not be, I'm definitely open to suggestions and would be happy to rework this pr based on your direction.

e0ipso:
@emisaacson I have created https://github.com/RESTful-Drupal/restful/pull/837 to address this issue from a different angle. Now we'll be able to leverage this in other plugin managers if needed.

Can you check that the performance gain is the same?

e0ipso:
I'm closing this one out for organization purposes. Please, feel free to request to re-open.

Thanks for all your work!

PR #834, opened 2016-01-27
Ref 833, Hotfix for hyphen/dash auth headers issue.

### LINKED ISSUE #833: Critical Bug: Access-Token Header Doesn't Work
Using the latest branch `7.x-2.x`, when sending a request with a header like:

```
Access-Token: 123
```

The access token is not read from the headers.

This is because on line 95 of `vendor/drupal/restful/modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php` the condition is incorrect. It is currently:

```
if (empty($input[$key_name]) && $request->getHeaders()->get($key_name)->getValueString()) {
```

And it should be something like:

```
if (empty($input[$key_name]) && empty($request->getHeaders()->get($key_name)->getValueString())) {
```

lukewertz:
Fixes #833

### DIFF
--- a/modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
+++ b/modules/restful_token_auth/src/Plugin/authentication/TokenAuthentication.php
@@ -93,7 +93,7 @@ class TokenAuthentication extends Authentication {
     // If we don't have a $key_name on either the URL or the in the headers,
     // then check again using a hyphen instead of an underscore. This is due to
     // new versions of Apache not accepting headers with underscores.
-    if (empty($input[$key_name]) && $request->getHeaders()->get($key_name)->getValueString()) {
+    if (empty($input[$key_name]) && empty($request->getHeaders()->get($key_name)->getValueString())) {
       $key_name = str_replace('_', '-', $key_name);
     }
 


lukewertz:
This is a fairly urgent issue for me, @e0ipso. Sorry for needing to propose the hotfix.

@josephdpurcell outlines the problem here: https://github.com/RESTful-Drupal/restful/issues/833#issuecomment-175280749

This should address the issue that he is experiencing. The 2.x branch right now won't process any access-token headers!

josephdpurcell:
Suggestion: use `!` instead of `empty`. That should fix the failing test.

e0ipso:
@lukewertz I agree with the comment above. Also, tests are failing. :-(

fpalluel:
Hello,

The patch is failing with PHP 5.4 only, isn't it ?
Since PHP 5.4 is not supported any longer... ;-)

e0ipso:
@Moxide Agreed. Let's not hold this any longer and let's make it into the next release.

e0ipso:
Modified and merged.

e0ipso:
Thanks everyone!

PR #837, opened 2016-01-28
[FEATURE] Create semi singleton for plugin managers

e0ipso:
And apply it to the authentication manager. This adds a performance
boost.


xuxucode:
Cool! I test it, singleton is more performant than `drupal_static`.

e0ipso:
@edwardaa thanks for the review! I'm going to merge this one.

amitaibu:
Why not always go with the singleton?

PR #839, opened 2016-01-28
[BUGFIX] Move the resource to a new test module

e0ipso:
Allow implementors to use the 'articles' namespace when Token Auth
is enabled.


e0ipso:
Merged.

PR #842, opened 2016-01-30
Fix: get correct range to output the next link

xuxucode:
When generating`next` link in [FormatterJson](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/formatter/FormatterJson.php#L175), the `$item_per_page` is always 50, it should change by request input, like what [DataProvider#parseRequestForListPagination()](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/resource/DataProvider/DataProvider.php#L425-L426) does.

### DIFF
--- a/src/Plugin/formatter/FormatterHalJson.php
+++ b/src/Plugin/formatter/FormatterHalJson.php
@@ -123,8 +123,9 @@ class FormatterHalJson extends Formatter implements FormatterInterface {
     $input = $request->getParsedInput();
     $page = !empty($input['page']) ? $input['page'] : 1;
 
+    $query = $input;
     if ($page > 1) {
-      $query = array('page' => $page - 1) + $input;
+      $query['page'] = $page - 1;
       $data['_links']['previous'] = array(
         'title' => 'Previous',
         'href' => $resource->versionedUrl('', array('query' => $query), TRUE),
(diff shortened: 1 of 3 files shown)


e0ipso:
Good catch! Is this also fixing it for JSON API?

e0ipso:
We seem to be addressing this in another way in `FormatterJsonApi`. Do you think it would be appropriate to do it like that?

https://github.com/edwardaa/restful/blob/json_range_next_link/src/Plugin/formatter/FormatterJsonApi.php#L195-L227

src/Plugin/formatter/FormatterHalJson.php
```diff
@@ -121,14 +121,14 @@ protected function addHateoas(array &$data) {
...
     $page = !empty($input['page']) ? $input['page'] : 1;
 
+    $query = $input;
     if ($page > 1) {
-      $input['page'] = $page - 1;
+      $query['page'] = $page - 1;
       $data['_links']['previous'] = array(
         'title' => 'Previous',
-        'href' => $resource->getUrl(),
+        'href' => $resource->versionedUrl('', array('query' => $query), TRUE),
```
e0ipso, line 131: You can pass the `$query` to `->getUrl()` as well. Is there any other reason why you want to use `->versionedUrl()`?

xuxucode, line 131: No, just keep it consistent with [FormatterJson#L168](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/formatter/FormatterJson.php#L168).

e0ipso:
I had to fix something related to this in https://github.com/RESTful-Drupal/restful/pull/856. Is this PR still needed after that one?

xuxucode:
Thanks, closed.

PR #856, opened 2016-02-15
[BUGFIX] Use correct range in pagination when exceeding max

e0ipso:
When a range is passed via query string parameter, the correct
pagination links should be calculated. Even if the passed range
exceeds the maximum allowed range.

### DIFF
--- a/src/Plugin/formatter/Formatter.php
+++ b/src/Plugin/formatter/Formatter.php
@@ -7,7 +7,6 @@
 
 namespace Drupal\restful\Plugin\formatter;
 
-use Doctrine\Common\Collections\ArrayCollection;
 use Drupal\Component\Plugin\PluginBase;
 use Drupal\restful\Exception\ServerConfigurationException;
 use Drupal\restful\Plugin\ConfigurablePluginTrait;
@@ -17,8 +16,12 @@ use Drupal\restful\Plugin\resource\Field\ResourceFieldCollectionInterface;
 use Drupal\restful\Plugin\resource\ResourceInterface;
 use Drupal\restful\RenderCache\Entity\CacheFragment;
 use Drupal\restful\RenderCache\RenderCache;
-use Drupal\restful\RenderCache\RenderCacheInterface;
 
+/**
+ * Class Formatter.
+ *
+ * @package Drupal\restful\Plugin\formatter
+ */
 abstract class Formatter extends PluginBase implements FormatterInterface {
 
   use ConfigurablePluginTrait;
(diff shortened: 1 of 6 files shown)


tests/RestfulSimpleJsonTestCase.test
```diff
@@ -81,6 +81,23 @@ class RestfulSimpleJsonTestCase extends RestfulCurlBaseTestCase {
     $this->assertEqual($decoded_json['count'], 2, 'Count was populated correctly.');
     $this->assertEqual(count($decoded_json['data']), 1, 'The correct number of items was listed.');
     $this->assertNotNull($decoded_json['next'], '"Next" property added.');
+
+    // Test the pagination links for bigger ranges.
+    $handler->setRequest(Request::create('/api/v1.5/articles', array(
+      'range' => 3,
```
amitaibu: Maybe change this to `1000` just so it's clear we're passing the allowed values?

e0ipso: Will do.

e0ipso:
Tests are green.

src/Plugin/formatter/Formatter.php
```diff
@@ -309,4 +314,23 @@ protected static function unprefixInputOptions($allowed_fields, $prefix) {
...
+   * Helper function that calculates the number of items per page.
+   *
+   * @param ResourceInterface $resource
+   *   The associated resource.
+   *
+   * @return int
+   *   The items per page.
+   */
+  protected function calculateItemsPerPage(ResourceInterface $resource = NULL) {
+    $resource = $resource ?: $this->getResource();
```
amitaibu: last nitpick, out of curiosity -- why allow both options, and not force passing the `$resource`?

e0ipso: `TRUE` you'll need a resource to get the items per page.

amitaibu: yeah, it's clear you need the resource, my question is why have this `$resource = $resource ?: $this->getResource();` -- instead of forcing passing it as an argument?

e0ipso: Sorry, I was agreeing with you're line of thought. There's a fix for it already ;-)

e0ipso:
Let me know when you're OK with it. Instead of clicking the merge button I'd rather squash and craft a single meaningful commit message for the release notes.

amitaibu:
:+1:

amitaibu:
looks fine to me

PR #857, opened 2016-02-16
[CLEANUP] Deprecate PHP 5.4 since Plug depends on 5.5.9

e0ipso:
The 5.4 version support was unrealistic since one of the
dependencies already depens on PHP 5.5.9.


e0ipso:
This goes in.

PR #858, opened 2016-02-16
[FEATURE] Switch user during RESTful execution

e0ipso:


e0ipso:
Yay! Back to green.

@amitaibu @jeff-h I'm still a bit uneasy about doing this. I fear some sort of privilege escalation.

Mind giving a review? This should be a port from the 1.x feature.

jeff-h:
This is fantastic! Will try it out as soon as I can, although my head's pretty deep in the client side right now.

e0ipso:
This was merged, if anything arises from your analysis we can address it as a follow up.

amitaibu:
This looks like a direct(ish) port from 1.x -- which is working fine, so for an eyeball review, it looks good.

amitaibu:
I think that on this specific issue, my feeling was "what can possibly go wrong" :wink:

e0ipso:
hahaha. Awesome!

PR #860, opened 2016-02-22
[FEATURE] Allow other auth providers to use Authorization

e0ipso:
Let other providers to use the Authorization header. The Basic
prefix will be used to detect basic authentication when that
header is used. This allows Simple OAuth.

Additionally changes the String class name, since it's not
compatible with PHP7.


e0ipso:
@amitaibu I'm not too concerned about the class name change since it was not used in RESTful (the uses were refactored out). Since the only method was not documented I don't expect anyone to be using it in the wild.

src/Util/StringHelper.php
```diff
@@ -0,0 +1,47 @@
...
+ * @file
+ * Contains \Drupal\restful\Util\String.
+ */
+
+namespace Drupal\restful\Util;
+
+class StringHelper {
+
+  /**
+   * Turns a string into camel case. From search_api_index to SearchApiIndex.
```
amitaibu: `From search_api_index to SearchApiIndex.` can be removed

amitaibu:
:+1:

e0ipso:
Tests seem to be failing.

src/Util/StringHelper.php
e0ipso: Done.

rsanzante:
To compatibilize RESTfull with PHP 7 a change is needed in src/Plugin/resource/Field/ResourceFieldEntity.php. There the old Drupal\restful\Util\String is still used.

Patch pasted. Probably is not the way to go but I don't know to do it, I'm too used to Drupal.org ussue queues, and create another pull request seems an overkill for such a tiny change.

```
diff --git a/sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php b/sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php
index 31b43d2..6456c78 100644
--- a/sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/sites/all/modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -17,7 +17,7 @@
 use Drupal\restful\Plugin\resource\Field\PublicFieldInfo\PublicFieldInfoEntity;
 use Drupal\restful\Plugin\resource\Field\PublicFieldInfo\PublicFieldInfoEntityInterface;
 use Drupal\restful\Plugin\resource\Field\PublicFieldInfo\PublicFieldInfoInterface;
-use Drupal\restful\Util\String;
+use Drupal\restful\Util\StringHelper;

 /**
  * Class ResourceFieldEntity.
```

Oh, and... is this class really used? I don't see a direct use of it in the RESTful code.

rsanzante:
Created a pull request.

https://github.com/rsanzante/restful/pull/1

e0ipso:
@rsanzante thanks for the PR! I could not merge it, but I realized that we're not using that class anywhere after all. See 7d3acf5.

e0ipso:
I'm merging this one.

rsanzante:
Ok, anyway now it's PHP7 compatible. Thanks!

PR #871, opened 2016-02-26
Be able to used the overrided setPath.

iStryker:
Line 652 sets the Path, but if you override it, it is not passed on to the next method.

### DIFF
--- a/plugins/restful/RestfulBase.php
+++ b/plugins/restful/RestfulBase.php
@@ -670,7 +670,7 @@ abstract class RestfulBase extends \RestfulPluginBase implements \RestfulInterfa
       $this->getRateLimitManager()->checkRateLimit($request);
     }
 
-    return $this->{$method_name}($path);
+    return $this->{$method_name}($this->path);
   }
 
   /**


e0ipso:
There are merge conflicts.

e0ipso:
Also, what is the use case of using `->process()` programmatically? In tests we use `doGet()`, etc.

iStryker:
Yeah, I do not know why there is a merge conflict.  I pulled down the latest 7.x-2.x.   

There are modules that use the path of request to build their "thing".  The path will be of the api (for example api/v1.0/{resource}).   This will cause problems.   For me, it was the workbench_moderation module, which builds the moderation tabs from the url.

For my project, I am using 7.x-1.x.   I put a pull request for this #870.  I saw the problem was still there in 7.x-2.x so I put in a pull request for it as well.

e0ipso:
> Yeah, I do not know why there is a merge conflict. I pulled down the latest 7.x-2.x.

It's probably some recent commit.

e0ipso:
I believe that you may have a misconfigured git repo. 2.x doesn't even have a `RestfulBase`.

Additionally, this is not an issue in 2.x.

Thanks for contributing though!!

PR #875, opened 2016-03-06
Field access for sorting and filtering (1.x)

bitamar:
Preventing filtering and sorting by fields that were hidden with hook_field_access.

PR #876, opened 2016-03-06
Field access for sorting and filtering (2.x)

bitamar:
Preventing filtering and sorting by fields that were hidden with hook_field_access.


amitaibu:
Thanks. Same comments as #875 :)

PR #885, opened 2016-03-23
[FEATURE] Allow altering filters to support custom filters

e0ipso:
Allow an easy way of altering the filtering by a custom data
provider. This is the only recommended way to add support for
filtering based on:
- [x] Computed public fields: reverse the computation for the filter.
- [x] Fields with preprocesses: reverse the preprocess for the
  filter. For instance, in a date field that is stored as a UNIX
  timestamp but displayed as a ISO-8601 the consumer wants to be
  able to accept filters with ISO-8601 dates in them. The
  alterFieldQuery would take the date string and parse it back
  to a UNIX timestamp.


e0ipso:
It took more than I expected for this to be needed.

e0ipso:
tests are green. Merging.

PR #889, opened 2016-03-28
Fix content type format sent from chrome on iPads.

bricel:
This cause restful to choose the wrong formatter.
see https://github.com/Gizra/productivity/issues/248.


e0ipso:
This looks good. Any chance to get a PR for `7.x-2.x` version as well?

bricel:
@e0ipso - Looks like `7.x-2.x` already handles this case [here](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Formatter/FormatterManager.php#L101).

PR #893, opened 2016-04-07
[FEATURE] Allow joining queries for filters on a target column

e0ipso:
Instead of assuming that joins on entity tables always happen at
the entity key level.

To do so use the `targetColumn` property in the field definition.


e0ipso:
This also fixes a bug that went out with 2.8

PR #906, opened 2016-04-28
[BUGFIX] Better compound ID getter to fix targetted filters

e0ipso:
Fix the targetted filters when using JSON API.


amitaibu:
Can you please give an example for this - I'm not clear from the code.

e0ipso:
@amitaibu this is to support https://gist.github.com/e0ipso/1acbced683908ff631e3 in JSON API.

We are already supporting it for the other formatters, but JSON API had a bug and it was not reading the request parameters. Rendering the relationship will take care of it automagically.

amitaibu:
👍

PR #918, opened 2016-05-04
For to-many relationships resource identifier objects must be unique.

d2ev:
For to-many relationships, It should filter out the duplicates `resource identifier object` in relationships.
Ref - https://github.com/json-api/json-api/blob/gh-pages/schema#L244-L251


e0ipso:
Thanks for the contribution @d2ev!

PR #920, opened 2016-05-05
[FEATURE] Allow RESTful module to support \EntityStructureWrapper types in the buildResourceMetadataItem function.

### LINKED ISSUE #919: Support EntityStructureWrapper when building the metadata item for a field value
We have a custom field on our site that stores node ids in three columns, essentially three entity references in one field.  It doesn't work with restul module b/c restful doesn't know ho to traverse the relationship, however after much debugging I was able to find a simple patch that makes things work.  Pull request imminent.

m4olivei:
Resolves #919.

I'm wondering if this needs to be careful about calling `$this->getSubProperty()`?  For me it works b/c the field types that end up using this are always defined with a `sub_property`, eg.

``` php
$public_fields['show'] = array(
  'property' => 'field_tv_shows',
  'sub_property' => 'show',
  'resource' => array(
    'name' => 'shows',
    'majorVersion' => '1',
    'minorVersion' => '0',
  ),
);
```

### DIFF
--- a/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -1118,6 +1118,9 @@ class ResourceFieldEntity implements ResourceFieldEntityInterface {
     if ($wrapper instanceof \EntityValueWrapper) {
       $wrapper = entity_metadata_wrapper($this->getEntityType(), $wrapper->value());
     }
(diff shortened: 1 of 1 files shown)


m4olivei:
In addition to the failing tests, I just noticed it fails with my setup as well.  I believe it stems from the fact that `EntityDrupalWrapper` extends `EntityStructureWrapper`, so the if statement is catching too much.

PR #925, opened 2016-06-16
[FEATURE] Cache the already embedded content

e0ipso:
If two entities of the top resource point to the same entity, then
it's embedded twice. For performance reasons we want to calculate
it once and cache the result througout the request.


e0ipso:
Green! Yay!

PR #927, opened 2016-06-27
[BUGFIX] Fix the JSON API output with relationships

e0ipso:
If the relationship contains a resource but the fullView is set
to FALSE, we should not consider that element for inclusion.

### DIFF
--- a/src/Plugin/formatter/FormatterJsonApi.php
+++ b/src/Plugin/formatter/FormatterJsonApi.php
@@ -754,20 +754,17 @@ class FormatterJsonApi extends Formatter implements FormatterInterface {
    *   The links.
    */
   protected function buildRelationshipLinks($parent_id, $public_field_name, ResourceInterface $resource_plugin, $identifier) {
-    // If there is a resource plugin for the parent, set the related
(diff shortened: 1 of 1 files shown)


src/Plugin/formatter/FormatterJsonApi.php
```diff
@@ -738,4 +738,36 @@ protected static function retrieveIncludedItem($type, $id, array $included) {
...
+    $links = array();
+    if ($resource = $this->getResource()) {
+      $links['related'] = $resource->versionedUrl('', array(
+        'absolute' => TRUE,
+        'query' => array(
+          'filter' => array($public_field_name => $identifier),
+        ),
+      ));
+      $links['self'] = $resource_plugin->versionedUrl($parent_id . '/relationships/' . $public_field_name);
+      return $links;
```
amitaibu: you probably don't need that, or change `IF` to return early.

e0ipso: You are right. PHPStorm refactor feature is very smart, but needs human 👀 on it.

PR #935, opened 2016-08-02
Update DataProviderDbQuery.php

jeff-h:
Data providers' `create()` method must return an array.


jeff-h:
My previous PR (https://github.com/RESTful-Drupal/restful/pull/786) overlooked the need for the `create()` method to also return an array. This fixes that.

e0ipso:
Sorry it took so long.

Thanks for all your effort with this module! Would you feel confident enough with it to become a maintainer?

jeff-h:
I'm honoured to be asked! :) However I don't feel I could do justice to this; there are still many use-cases that I have no clue about. I will keep trying to do what I can whenever I can in the wiki though!

Issue #939, opened 2016-08-24
Access resource plugin definition from an authentication plugin


hussainweb:
I am trying to write an authentication plugin for OAuth2 support and need to access resource's plugin definition. The idea is to allow defining the oauth2 server and scope required to access a particular resource.

OAuth2 server supports RESTful 7.x-1.x but needs variables to specify server and scope (see #111, and https://www.drupal.org/node/2451303). This is simple if you only have one server and scope. I imagine more than one OAuth2 server would be rare but more than one scope could be common. Anyway, variables don't seem like the best way to handle the issue anyway.

My problem now is that I cannot access the plugin definition from authentication plugin at all. I only get a Request object and using that (and the global restful controller using `restful()`), I can recreate the plugin and get the definition but that is a lot of code for something that could be done simply. Also, it might involve recreating the plugin just to get the definition.

Am I missing something here?

hussainweb:
I started working with the global way to access plugins and noticed another problem. This is my code:

```
$resource = restful()->getResourceManager()->negotiate();
$plugin = $resource->getPluginDefinition();
```

This works but fails for discovery endpoint, because the above returns the definition for discovery resource. This is called by the discovery plugin's view logic that checks for user's access to each plugin.

Ideally, there should be a way to pass in the context to the authentication plugin which contains the resource plugin being checked. In case of a direct request, it will be the same plugin returned as above. In case of discovery plugin (and maybe other similar cases), it will be different.

e0ipso:
Going the other way around is easier, since given a plugin definition and a request getting the authentication plugin is piece of cake.

As you point out the best way to do this is to use the _"service container"_ to get the resource manager and the resource plugin from there.

> This works but fails for discovery endpoint, because the above returns the definition for discovery resource

I'm not sure what you mean by _discovery resource_, all resources should have the discovery and data capabilities. That is triggered based on the HTTP verb (OPTIONS/GET).

hussainweb:
@e0ipso, thanks for your response.

Two lines to load the plugin is fair, but does this not instantiate the plugin again? I hope there are no side effects in doing this. I am only interested in the definition. This is how the plugin definition currently looks:

```
/**
 * Users endpoint.
 *
 * @Resource(
 *   name = "user:1.0",
 *   resource = "user",
 *   label = "Users",
 *   description = "List, get, and update users on the site.",
 *   authenticationTypes = {
 *     "oauth2_auth"
 *   },
[...]

PR #940, opened 2016-08-24
Add ability to retrieve resource name from the resource manager

hussainweb:
The ResourceManager class provides methods to get version information and even create a new
instance of a plugin based on the current request. However, the methods that help in getting
the resource id (such as ::getMenuItem and getPageArguments) are protected.

Since we are already able to get information on version, we should also be able to get the
resource id using the resource manager class. This commit adds a method to do that.

This blocks issue #939 where we need to get the resource plugin definition from an
authentication plugin. Calling negotiate or trying to retrieve menu item information via
menu_get_item results in an recursive call (infinite).

### DIFF
--- a/src/Resource/ResourceManager.php
+++ b/src/Resource/ResourceManager.php
@@ -121,6 +121,19 @@ class ResourceManager implements ResourceManagerInterface {
   /**
    * {@inheritdoc}
    */
+  public function getResourceIdFromRequest() {
+    $resource_name = &drupal_static(__METHOD__);
+    if (isset($resource_name)) {
+      return $resource_name;
+    }
+    $path = $this->request->getPath(FALSE);
+    list($resource_name,) = static::getPageArguments($path);
+    return $resource_name;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
   public function getVersionFromRequest() {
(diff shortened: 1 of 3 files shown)


e0ipso:
Can we get a test for this? I'm specially interested in seeing that this doesn't break unexpectedly with the _synthetic_ request object used for the subrequests in the `includes` (for the embedded resources).

hussainweb:
@e0ipso, thanks for your response.

I consider that this method is related to `getVersionFromRequest()` which is tested in `RestfulHookMenuTestCase`. I can add a test there in `testVersionNegotiation()` to see that the resource name is 'aricles'. Does that sound good to you?

I saw the tests for subrequests but they hardly seem to be calling any methods on the resource manager to identify the resource. The closest I saw was checking the major and minor version of sub resource in `\RestfulSubResourcesCreateEntityTestCase::testCreateAndUpdateEntity` but that was done by directly getting the field definitions and checking the array. Am I missing something here?

hussainweb:
@e0ipso, any reviews on this one? I am just wondering if you think the approach here is in the right direction.

e0ipso:
Merged! Thanks @hussainweb for your contributions and your patience.

PR #941, opened 2016-08-29
Update ResourceManagerInterface.php

Yonimdo:
The doc says 

> Ex: array(array($this, 'processing'), array('param1', 2)) 

but the function require another wrapping array

### DIFF
--- a/src/Resource/ResourceManagerInterface.php
+++ b/src/Resource/ResourceManagerInterface.php
@@ -79,7 +79,7 @@ interface ResourceManagerInterface {
    *       Ex: array($this, 'format').
    *     - An array containing any of the methods before and an array of
    *       parameters to pass to the callback.
-   *       Ex: array(array($this, 'processing'), array('param1', 2))
+   *       Ex: array(array(array($this, 'processing'), array('param1', 2)))
    * @param array $params
    *   Array of additional parameters to pass in.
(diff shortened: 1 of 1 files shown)


e0ipso:
Hmm, I could swear that no wrapping array is necessary. Can you provide an example?

Yonimdo:
snapshot1
snapshot2

e0ipso:
You can see that the extra array is because you can have multiple process
callbacks.

Yonimdo:
so the doc is still wrong...

e0ipso:
No... They're fine.

I think that you missed the fact that 'process_callbacks' calls multiple times to this method.

Yonimdo:
Got your point.. thank you

amitaibu:
@MoodyUser can you please confirm you were able to get the result with the current setup (I remember it didn't work for you)

Yonimdo:
It didn't. but as @e0ipso said the doc talks about a callback array inside that list (array of arrays) that calls multiple time therefore the doc is right but not intuitive

snapshot8

the same problem?
https://github.com/MoodyUser/restful/blob/e6e6f61aa97917d6d666503c1f19669d2dc9f9de/src/Resource/ResourceManagerInterface.php#L79 
snapshot9

PR #942, opened 2016-08-30
Add OAuth2 authentication provider

hussainweb:
This change depends on the changes in PR #940. It uses the new method introduced in that PR
to identify the resource being requested and get the plugin definition to read the oauth2
server and scope to use for authorizing the request.

This would probably need to be changed once #940 is merged or a better method to access the
request being processed is identifed (as mentioned in #939). I am creating this PR in hope
of clarifying the changes in PR #940.

### DIFF
--- a/src/Plugin/authentication/OAuth2ServerAuthentication.php
+++ b/src/Plugin/authentication/OAuth2ServerAuthentication.php
@@ -78,13 +78,13 @@ class OAuth2ServerAuthentication extends Authentication {
 
     $plugin_definition = ResourcePluginManager::create('cache', $request)->getDefinition($plugin_id);
 
-    $server = !empty($plugin_definition['oauth2Server']) ? $plugin_definition['oauth2Server'] : variable_get('oauth2_server_restful_server');
-    if (!$server) {
+    if (empty($plugin_definition['oauth2Server'])) {
       return NULL;
     }
 
-    $scope = !empty($plugin_definition['oauth2Scope']) ? $plugin_definition['oauth2Scope'] : variable_get('oauth2_server_restful_scope');
-    return ['server' => $server, 'scope' =>$scope];
+    $server = $plugin_definition['oauth2Server'];
+    $scope = !empty($plugin_definition['oauth2Scope']) ? $plugin_definition['oauth2Scope'] : '';
+    return ['server' => $server, 'scope' => $scope];
   }
 
   /**
--- a/src/Plugin/authentication/OAuth2ServerAuthentication.php
+++ b/src/Plugin/authentication/OAuth2ServerAuthentication.php
@@ -3,6 +3,7 @@
 namespace Drupal\restful\Plugin\authentication;
 
 use Drupal\Component\Plugin\PluginBase;
+use Drupal\restful\Exception\ServerConfigurationException;
 use Drupal\restful\Exception\UnauthorizedException;
 use Drupal\restful\Http\RequestInterface;
 use Drupal\restful\Plugin\ResourcePluginManager;
@@ -11,7 +12,7 @@ use Drupal\restful\Plugin\ResourcePluginManager;
  * Authentication support for oauth2_server.
  *
  * @Authentication(
- *   id = "oauth2_auth",
+ *   id = "oauth2",
  *   label = "OAuth2 authentication",
  *   description = "Authenticate requests based on oauth2_server auth.",
  * )
@@ -19,10 +20,22 @@ use Drupal\restful\Plugin\ResourcePluginManager;
 class OAuth2ServerAuthentication extends Authentication {
 
   /**
+   * The resource manager.
+   *
+   * @var \Drupal\restful\Resource\ResourceManagerInterface
+   */
+  protected $resourceManager;
+
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->resourceManager = restful()->getResourceManager();
+  }
+
+  /**
    * {@inheritdoc}
    */
   public function applies(RequestInterface $request) {
-    return module_exists('oauth2_server') && $this->getResourcePluginIdFromRequest();
+    return module_exists('oauth2_server') && $this->getOAuth2Info($request);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+use Drupal\Component\Plugin\PluginBase;
+use Drupal\restful\Exception\UnauthorizedException;
+use Drupal\restful\Http\RequestInterface;
+use Drupal\restful\Plugin\ResourcePluginManager;
+
+/**
+ * Authentication support for oauth2_server.
+ *
+ * @Authentication(
+ *   id = "oauth2_auth",
```
e0ipso: `oauth2_auth` ➡️  `oauth2`

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+ *   description = "Authenticate requests based on oauth2_server auth.",
+ * )
+ */
+class OAuth2ServerAuthentication extends Authentication {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function applies(RequestInterface $request) {
+    return module_exists('oauth2_server') && $this->getResourcePluginIdFromRequest();
```
e0ipso: This method should check the request to see if it contains the necessary info to perform the required checks.

This probably means moving `$oauth2_info = $this->getOAuth2Info($request);` here.

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+    $result = oauth2_server_check_access($oauth2_info['server'], $oauth2_info['scope']);
+    if ($result instanceof \OAuth2\Response) {
+      throw new UnauthorizedException($result->getResponseBody(), $result->getStatusCode());
+    }
+    elseif (is_array($result) && !empty($result['user_id'])) {
+      return user_load($result['user_id']);
+    }
+  }
+
+//  protected function getOAuth2Info() {
```
e0ipso: Leftovers.

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+    $oauth2_info = $this->getOAuth2Info($request);
+    if (!$oauth2_info) {
+      return NULL;
+    }
+
+    $result = oauth2_server_check_access($oauth2_info['server'], $oauth2_info['scope']);
+    if ($result instanceof \OAuth2\Response) {
+      throw new UnauthorizedException($result->getResponseBody(), $result->getStatusCode());
+    }
+    elseif (is_array($result) && !empty($result['user_id'])) {
```
e0ipso: Make this whole `ifelse`:

``` php
if (empty($result['user_id'])) {
  return NULL;
}
return user_load($result['user_id']);
```

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+      return user_load($result['user_id']);
+    }
+  }
+
+//  protected function getOAuth2Info() {
+//    return [variable_get('oauth2_server_restful_server'), variable_get('oauth2_server_restful_scope')];
+//  }
+
+  protected function getOAuth2Info($request) {
```
e0ipso: Needs docblock.

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+    $plugin_id = $this->getResourcePluginIdFromRequest();
+    $plugin = ResourcePluginManager::create('cache', $request)->getDefinition($plugin_id);
```
e0ipso: `$plugin` ➡️  `$plugin_definition`.

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+
+    $server = !empty($plugin['oauth2Server']) ? $plugin['oauth2Server'] : variable_get('oauth2_server_restful_server');
```
e0ipso: Let's drop the global variables: `oauth2_server_restful_server` and `oauth2_server_restful_scope` and require the annotation to contain the info. If the info is not there we can throw a 5XX exception.

e0ipso: And let's move it under the `'options'` key in the annotation. See https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Annotation/Authentication.php#L35-L40

src/Plugin/authentication/OAuth2ServerAuthentication.php
e0ipso: Also, did you try using the resource manager's: `->getPlugin($plugin_id, $request)` instead?

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,74 @@
...
+    if (!$server) {
+      return NULL;
+    }
+
+    $scope = !empty($plugin['oauth2Scope']) ? $plugin['oauth2Scope'] : variable_get('oauth2_server_restful_scope');
+    return ['server' => $server, 'scope' =>$scope];
+  }
+
+  protected function getResourcePluginIdFromRequest() {
+    $resource_name = restful()->getResourceManager()->getResourceIdFromRequest();
```
e0ipso: Add a protected attribute in the class `$resourceManager` and initialize it in this class' constructor to `restful()->getResourceManager()`. That way there is a single instance used across the class.

src/Plugin/authentication/OAuth2ServerAuthentication.php
e0ipso: If you do that you'll be able to leverage https://github.com/RESTful-Drupal/restful/pull/942/files#r76750941 here as well.

e0ipso:
Did you try that this works well with nested resources?

Imagine that the user can only view articles, but not user profiles. Would `/articles/123?include=author` fail accordingly to show the profile in the response?

This resource embedding clones part of the request. We need to make sure that permissions are still enforced.

e0ipso:
@hussainweb thank you for working on this!! I think that this is a very important feature.

I added a couple of comments above.

src/Plugin/authentication/OAuth2ServerAuthentication.php
hussainweb: I was using `basic_auth` as an example. I see other plugins don't follow this convention. Changing...

src/Plugin/authentication/OAuth2ServerAuthentication.php
hussainweb: @e0ipso, Not really leftover, but until #940 goes in, I was keeping in an alternate version I could uncomment for testing. I will remove it once you approve the direction.

src/Plugin/authentication/OAuth2ServerAuthentication.php
hussainweb: I am not very sure. I thought there might be something useful in checking for `\OAuth2\Response` and handling that?

src/Plugin/authentication/OAuth2ServerAuthentication.php
hussainweb: I did use `->getPlugin` in the beginning. It results in an infinite recursion. Calling `getPlugin` in `authenticate` or `applies` ultimately tries to create the plugin again, which tries to run authentication again. That is why I created #940.

src/Plugin/authentication/OAuth2ServerAuthentication.php
e0ipso: For clarity, I meant «Make this whole `elseif`:»

I'd leave the `if` above alone, but simplify the logic (without leaving any execution path without a `return`).

src/Plugin/authentication/OAuth2ServerAuthentication.php
hussainweb: The `options` key is on `@Authentication` annotation. I must put that on the OAuth2Authentcation class. That does not help the use case where resources would want to specify the server and scope.

src/Plugin/authentication/OAuth2ServerAuthentication.php
hussainweb: It was a simple getter, so I thought it would always be the same instance. But I see your point.

hussainweb:
@e0ipso, thank you for all the comments and an elaborate review. I am pushing changes now.

> Imagine that the user can only view articles, but not user profiles. Would /articles/123?include=author fail accordingly to show the profile in the response?

I am not very sure. If the request in resource manager is replaced for the sub-request (which I suspect), then yes. It should work as expected. I don't have any resources right now which embed this partially. Any suggestions on how to easily test this?

Also, how do other authentication types enforce this? e.g., BasicAuthentication also reads the code from the request, and if that works for sub-requests, this should too.

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    $plugin_definition = ResourcePluginManager::create('cache', $request)->getDefinition($plugin_id);
+
+    $server = !empty($plugin_definition['oauth2Server']) ? $plugin_definition['oauth2Server'] : variable_get('oauth2_server_restful_server');
+    if (!$server) {
+      return NULL;
+    }
+
+    $scope = !empty($plugin_definition['oauth2Scope']) ? $plugin_definition['oauth2Scope'] : variable_get('oauth2_server_restful_scope');
```
e0ipso: Can we get rid of the `variable_get`?

e0ipso: (applies to other places as well)

e0ipso:
This looks good to me 👍. Thanks for this killer feature.

I'd love to see a section to the [wiki](https://github.com/RESTful-Drupal/restful/wiki/3.-Authentication) on how to configure this to have a resource use this authentication method. That will empower users even more.

hussainweb:
I'll try to get around to the wiki this week. Thank you for merging this!

Issue #937, opened 2016-08-18
Issues with PHP7


vincenzo:
Hi guys, 

Is `restful` actually supposed to work with PHP7?

In my installation, I get this:

```
PHP Fatal error:  Interface 'Drupal\restful\Util\ExplorableDecoratorInterface' not found in .... modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldResource.php on line 17
```

even though the file is in the right place. To work around this temporarily, I pasted the code of the interface definition in the file that is requiring it. Done that, it comes up with a similar error next, about another interface or class not found. 

Thoughts?

Restful 2.11
Drupal 7.44
PHP 7.0.9

e0ipso:
This seems to be an issue with the autoloader. I believe that other people have tested this with PHP7. I remember fixing some issues in the issue queue.

vincenzo:
@e0ipso Do you mean by that, that the issue is something external to Restul? As in my setup/config?

vincenzo:
@e0ipso I did something really simple: patched the restful.info file so that it has this line

```
 registry_autoload_files[] = src/Util/ExplorableDecoratorInterface.php
```

And now it all works. 
It does seem to suggest an issue with the autoloader, but do you have any idea what this may be? Are you guys testing this on PHP7 currently?

e0ipso:
Interesting... I'm not testing this with PHP7 atm. That seems like an acceptable workaround to make this work.

Can you submit a patch with a comment saying that this is a temporary workaround?

vincenzo:
I sure can.

rhclayto:
I am running the latest RESTful 2.x on PHP7 with no problems. I upgraded to PHP7 from 5.6, keeping RESTful as it was, & have had no problems. Just my own experience.

PR #943, opened 2016-09-02
Temporary workaround to allow RESTful to work on PHP7.

vincenzo:
This PR provides a temporary workaround for issue #937. For reasons I do not fully understand at the moment, the autoloader on PHP7 (at least v. 7.0.9 I use) fails to load `ExplorableDecoratorInterface`: 

```  
PHP Fatal error:  Interface 'Drupal\restful\Util\ExplorableDecoratorInterface' not found in .... modules/contrib/restful/src/Plugin/resource/Field/ResourceFieldResource.php on line 17
```

This is while using the latest `restful` stable release (it actually fails with the HEAD dev version, too) and Drupal 7.44. This simple change allows the module to work fine, so following @e0ipso's suggestion, I am issuing this PR.

### DIFF
--- a/restful.info
+++ b/restful.info
@@ -9,6 +9,9 @@ configure = admin/config/services/restful
 registry_autoload[] = PSR-0
 registry_autoload[] = PSR-4
 
+; Temporary workaround to allow RESTful to work fine on PHP7.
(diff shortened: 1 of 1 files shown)


e0ipso:
Thanks for contributing!

PR #944, opened 2016-09-05
Add support for Client Credentials Grant Type to OAuth2 authentication

hussainweb:
OAuth2 support requires that only tokens associated with users may be used with authenticated resources. However, OAuth2 also support a client credentials grant type which
does not require any user interaction. This is typically used directly by applications to access such API. This does not work with RESTful as once a resource is marked as
authenticated, there has to be an associated user.

This change allows plugin developers to specify a user which will be used when a client credentials token is used. The user may be specified with the uid or name. The uid
has the higher priority.

### DIFF
--- a/src/Plugin/authentication/OAuth2ServerAuthentication.php
+++ b/src/Plugin/authentication/OAuth2ServerAuthentication.php
@@ -51,8 +51,22 @@ class OAuth2ServerAuthentication extends Authentication {
     if ($result instanceof \OAuth2\Response) {
       throw new UnauthorizedException($result->getResponseBody(), $result->getStatusCode());
     }
-    elseif (empty($result['user_id'])) {
-      return NULL;
+
+    if (empty($result['user_id'])) {
+      // If the user_id is not set, it could mean that this is a client
+      // credentials grant token, in which case the client_id would be set.
+      if (empty($result['client_id'])) {
+        return NULL;
+      }
+
+      // We are dealing with client credentials flow. See if the resource has
+      // defined an user for this grant type.
+      if (!empty($oauth2_info['client_credentials_uid'])) {
+        $result['user_id'] = $oauth2_info['client_credentials_uid'];
+      }
+      elseif (!empty($oauth2_info['client_credentials_user'])) {
+        $result['user_id'] = user_load_by_name($oauth2_info['client_credentials_user'])->uid;
+      }
     }
     return user_load($result['user_id']);
   }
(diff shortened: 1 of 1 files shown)


src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -51,8 +51,22 @@ public function authenticate(RequestInterface $request) {
...
+        return NULL;
+      }
+
+      // We are dealing with client credentials flow. See if the resource has
+      // defined an user for this grant type.
+      if (!empty($oauth2_info['client_credentials_uid'])) {
+        $result['user_id'] = $oauth2_info['client_credentials_uid'];
+      }
+      elseif (!empty($oauth2_info['client_credentials_user'])) {
+        $result['user_id'] = user_load_by_name($oauth2_info['client_credentials_user'])->uid;
```
e0ipso, line 68: We are loading the user by name and then by id. We should load the user only once.

src/Plugin/authentication/OAuth2ServerAuthentication.php
```diff
@@ -84,7 +98,14 @@ protected function getOAuth2Info(RequestInterface $request) {
 
     $server = $plugin_definition['oauth2Server'];
     $scope = !empty($plugin_definition['oauth2Scope']) ? $plugin_definition['oauth2Scope'] : '';
-    return ['server' => $server, 'scope' => $scope];
+    $cc_user = !empty($plugin_definition['oauth2ClientCredentialsUser']) ? $plugin_definition['oauth2ClientCredentialsUser'] : '';
+    $cc_uid = !empty($plugin_definition['oauth2ClientCredentialsUid']) ? $plugin_definition['oauth2ClientCredentialsUid'] : '';
+    return [
+      'server' => $server,
+      'scope' => $scope,
+      'client_credentials_user' => $cc_user,
```
e0ipso, line 106: Is there any impediment to make it camel case? If not, I'd rather camel case since it's consistent with the rest of the annotation keys.

e0ipso (changes requested):
This implementation is getting complex. That is fine because this is a complex integration.

However we missed in the last iteration one of the key factors of RESTful, which is providing an Example on how to use all these features.

This PR is blocked by https://github.com/RESTful-Drupal/restful/pull/945. Once that one is merged this one should be adapted to use the same structure.

THANKS for the great work. Sorry it took so long, but I am currently spread too thin across multiple contribs and core.

🙌

PR #945, opened 2016-09-05
Use arrays to define OAuth2 information in the plugin definition

hussainweb:
Currently, we use two pieces of information for OAuth2 authentication - server and scope. We define them as separate items in the plugin definition. There might be more items required in the future (one of which could be support for client credentials in #944). It might be cleaner to change all the definitions to an array.


e0ipso:
Unfortunately this has already been tagged. If we apply this changes we'd be breaking backwards compatibility. I'd be however happy to merge this if there was support for specifying `oauth2Server`and `oauth2Scope` in the root level.

hussainweb:
Sorry for the long delay. I added support for the older (current) style of specifying OAuth2 information as well. I also changed the structure of `if` conditions to be more readable, hopefully.

PR #959, opened 2016-11-22
Minor correction in api_url.md

bnagaonkar:
Array index is wrong in the query formation example.


e0ipso:
Thanks for the correction!

PR #971, opened 2017-01-29
Allow filtering by sub_properties of addressfield

paean99:
A simple resource with addressfield fields:
```php
protected function publicFields()
    {
        $fields = parent::publicFields();

        $fields['country'] = array(
            'property' => 'field_address',
            'sub_property' => 'country'
        );

        $fields['locality'] = array(
            'property' => 'field_address',
            'sub_property' => 'locality'
        );

        return $fields;
    }
```
Cannot be filtered by the sub_properties of the addressfield. Every sub_properties will be default to the first column. [See line 744 in the code.](https://github.com/RESTful-Drupal/restful/blob/7.x-2.x/src/Plugin/resource/Field/ResourceFieldEntity.php#L737-L744)
Here is an exemple of the query:
`?filter[locality][value]=query&filter[locality][operator]=CONTAINS`

This PR adds a specific test for the addressfield sub_properties.
A more general solution could be achieved with testing the $field['type'] to an array of names of fields modules that have similar problems instead of just a simple string.

### DIFF
--- a/src/Plugin/resource/Field/ResourceFieldEntity.php
+++ b/src/Plugin/resource/Field/ResourceFieldEntity.php
@@ -739,7 +739,7 @@ class ResourceFieldEntity implements ResourceFieldEntityInterface {
         // Do not default to format.
         $this->setColumn('value');
       }
-      elseif ($field['type'] == 'addressfield' && array_key_exists($this->subProperty, $field['columns']) ) {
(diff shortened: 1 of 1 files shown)


paean99:
I forgot the "column" key, that should solve the problem without this patch. 
Sorry :(

e0ipso:
No need to apologize @paean99. I'm sure this will serve as great documentation for other people!

Thanks for documenting the whole process!

PR #975, opened 2017-02-23
ISSUE 966 allow restful to work correctly with drush commands 

stovak:
When executing a drush command, many times the URL() function is not available. It' needs to be revoked if drupal_is_cli() returns true.


mkochendorfer (approved):
This fixed worked for me

Issue #977, opened 2017-03-01
token_auth on multiple devices


cedatif:
Hi,

Can I have multiple access token (and refresh_token) for one user : one account used on multiple devices with the restful_token_auth module ?

Thanks !

tnightingale:
This doesn't seem to be supported; restful_token_auth makes sure to only allow **one** <code>access_token</code>/<code>refresh_token</code> pair per user. This functionality also surprised us as (we have an app talking to Drupal via RESTful & RESTful Token) it's not uncommon for a user to want to log into the application with two or more devices.

Here's a **very** rough idea of the changes required to loosen this requirement:
```diff
diff --git a/modules/restful_token_auth/src/Entity/RestfulTokenAuthController.php b/modules/restful_token_auth/src/Entity/RestfulTokenAuthController.php
index a730f07..e136453 100644
--- a/modules/restful_token_auth/src/Entity/RestfulTokenAuthController.php
+++ b/modules/restful_token_auth/src/Entity/RestfulTokenAuthController.php
@@ -57,17 +57,17 @@ class RestfulTokenAuthController extends \EntityAPIController {
    */
   private function generateRefreshToken($uid) {
     // Check if there are other refresh tokens for the user.
-    $query = new \EntityFieldQuery();
-    $results = $query
-      ->entityCondition('entity_type', 'restful_token_auth')
-      ->entityCondition('bundle', 'refresh_token')
-      ->propertyCondition('uid', $uid)
-      ->execute();
-
-    if (!empty($results['restful_token_auth'])) {
-      // Delete the tokens.
-      entity_delete_multiple('restful_token_auth', array_keys($results['restful_token_auth']));
-    }
+    // $query = new \EntityFieldQuery();
+    // $results = $query
+    //   ->entityCondition('entity_type', 'restful_token_auth')
+    //   ->entityCondition('bundle', 'refresh_token')
+    //   ->propertyCondition('uid', $uid)
+    //   ->execute();
+    //
+    // if (!empty($results['restful_token_auth'])) {
+    //   // Delete the tokens.
+    //   entity_delete_multiple('restful_token_auth', array_keys($results['restful_token_auth']));
+    // }

     // Create a new refresh token.
     $values = array(
diff --git a/modules/restful_token_auth/src/Plugin/resource/AccessToken__1_0.php b/modules/restful_token_auth/src/Plugin/resource/AccessToken__1_0.php
index 7782658..f161ff7 100644
[...]

PR #982, opened 2017-03-10
Add support for multiple access tokens per user (#977)

tnightingale:
Cleaned up and fixed the diff posted in #977.

This adds support for multiple access tokens per-user, a common use-case where a user might want to authenticate with Drupal from multiple devices (phone, tablet, 3rd party service, etc).

This modifies previous functionality of reusing existing access token / removing extra refresh tokens (which broke if you had two+ devices attempting to refresh with the same refresh token). An alternative would be to support both strategies and add configuration option to toggle between them.


prudloff-insite:
Thanks for the patch!
We got bitten by this and I must say the non-patched behavior is quite surprising.

PR #986, opened 2017-04-02
Fix Return of Multiple bundles resource when there are no items

amitaibu:
Current PR makes sure we get the current return value, instead of an empty response.

localhost_unio_server_www_api_embeds_slide_31741_filter_label___foo_


amitaibu:
Error seems unrelated. I'll check it

PR #988, opened 2017-04-07
Token Auth PHP notice

AronNovak:
`user_multiple_role_edit([$uid], 'add_role', $rid);` results in a PHP notice and also it drops the token, it seems. We should not rely on all the magic that PHP does here, but check if we have that key in the array first.

### DIFF
--- a/modules/restful_token_auth/restful_token_auth.module
+++ b/modules/restful_token_auth/restful_token_auth.module
@@ -131,7 +131,7 @@ function restful_token_auth_restful_resource_alter(ResourceInterface &$resource)
  * Implements hook_user_update().
  */
 function restful_token_auth_user_update(&$edit, $account, $category) {
-  if ($edit['status']) {
+  if (isset($edit['status']) && $edit['status']) {
     return;
   }
 


modules/restful_token_auth/restful_token_auth.module
```diff
@@ -131,7 +131,7 @@ function restful_token_auth_restful_resource_alter(ResourceInterface &$resource)
  * Implements hook_user_update().
  */
 function restful_token_auth_user_update(&$edit, $account, $category) {
-  if ($edit['status']) {
+  if (isset($edit['status']) && $edit['status']) {
```
amitaibu, line 134: should be enough to do `if (!empty($edit['status'])) {`

AronNovak: That's unfortunately not enough.
We have three cases:
 - `status` is missing from the array
 - `status` is there, user is blocked
 - `status` is there, user is active

If we use `empty`, we blur the difference between the first two, we need to inspect first and last option together, not first and the second one. If we simplify it to `empty`, it means that if the `status` property is missing from the array, we drop the token as we would do for a blocked user.

PR #1002, opened 2017-09-05
Data provider: move to function apart isListRequest function

omarlopesino:
Hi.

The changes from this PR are make a method in _Drupal\restful\Plugin\resource\DataProvider_ class to check is a list request. In this way other class can extend and determine if this is a list request by themselves.

Use case: a resource wich is using path like this 'resource/%' with only one value, no commas, that returns a list of entities using a custom data provider. This is useful because the custom data provider can use the filters   in his query for list.

Please review, thanks!

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProvider.php
+++ b/src/Plugin/resource/DataProvider/DataProvider.php
@@ -364,6 +364,16 @@ abstract class DataProvider implements DataProviderInterface {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


omarlopesino:
There is a strange error in travis console `The program 'drush' is currently not installed. To run 'drush' please ask your administrator to install the package 'drush'`, but the changes of PR shouldn't be the cause.

What can be happening?

PR #1003, opened 2017-09-15
Add sorts to the count query.

jsacksick:
When sorting on a field, a join on the field data table is added to the query.
So entities that don't have any values for that field will be filtered out.
However, the count query doesn't add sorts to the query causing a mismatch between the count() and the query for listing.

Additionally, the getQueryForList() catches exceptions that are not thrown, we should catch "BadRequestException" exceptions instead of "ServerConfigurationException".

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderEntity.php
+++ b/src/Plugin/resource/DataProvider/DataProviderEntity.php
@@ -527,13 +527,13 @@ class DataProviderEntity extends DataProvider implements DataProviderEntityInter
     try {
       $this->queryForListSort($query);
     }
-    catch (ServerConfigurationException $e) {
+    catch (BadRequestException $e) {
       watchdog_exception('restful', $e);
     }
     try {
       $this->queryForListFilter($query);
     }
-    catch (ServerConfigurationException $e) {
+    catch (BadRequestException $e) {
       watchdog_exception('restful', $e);
     }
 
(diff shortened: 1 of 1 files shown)


src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -552,8 +552,14 @@ protected function getQueryForList() {
   protected function getQueryCount() {
     $query = $this->getEntityFieldQuery();
 
-    // If we are trying to filter on a computed field, just ignore it and log an
-    // exception.
+    // If we are trying to filter or sort on a computed field, just ignore it
+    // and log an exception.
+    try {
+      $this->queryForListSort($query);
```
e0ipso, line 558: We seem to be doing this twice. Any reason?

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -527,13 +527,13 @@ protected function getQueryForList() {
     try {
       $this->queryForListSort($query);
     }
-    catch (ServerConfigurationException $e) {
+    catch (BadRequestException $e) {
```
e0ipso, line 530: Can you elaborate a bit more on why this is needed?

e0ipso (changes requested):

src/Plugin/resource/DataProvider/DataProviderEntity.php
jsacksick: No ServerConfigurationException is thrown inside queryForListSort(), if you look at the method body, you'll see that only BadRequestException are thrown.

src/Plugin/resource/DataProvider/DataProviderEntity.php
jsacksick: What do you mean twice? The other queryForListSort() call happens in getQueryForList(), and I added it inside getQueryCount(), which contains almost the same logic, besides the call to $this->queryForListSort($query);

e0ipso: Ah! You're right. Never review code before 5:30am.

e0ipso (approved):

PR #1007, opened 2017-11-02
DataProviderNode: use the language code

vermario:
Without this, endpoints exposing nodes will return nodes of all languages, even if the endpoint is called with a language prefix for example.

This small change allows to filter the nodes by the language prefix, which is what we need in our use case, and also it seems some other people in the issue queue: https://github.com/RESTful-Drupal/restful/issues/980 for example.

Should we merge this then?

### DIFF
--- a/src/Plugin/resource/DataProvider/DataProviderNode.php
+++ b/src/Plugin/resource/DataProvider/DataProviderNode.php
@@ -17,22 +17,24 @@ class DataProviderNode extends DataProviderEntity implements DataProviderInterfa
   /**
    * Overrides DataProviderEntity::getQueryForList().
    *
-   * Expose only published nodes.
+   * Expose only published nodes, and of the current language.
    */
   public function getQueryForList() {
     $query = parent::getQueryForList();
     $query->propertyCondition('status', NODE_PUBLISHED);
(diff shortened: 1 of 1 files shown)


e0ipso:
I'm hesitant to merge this since I can see the opposite argument for this feature.

vermario:
hi! First of all, great module!

 I understand the problem, and there are 2 other solutions that I tried to get to what I need:

- **you can add a "language" public field and add then a filter for that**: the problem I see there is that then the request has the default language set (In my case, finnish) and if you load other entities as relationships (taxonomy terms) they might show up in finnish even if you have specified the fillter for english content (drupal7 has different ways to set up taxonomy terms)

- **you can create your own "LanguageNodeResource and DataProviderLanguageNode"**: `LanguageNodeResource` extends `NodeResource` and has the `dataProviderClassName` method that points to your new `DataProviderLanguageNode` class, that has the filtering by language like in this pull request. 

At the moment I am using this second solution, instead of a patch of this pull request, because I also needed other custom stuff in "my" types of nodes.

So, this is achievable with the current module, if you are willing to code a bit yourself...

e0ipso:
Thanks for documenting this. It may be handy for other people in your same scenario.

PR #1010, opened 2017-12-01
 Support for filter with commerce product

jmaties:


src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -1146,6 +1147,10 @@ protected function getReferencedId($value, ResourceFieldInterface $resource_fiel
       $target_entity_type = $field_info['settings']['target_type'];
       $bundles = empty($field_info['settings']['handler_settings']['target_bundles']) ? array() : $field_info['settings']['handler_settings']['target_bundles'];
     }
+    elseif (!empty($field_info['type']) && $field_info['type'] == 'commerce_product_reference') {
```
e0ipso: Is this duplicated code? If so, we should have an `||` condition in the `if` clause instead of duplicating.

e0ipso (changes requested):

jmaties:
You're right 👍

src/Plugin/resource/DataProvider/DataProviderEntity.php
jmaties: You're right!! 👍

jmaties (commented):
Clean duplicate code

src/Plugin/resource/DataProvider/DataProviderEntity.php
```diff
@@ -1142,7 +1143,8 @@ protected function getReferencedId($value, ResourceFieldInterface $resource_fiel
         $target_entity_type = 'taxonomy_vocabulary';
       }
     }
-    elseif (!empty($field_info['type']) && $field_info['type'] == 'entityreference') {
+    elseif (!empty($field_info['type'])
```
e0ipso, line 1146: I don't think this is correct. This will prevent the next `elseif` from being executed.

e0ipso (changes requested):

PR #1016, opened 2018-03-14
Allow authenticated users to generate tokens via the UI

juampynr:
I made this addition to the module to make it easier for users to get a token for the API.

Here is what a user would see at its profile:

image

And then, after generating the token:

image

### DIFF
--- a/modules/restful_token_auth/restful_token_auth.pages.inc
+++ b/modules/restful_token_auth/restful_token_auth.pages.inc
@@ -1,0 +1,92 @@
+<?php
+
+/**
+ * @file
+ * Page callbacks for restful_token_auth module.
+ */
+
+/**
+ * Generates and shows an API key.
+ *
+ * @param \stdClass $account
+ *   A user account object.
+ *
+ * @return array
+ *   A render array.
+ */
+function restful_token_auth_get_credentials($account) {
+  $query = new EntityFieldQuery();
+  $result = $query
+    ->entityCondition('entity_type', 'restful_token_auth')
+    ->entityCondition('bundle', 'access_token')
+    ->propertyCondition('uid', $account->uid)
+    ->execute();
+
+  if (empty($result['restful_token_auth'])) {
+    return drupal_get_form('restful_token_auth_generate_token_form', $account->uid);
+  }
+  else {
+    $id = key($result['restful_token_auth']);
+    $auth_token = entity_load_single('restful_token_auth', $id);
+    return array(
+      '#theme' => 'restful_token_auth_user_key',
+      '#token' => $auth_token->token,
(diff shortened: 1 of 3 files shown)


modules/restful_token_auth/restful_token_auth.pages.inc
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Form validator for generating a token.
+ *
+ * @param array $form
+ *   The form array.
+ * @param array $form_state
+ *   The form state array.
+ */
+function restful_token_auth_generate_token_form_validate($form, &$form_state) {
```
e0ipso, line 76: Do we need this?

modules/restful_token_auth/restful_token_auth.pages.inc
```diff
@@ -0,0 +1,92 @@
...
+ *   The form state array.
+ * @param int $uid
+ *   The user identifier.
+ *
+ * @return array
+ *   The form as a render array.
+ */
+function restful_token_auth_generate_token_form($form, &$form_state, $uid) {
+  $form['description'] = array(
+    '#markup' => t('<p>You don\'t have an API key yet. Click below to generate one.</p>'),
```
e0ipso, line 53: Let's avoid markup in the translation string. Maybe consider `#prefix` and `#suffix` for this.

modules/restful_token_auth/templates/user-key.tpl.php
```diff
@@ -0,0 +1,23 @@
...
+ * Available variables:
+ * - $token: The API token.
+ *
+ * @ingroup themeable
+ */
+?>
+<p>Your authentication token is: <code><?php print $token; ?></code></p>
+
+<p>Discover the API by running the following command in a terminal and replacing
+    [your-token] by the above string:</p>
```
e0ipso, line 17: I don't see `[your-token]` anywhere. Maybe it's a stale comment?

modules/restful_token_auth/templates/user-key.tpl.php
```diff
@@ -0,0 +1,23 @@
...
+
+<p>
+    <code>
+        curl -H 'access-token:<?php print $token; ?>' <?php print $GLOBALS['base_url'] ?>/api?all=true
```
e0ipso, line 21: Maybe it's better to have a preprocess function that uses `url(…)` to set up a variable. Then we only need to print the variable here. The current code will fail if the site is installed in a subdir, or there is a language path prefix, etc.

e0ipso (changes requested):
Wow! An awesome feature @juampynr.

Just a couple of minor thoughts.

PR #1020, opened 2018-03-18
Make the access token key in header doc more prominent

amitaibu:
@ordavidil, so we won't forget next time :)


ordavidil:
@amitaibu 
I thought it's handled via #584 ...
:|

PR #1025, opened 2018-05-05
Make sure the render cache is set on create/update.

jsacksick:
I realized the render cache was set only on view() (after doing a doGet on the Resource for instance).

I noticed that when creating/updating an entity (by calling doPost() & doPatch()) on the resource, no cache entry is created.

I spent quiet some time debugging the issue, and realized the check performed in isCacheEnabled() in the Formatter was returning FALSE, because of the $data->getContext().

On view(), CacheDecoratedResource is calling the view method on the CacheDecoratedDataProvider object.

While on create/update, we're not calling the methods on the CacheDecoratedDataProvider.

Additionally, we aren't calling setCacheFragments on the ResourceFieldCollection.

After the patch cache entry is created when I create an entity via the API, On update for some reasons, the cache entry seems to be created from time to time, unsure why...

### DIFF
--- a/src/Plugin/resource/DataProvider/CacheDecoratedDataProvider.php
+++ b/src/Plugin/resource/DataProvider/CacheDecoratedDataProvider.php
@@ -175,7 +175,16 @@ class CacheDecoratedDataProvider implements CacheDecoratedDataProviderInterface,
    * {@inheritdoc}
    */
   public function create($object) {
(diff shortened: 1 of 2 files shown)


e0ipso:
It seems that tests are failing due to the Travis build. Can you please run the tests and report if they pass?

Issue #1030, opened 2018-06-21
Sub-property cannot be set [1h]


AronNovak:
Using the 7.x-1.x branch, trying to set AddressField values

`EntityMetadataWrapperException: Invalid data value given. Be sure it matches the required data type and format. Value at node()->field_address: IN`

PR #1031, opened 2018-06-21
Setting sub_property on the entity

AronNovak:
#1030 

## Status
Addresses can be created with this patch, without that, the error in the issue appears.
image

### DIFF
--- a/plugins/restful/RestfulEntityBase.php
+++ b/plugins/restful/RestfulEntityBase.php
@@ -699,7 +699,7 @@ abstract class RestfulEntityBase extends \RestfulDataProviderEFQ implements \Res
       }
 
       $sub_property = $info['sub_property'];
-      if (!empty($sub_property)) {
+      if (!empty($sub_property) && is_object($wrapper->{$property_name}->{$sub_property})) {
         try {
           $wrapper->{$property_name}->{$sub_property}->set($field_value);
         }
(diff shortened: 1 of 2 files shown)


plugins/restful/RestfulEntityBase.php
```diff
@@ -698,7 +698,13 @@ protected function setPropertyValues(EntityMetadataWrapper $wrapper, $null_missi
         $field_value = $this->propertyValuesPreprocess($property_name, $request[$public_field_name], $public_field_name);
       }
 
-      $wrapper->{$property_name}->set($field_value);
+      $sub_property = $info['sub_property'];
+      if (empty($sub_property)) {
+        $wrapper->{$property_name}->set($field_value);
+      }
+      else {
+        $wrapper->{$property_name}->{$sub_property}->set($field_value);
```
AronNovak: @amitaibu is it correct or is it supposed to be handled on a higher level?

amitaibu: I believe looks right.

AronNovak:
@amitaibu should i create another issue to fix Travis there? Does it worth?

amitaibu:
Yes, I'd be happy to see a working Travis.

AronNovak:
@amitaibu now with green Travis

AronNovak:
@amitaibu This is green, but I absolutely dislike the way how it's done, however at the project we need it, it does the job for the address field and it does not break tests there either.

Issue #1051, opened 2019-02-16
1.x "queryForListFilter()" does not support "NOT IN" operator


ordavidil:
Hey,

Using _7.x-1.x_ version, it seems that the `queryForListFilter()` method doesn't support the **NOT IN** operator.

//cc @amitaibu

ordavidil:
Merged, closing...

PR #1054, opened 2019-02-17
 [BUGFIX] Prevent filtering by an empty array.

ordavidil:
//cc @amitaibu

### DIFF
--- a/plugins/restful/RestfulDataProviderDbQuery.php
+++ b/plugins/restful/RestfulDataProviderDbQuery.php
@@ -198,10 +198,15 @@ abstract class RestfulDataProviderDbQuery extends \RestfulBase implements \Restf
     foreach ($this->parseRequestForListFilter() as $filter) {
       if (in_array(strtoupper($filter['operator'][0]), array('IN', 'NOT IN', 'BETWEEN'))) {
         $column_name = $this->getPropertyColumnForQuery($public_fields[$filter['public_field']]);
-        if (is_array($filter['value']) && empty($filter['value']) && strtoupper($filter['operator'][0]) == 'NOT IN') {
-          // Skip filtering by an empty value when operator is 'NOT IN',
-          // since it throws an SQL error.
-          continue;
+        if (is_array($filter['value']) && empty($filter['value'])) {
+          if (strtoupper($filter['operator'][0]) == 'NOT IN') {
+            // Skip filtering by an empty value when operator is 'NOT IN',
+            // since it throws an SQL error.
+            continue;
+          }
+          // Since Drupal doesn't know how to handle an empty array within a
+          // condition we add the `NULL` as an element to the array.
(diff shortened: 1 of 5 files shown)


ordavidil:
@amitaibu - ready for review.

tests/RestfulListTestCase.test
```diff
@@ -552,6 +552,20 @@ class RestfulListTestCase extends RestfulCurlBaseTestCase {
     $result = $handler->get('', $request);
     $this->assertEqual(count($result), 2, 'List filtered by the "author" entity metadata wrapper property, which maps to the "uid" DB column name, using the "NOT IN" operator.');
 
+    // Filter by an entity metadata wrapper property that is different from the
+    // DB column name using the 'NOT IN' operator.
+    $request = array(
+      'filter' => array(
+        'user' => array(
+          'value' => array(),
+          'operator' => 'NOT IN',
```
amitaibu, line 561: Lets try this one with `IN` -- and I'd expect to get `this->assertEqual(count($result), 0,)`

ordavidil: Since we skip the filter when the array is empty then we won't get this result.

Those queries seems problematic... I can add a special case for it though...

ordavidil:
@amitaibu - ready for another review (not sure about the implementation though)

plugins/restful/RestfulDataProviderDbQuery.php
```diff
@@ -198,6 +198,16 @@ protected function queryForListFilter(\SelectQuery $query) {
...
       if (in_array(strtoupper($filter['operator'][0]), array('IN', 'NOT IN', 'BETWEEN'))) {
         $column_name = $this->getPropertyColumnForQuery($public_fields[$filter['public_field']]);
+        if (is_array($filter['value']) && empty($filter['value'])) {
+          if (strtoupper($filter['operator'][0]) == 'NOT IN') {
+            // Skip filtering by an empty value when operator is 'NOT IN',
+            // since it throws an SQL error.
+            continue;
+          }
+          // Since Drupal doesn't know how to handle an empty array within a
+          // condition we add the `NULL` as an element to the array.
```
amitaibu, line 208: So the idea here is to make sure we get no results back in such a case?

ordavidil: I think yes, if someone is trying to filter by a value that is actually an empty array, it means by `NULL`.
At least as I see it.
I've split the `NOT IN` case since it's really a different scenario and when you pass the same `[NULL]` with `NOT IN` you also get no results, which is wrong...

ordavidil:
@amitaibu - also fixed the arrays to support older PHP versions.

amitaibu:
Thanks.

ordavidil:
Just adding the reference to #1051

